PR #31, opened 2023-01-31
circuits: zk-circuits: `VALID MATCH ENCRYPTION`: Implement fixed-point arithmetic

joeykraut:
### Purpose
This PR makes updates to the existing `VALID MATCH ENCRYPTION` circuit structure after the match process was refactored to use `FixedPoint` variables instead of direct integers. As a result, much of the note validation had to change; including defining how rounding is done in the case of fixed-point imprecision.

A follow up PR will add more test cases for invalid witness types; largely around note structure. This PR already includes invalid ciphertext tests.

### Testing
- All unit tests pass
- Added invalid witness tests for all encrypted fields


joeykraut:
Merging around CI checks; project does not build in CI currently, see [this issue](https://github.com/mcginty/snow/issues/146)

PR #57, opened 2023-03-13
core: state: initialize: Generate `VALID COMMITMENTS` from on-chain Merkle state

joeykraut:
### Purpose
This PR makes the initial connection between the p2p network and the on-chain Merkle state.

Specifically, when a wallet is added to a node's list of managed wallets, the node will:
1.  Scan on-chain state to find the entry of the wallet in the commitment tree, this gives the wallet's leaf index.
2. Scan on-chain state to find the last known change to each of the wallet's siblings in its commitment authentication path.
3. Construct the Merkle authentication proof for the wallet from the last updates to the siblings
4. Use this proof in `VALID COMMITMENTS` and index it with the wallet.

### Testing
- Unit tests
- Added two wallets' commitments to the Goerli state by calling `new_wallet`, ran two relayers against these wallets, verified that both created valid authentication proofs and then matched orders in the wallets.

### DIFF
--- a/core/src/config.rs
+++ b/core/src/config.rs
@@ -40,7 +40,7 @@ struct Cli {
     // | Environment Configs |
     // -----------------------
     /// The address of the darkpool contract, defaults to the Goerli deployment
-    #[clap(long, value_parser, default_value = "0x07d45b975709b894c87ef0318b4f20924e5542ce85116d523c2faef602e491a0")]
+    #[clap(long, value_parser, default_value = "0x03d1937bccc781f0ccf9a3fe5432e7fc205abd1eb9bfa381efd70b712c05a0c7")]
     pub contract_address: String,
     
     // -------------------------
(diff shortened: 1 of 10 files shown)


core/src/config.rs
```diff
@@ -40,7 +40,7 @@ struct Cli {
     // | Environment Configs |
     // -----------------------
     /// The address of the darkpool contract, defaults to the Goerli deployment
-    #[clap(long, value_parser, default_value = "0x07d45b975709b894c87ef0318b4f20924e5542ce85116d523c2faef602e491a0")]
+    #[clap(long, value_parser, default_value = "0x03d1937bccc781f0ccf9a3fe5432e7fc205abd1eb9bfa381efd70b712c05a0c7")]
```
joeykraut, line 43: Re-deployed contract on Goerli with fresh state

PR #80, opened 2023-04-24
core: network-manager: Discover `local_addr` through `Identify` gossip

joeykraut:
### Purpose
This PR updates the core networking logic around peer discovery; part of this involves upgrading libp2p to incorporate some recent, convenient interface changes.

The core protocol changes are as follows:
- Make more direct use of the `Identify` protocol that was already in place. That is, when a response is received from this protocol with a piggybacked `observed_addr`, the local node stores the `observed_addr` so that it may gossip its now-known public IP address to its peers.
- Accept a known `public_ip` as a config parameter. This is especially useful for bootstrap servers, which should know their public IP ahead of time so that they may gossip a dialable address to all bootstrap clients.

### Testing
- Unit tests pass
- Tested bootstrapping of both remote and locally run nodes, both intra-cluster and inter-cluser
- Tested handshake logic after bootstrapping, verified that a full handshake + settle worked properly

### DIFF
--- a/core/src/gossip/orderbook.rs
+++ b/core/src/gossip/orderbook.rs
@@ -333,9 +333,10 @@ impl GossipProtocolExecutor {
             .map_err(|err| GossipError::StarknetRequest(err.to_string()))?
         {
             log::info!("got order with invalid merkle root, skipping...");
-            return Err(GossipError::ValidCommitmentVerification(
-                "invalid merkle root, not in contract history".to_string(),
-            ));
+            // TODO: Once the contract implements foreign field Merkle, error on this test
+            // return Err(GossipError::ValidCommitmentVerification(
+            //     "invalid merkle root, not in contract history".to_string(),
+            // ));
         }
 
         // Verify the proof
--- a/core/src/network_manager/composed_protocol.rs
+++ b/core/src/network_manager/composed_protocol.rs
@@ -9,12 +9,16 @@ use async_trait::async_trait;
 use libp2p::{
     core::upgrade::{read_length_prefixed, write_length_prefixed},
     futures::{AsyncRead, AsyncWrite, AsyncWriteExt},
-    gossipsub::{Gossipsub, GossipsubConfig, GossipsubEvent, MessageAuthenticity},
+    gossipsub::{
+        Behaviour as Gossipsub, Config as GossipsubConfig, Event as GossipsubEvent,
+        MessageAuthenticity,
+    },
     identify::{Behaviour as IdentifyProtocol, Config as IdentifyConfig, Event as IdentifyEvent},
     identity::Keypair,
     kad::{record::store::MemoryStore, Kademlia, KademliaEvent},
     request_response::{
-        ProtocolName, ProtocolSupport, RequestResponse, RequestResponseCodec, RequestResponseEvent,
+        Behaviour as RequestResponse, Codec as RequestResponseCodec, Event as RequestResponseEvent,
+        ProtocolName, ProtocolSupport,
     },
     PeerId,
 };
--- a/core/src/network_manager/composed_protocol.rs
+++ b/core/src/network_manager/composed_protocol.rs
@@ -33,11 +33,11 @@ use crate::gossip_api::gossip::{AuthenticatedGossipRequest, AuthenticatedGossipR
 
 use super::error::NetworkManagerError;
 
-/**
- * Constants
- */
+// -------------
+// | Constants |
+// -------------
 
-// The maximum size libp2p should allocate buffer space for
+/// The maximum size libp2p should allocate buffer space for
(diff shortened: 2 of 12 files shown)


core/src/gossip/orderbook.rs
```diff
@@ -333,9 +333,10 @@ impl GossipProtocolExecutor {
             .map_err(|err| GossipError::StarknetRequest(err.to_string()))?
         {
             log::info!("got order with invalid merkle root, skipping...");
-            return Err(GossipError::ValidCommitmentVerification(
-                "invalid merkle root, not in contract history".to_string(),
-            ));
+            // TODO: Once the contract implements foreign field Merkle, error on this test
```
joeykraut, line 336: This blocks MPCs from happening correctly, we will re-enable this check once the Cairo 1.0 rewrite supports `Dalek` scalars as inputs/outputs to/from the Merkle tree

core/src/network_manager/composed_protocol.rs
```diff
@@ -29,11 +33,11 @@ use crate::gossip_api::gossip::{AuthenticatedGossipRequest, AuthenticatedGossipR
...
-/**
- * Constants
- */
+// -------------
+// | Constants |
+// -------------
 
-// The maximum size libp2p should allocate buffer space for
+/// The maximum size libp2p should allocate buffer space for
 const MAX_MESSAGE_SIZE: usize = 1_000_000_000;
```
akirillo, line 41: This is not pressing, but down the line, should we keep track of these sorts of constants that are bounded by hardware specs so we can expose them as config options?

I imagine if people are gonna be running their own relayers they may want to tweak this sort of thing without forking & digging through the code

core/src/network_manager/manager.rs
```diff
@@ -376,23 +428,35 @@ impl NetworkManagerExecutor {
...
+                        // Map each resolved address into a SocketAddr and remove local addresses
+                        let peer_addr = all_peer_addrs
+                            .into_iter()
+                            .find(|addr| !is_local_addr(addr))
+                            .map(|addr| multiaddr_to_socketaddr(&addr, peer_port))
+                            .unwrap_or(None)
                             .ok_or_else(|| {
-                                NetworkManagerError::SerializeDeserialize(
-                                    ERR_PARSING_ADDR.to_string(),
-                                )
```
akirillo, line 387: Can we still get serialization/deserialization errors in `multiaddr_to_socketaddr`?
It's nice to have specific errors and idk if `ERR_NO_KNOWN_ADDR` is exactly fitting here.

Regardless though this is pretty nitpicky so feel free to dismiss

core/src/network_manager/manager.rs
```diff
@@ -427,6 +491,35 @@ impl NetworkManagerExecutor {
...
+    async fn handle_identify_event(
+        &mut self,
+        event: IdentifyEvent,
+    ) -> Result<(), NetworkManagerError> {
+        // Update the local peer's public IP address if it has not already been discovered
+        if let IdentifyEvent::Received { info, .. } = event {
+            if !self.discovered_identity {
+                // Replace the port if the discovered NAT port is incorrect
+                let mut local_addr = info.observed_addr;
+                replace_port(&mut local_addr, self.p2p_port);
```
akirillo, line 508: might be my lack of networking knowledge here... but if the observed IP from the Identify protocol says we communicated on port `x`, shouldn't we stick with that instead of changing it to `self.p2p_port` and gossiping that around to peers?

I guess my real question here is why would these 2 ports (observed vs `self.p2p_port`) be in mismatch? Could it be that our outbound messages come from a different port than the one we're receiving on?

akirillo (commented):
still building context but lgtm

akirillo (approved):
whoops meant to approve lol

core/src/network_manager/composed_protocol.rs
joeykraut: Yes definitely, we can get to this down the line like you said, but if you have any ideas now, let me know.

I'm thinking they just make their way into the system config, or a second layer config for tunables

core/src/network_manager/manager.rs
joeykraut: Yeah good question; one thing is that [NAT](https://en.wikipedia.org/wiki/Network_address_translation) boxes in between the connection endpoints may swizzle these ports to their own internal mapping. This is especially the case in e.g. a symmetric NAT. 

The observed address may then be the address that the NAT has assigned to the connection, not the address that the local node has bound to. Anecdotally, I noticed this happening during testing

core/src/network_manager/manager.rs
joeykraut: Yeah good suggestion, fixed

PR #81, opened 2023-04-28
circuits: types: Add secret share types

joeykraut:
### Purpose
This PR adds type definitions for wallet secret shares, including types that are allocated in a constrain system, and commitments to secret shares. These types will be the building blocks for the circuit redesign as a next step.

The refactored types are:
- balances
- orders
- fees
- keychain
- wallet

### Testing
- CI will fail while the refactor is in progress, no testing done.

### DIFF
--- a/circuits/src/types/balance.rs
+++ b/circuits/src/types/balance.rs
@@ -1,14 +1,17 @@
 //! Groups base and derived types for the `Balance` object
 
-use crypto::fields::biguint_to_scalar;
+use std::ops::Add;
+
+use crypto::fields::{biguint_to_scalar, scalar_to_biguint};
 use curve25519_dalek::{ristretto::CompressedRistretto, scalar::Scalar};
 use mpc_bulletproof::{
-    r1cs::{Prover, Variable, Verifier},
+    r1cs::{LinearCombination, Prover, Variable, Verifier},
     r1cs_mpc::{MpcProver, MpcVariable},
 };
 use mpc_ristretto::{
     authenticated_ristretto::AuthenticatedCompressedRistretto,
-    authenticated_scalar::AuthenticatedScalar, beaver::SharedValueSource, network::MpcNetwork,
+    authenticated_scalar::AuthenticatedScalar, beaver::SharedValueSource,
+    mpc_scalar::scalar_to_u64, network::MpcNetwork,
 };
 use num_bigint::BigUint;
 use rand_core::{CryptoRng, RngCore};
@@ -20,6 +23,10 @@ use crate::{
     types::{biguint_from_hex_string, biguint_to_hex_string},
     Allocate, CommitSharedProver, CommitVerifier, CommitWitness, LinkableCommitment,
 };
+
+// ---------------------
+// | Base Balance Type |
+// ---------------------
 
 /// Represents the base type of a balance in tuple holding a reference to the
 /// ERC-20 token and its amount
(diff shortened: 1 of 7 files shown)


circuits/src/types/balance.rs
```diff
@@ -284,3 +299,125 @@ impl<N: MpcNetwork + Send, S: SharedValueSource<Scalar>> CommitVerifier
...
+
+    fn add(self, rhs: BalanceSecretShareVar) -> Self::Output {
+        BalanceVar {
+            mint: self.mint + rhs.mint,
+            amount: self.amount + rhs.amount,
+        }
+    }
+}
+
+impl BalanceSecretShareVar {
```
akirillo, line 361: is it overbearing to make a `Blinder` trait that `BalanceSecretShare` and `BalanceSecretShareVar` both implement? assuming we'll also need blinding/unblinding for all the other secret share types for other wallet components

joeykraut: Not unreasonable, but that feels over-abstracted to me. It doesn't buy us any generic functionality at a higher level (these methods are always called through `Wallet::blind`) and their implementation is very simple, so I'd err on the side of readability + simplicity here

circuits/src/types/fee.rs
```diff
@@ -543,3 +556,173 @@ impl<N: MpcNetwork + Send, S: SharedValueSource<Scalar>> CommitVerifier
...
+        self.settle_key -= blinder;
+        self.gas_addr -= blinder;
+        self.gas_token_amount -= blinder;
+        self.percentage_fee -= blinder;
+    }
+}
+
+/// Represents a commitment to a fee secret share that has been allocated in a constraint system
+#[derive(Clone, Copy, Debug)]
+pub struct FeeSecretShareCommitment {
```
akirillo, line 663: just thinking out loud... don't know if this is a feature or a nuisance of Rust but we have these `X`, `XSecretShare`, `XSecretShareVar`, `XSecretShareCommitment` structs that all have the same field names (different types ofc), which feels annoying to redefine all of them... different impls too i know but still.

maybe, down the line, we can make a macro that derives these structs from a base `X` struct? It seems like `XSecretShare`, `XSecretShareVar`, `XSecretShareCommitment` all have fields of the same type (`Scalar`, `Variable`, `CompressedRistretto`, respectively), which would make this easier.

but definitely something to do down the line

joeykraut: Yeah this would be nice, and definitely something to consider, I'll paint some color into why the `circuits` crate is written this way.

Everything in the circuits crate is pretty verbose; we only really define traits to define interfaces for common data manipulation patterns, (i.e. commit, prove, verify). It's written this way to be maximally clear, auditable, and make bugs stick out more.

I think introducing macros has a convenience advantage at a high readability cost, and for the circuits crate so far I've thought that to be a negative tradeoff on balance considering how security critical this is. More than happy to re-evaluate this down the line :)

circuits/src/types/keychain.rs
```diff
@@ -275,118 +280,16 @@ impl CommitWitness for SecretSigningKey {
...
-        self.0.commit_witness(rng, prover)
-    }
-}
-
 // -----------------
 // | Keychain Type |
 // -----------------
 
-/// Represents the base type, defining four keys with various access levels
+/// Represents the base type, defining two keys with different access levels
```
akirillo, line 287: 🤌

circuits/src/types/wallet.rs
```diff
@@ -55,30 +70,34 @@ where
...
     /// The list of payable fees in the wallet
-    pub fees: [FeeVar; MAX_FEES],
+    pub fees: [FeeVar<L>; MAX_FEES],
     /// The key tuple used by the wallet; i.e. (pk_root, pk_match, pk_settle, pk_view)
-    pub keys: PublicKeyChainVar,
-    /// The wallet randomness used to blind commitments, nullifiers, etc
-    pub randomness: Variable,
+    pub keys: PublicKeyChainVar<L>,
+    /// The wallet randomness used to blind secret shares
+    pub blinder: Variable,
```
akirillo, line 100: Should this be `L` instead of `Variable`?

circuits/src/zk_gadgets/nonnative.rs
```diff
@@ -882,6 +894,159 @@ impl NonNativeElementVar {
...
+/// Represents an additive secret share of a non-native element
+#[derive(Clone, Debug)]
+pub struct NonNativeSecretShare {
+    /// The underlying BigUint
+    pub(crate) val: BigUint,
+    /// The modulus of the field this element is defined over
+    pub(crate) field_mod: FieldMod,
+}
+
+/// Represents an additive secret share of a non-native element
```
akirillo, line 910: nit: not immediately clear what the difference is between this and the `NonNativeSecretShare` struct above... could we make this comment different?

akirillo (commented):
Couple nits, couple suggestions for down the line, couple questions.
Overall LGTM, but still building context on a lot of things here ofc

circuits/src/types/fee.rs
akirillo: Mm yeah, I like this rationale - keeping in mind auditability as an axis along which to evaluate the code

circuits/src/types/wallet.rs
joeykraut: yeah, good catch, fixed

circuits/src/zk_gadgets/nonnative.rs
joeykraut: Oops, moving too fast and I duplicated it, remove the one above. Thanks!

PR #82, opened 2023-04-29
circuits: zk-circuits: `VALID WALLET CREATE`: Implement refactored circuit

joeykraut:
### Purpose
This PR adds in the new definition of the `VALID WALLET CREATE` circuit to the refactor. This circuit validates that:
1. The commitment to the private wallet shares is correctly computed. This commitment is given as a public variable.
2. The underlying wallet, constructed as a combination of secret shares, is a valid, zero'd out wallet.

### Testing
- Unit tests pass, tested cases for cheating prover on invalid commitment, non-zero order, and non-zero balance.

### DIFF
--- a/circuits/src/zk_circuits/mod.rs
+++ b/circuits/src/zk_circuits/mod.rs
@@ -39,8 +39,6 @@ mod test_helpers {
         pub(crate) static ref PUBLIC_KEYS: PublicKeyChain = PublicKeyChain {
             pk_root: BigUint::from(1u8).into(),
             pk_match: compute_poseidon_hash(&[PRIVATE_KEYS[1]]).into(),
-            pk_settle: compute_poseidon_hash(&[PRIVATE_KEYS[2]]).into(),
-            pk_view: Scalar::one().into(),
         };
         pub(crate) static ref INITIAL_BALANCES: [Balance; MAX_BALANCES] = [
             Balance { mint: 1u8.into(), amount: 5 },
@@ -78,7 +76,7 @@ mod test_helpers {
             orders: INITIAL_ORDERS.clone(),
             fees: INITIAL_FEES.clone(),
             keys: PUBLIC_KEYS.clone(),
-            randomness: Scalar::from(42u64)
+            blinder: Scalar::from(42u64)
         };
     }
 
--- a/circuits/src/zk_circuits/mod.rs
+++ b/circuits/src/zk_circuits/mod.rs
@@ -7,25 +7,33 @@ pub mod valid_wallet_create;
 mod test_helpers {
     use ark_crypto_primitives::sponge::{poseidon::PoseidonSponge, CryptographicSponge};
     use crypto::{
-        fields::{prime_field_to_scalar, scalar_to_prime_field, DalekRistrettoField},
+        fields::{
+            biguint_to_scalar, prime_field_to_scalar, scalar_to_prime_field, DalekRistrettoField,
+        },
         hash::default_poseidon_params,
     };
     use curve25519_dalek::scalar::Scalar;
     use itertools::Itertools;
     use lazy_static::lazy_static;
     use num_bigint::BigUint;
-    use rand_core::{CryptoRng, RngCore};
+    use rand_core::{CryptoRng, OsRng, RngCore};
 
     use crate::{
         native_helpers::{compute_poseidon_hash, compute_wallet_commitment},
         types::{
-            balance::Balance,
-            fee::Fee,
-            keychain::{PublicKeyChain, NUM_KEYS},
-            order::{Order, OrderSide},
-            wallet::Wallet,
+            balance::{Balance, BalanceSecretShare},
+            fee::{Fee, FeeSecretShare},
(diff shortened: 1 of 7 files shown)


circuits/src/zk_circuits/valid_wallet_create.rs
```diff
@@ -222,11 +180,50 @@ impl<const MAX_FEES: usize> CommitVerifier for ValidWalletCreateCommitment<MAX_F
...
+{
+    type VarType = ValidWalletCreateStatementVar<MAX_BALANCES, MAX_ORDERS, MAX_FEES>;
+    type ErrorType = (); // Does not error
+
+    fn commit_public<CS: RandomizableConstraintSystem>(
+        &self,
+        cs: &mut CS,
+    ) -> Result<Self::VarType, Self::ErrorType> {
+        let private_commitment_var = self.private_shares_commitment.commit_public(cs).unwrap();
+        let public_shares_var = self.public_wallet_shares.commit_public(cs).unwrap();
```
akirillo, line 220: Is it safe to be unwrapping here (and in the other commitment traits)? prob ok for now to get this out the door anyway

circuits/src/zk_circuits/valid_wallet_create.rs
```diff
@@ -301,107 +286,136 @@ where
...
+        assert_invalid_witness_statement(witness, statement);
+    }
+
+    /// Tests the case in which a non-zero order is given
+    #[test]
+    fn test_nonzero_order() {
+        let mut wallet = create_empty_wallet();
+        wallet.orders[0] = INITIAL_ORDERS[0].clone();
+
+        let (witness, statement) = create_witness_statement_from_wallet(&wallet);
```
akirillo, line 408: i may be missing something obvious, but why should this test (and the one below) fail? yeah, we fudge the orders to be non-zero, but we create a witness statement from the wallet after having done this, and afaict `create_witness_statement_from_wallet` makes no assumptions about the wallet thats passed in being an empty wallet.

perhaps in this line we should call `create_default_witness_statment`?

akirillo (commented):
had some questions, but assuming tests pass lgtm!
interfaces looking clean

akirillo:
Looks like you're getting some build errors, too?

circuits/src/zk_circuits/valid_wallet_create.rs
joeykraut: Most of these commitment traits do not error, the error is allowed for generics, but this `unwrap` and most others are safe

circuits/src/zk_circuits/mod.rs
```diff
@@ -96,11 +102,142 @@ mod test_helpers {
...
+        let mut wallet2 = SizedWalletShare {
+            balances: balances2.try_into().unwrap(),
+            orders: orders2.try_into().unwrap(),
+            fees: fees2.try_into().unwrap(),
+            keys: keychain2,
+            blinder: wallet.blinder - blinder_share,
+        };
+
+        // Blind the shares
+        wallet1.blind();
```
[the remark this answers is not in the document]
joeykraut: > Looks like you're getting some build errors, too?

Yeah, CI will fail for a while during this refactor. In fact the last step of the implementation plan is what will fix CI (integrate the updated circuits into `core`). This is why I'm targeting an integration branch; to keep `main` green.

joeykraut:
> 

Yeah, CI will fail for a while during this refactor. In fact the last step of the implementation plan is what will fix CI (integrate the updated circuits into core). This is why I'm targeting an integration branch; to keep main green.

circuits/src/zk_circuits/valid_wallet_create.rs
joeykraut: We want `create_witness_statement_from_wallet` to create an invalid witness; the fact that it makes no assumptions *is* the desired behavior. This test case is testing that the constraints are not satisfied for an invalid witness; and `create_default_witness_statement` creates a *valid* witness.

circuits/src/zk_circuits/mod.rs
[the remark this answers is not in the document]
joeykraut: You're correct, changed this for testing, forgot to change back, good catch!

circuits/src/zk_circuits/valid_wallet_create.rs
```diff
@@ -57,66 +47,54 @@ pub struct ValidWalletCreate<
...
-        // Check that the commitment is to an empty wallet with the given randomness
-        // keys, and fees
-        Self::check_commitment(cs, expected_commit, witness)?;
+        // Validate the commitment given in the statement is a valid commitment to the private secret shares
+        let commitment =
+            WalletShareCommitGadget::compute_commitment(&witness.private_wallet_share, cs)?;
+        cs.constrain(commitment - statement.private_shares_commitment);
+
+        // Unblind the shares then reconstruct the wallet
+        witness.private_wallet_share.unblind();
```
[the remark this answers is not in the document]
joeykraut: See above

PR #83, opened 2023-05-01
circuits: zk-circuits: `VALID WALLET UPDATE`: Implement new design

joeykraut:
### Purpose
This PR implements the new design of `VALID WALLET UPDATE` which checks that:
- Previous wallet shares are valid entries in the state tree
- The previous wallet share nullifiers were correctly computed
- The commitment to the new private wallet shares is correctly computed
- The wallet has a `pk_root` value equal to one given in the statement
- The wallet transition is valid:
    - Updates to balances are the correct application of an external transfer
    - Updates to orders are correctly timestamped

### Testing
- Unit tests pass; verified many different cases:
    - Valid order placement and cancellation
    - Invalid order updates with incorrect timestamps
    - Valid deposit/withdrawal
    - Invalid deposit/withdrawal with maliciously engineered balance updates

### DIFF
--- a/circuits/src/zk_circuits/valid_wallet_update.rs
+++ b/circuits/src/zk_circuits/valid_wallet_update.rs
@@ -9,7 +9,10 @@
 
 use curve25519_dalek::scalar::Scalar;
 use mpc_bulletproof::{
-    r1cs::{Prover, R1CSProof, RandomizableConstraintSystem, Variable, Verifier},
+    r1cs::{
+        LinearCombination, Prover, R1CSProof, RandomizableConstraintSystem, Variable, Verifier,
+    },
+    r1cs_mpc::R1CSError,
     BulletproofGens,
 };
 use rand_core::{CryptoRng, OsRng, RngCore};
@@ -19,15 +22,26 @@ use crate::{
     errors::{ProverError, VerifierError},
     types::{
         keychain::PublicSigningKey,
+        order::OrderVar,
         transfers::{ExternalTransfer, ExternalTransferVar},
         wallet::{
             Nullifier, WalletSecretShare, WalletSecretShareCommitment, WalletSecretShareVar,
-            WalletShareCommitment,
+            WalletShareCommitment, WalletVar,
         },
     },
     zk_gadgets::{
-        merkle::{MerkleOpening, MerkleOpeningCommitment, MerkleOpeningVar, MerkleRoot},
+        commitments::{NullifierGadget, WalletShareCommitGadget},
+        comparators::{
+            EqGadget, EqVecGadget, EqZeroGadget, GreaterThanEqZeroGadget, NotEqualGadget,
+        },
+        fixed_point::FixedPointVar,
+        gates::{AndGate, ConstrainBinaryGadget, OrGate},
+        merkle::{
+            MerkleOpening, MerkleOpeningCommitment, MerkleOpeningVar, MerkleRoot,
+            PoseidonMerkleHashGadget,
+        },
         nonnative::NonNativeElementVar,
+        select::CondSelectGadget,
     },
     CommitPublic, CommitVerifier, CommitWitness, SingleProverCircuit,
 };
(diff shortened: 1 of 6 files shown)


circuits/src/zk_circuits/valid_wallet_update.rs
```diff
@@ -0,0 +1,1242 @@
...
+                    NotGate::not(new_balance_zero, cs),
+                ],
+                cs,
+            );
+
+            // Condition 3 -- withdrawal of entire balance, mint is now zero
+            let valid_mint3 = AndGate::multi_and(&[mint_is_zero, new_balance_zero], cs);
+
+            // Constrain one of the three mint conditions to hold
+            let valid_mint = OrGate::multi_or(&[valid_mint1, valid_mint2, valid_mint3], cs);
```
akirillo, line 267: should this be an `XOR`?
i get that logically it's impossible for any two of these clauses to be true at the same time, but an adversarial prover could try to set `valid_mint_i = valid_mint_j = 1` since they're new wires in the constraint system, no?

circuits/src/zk_circuits/valid_wallet_update.rs
```diff
@@ -0,0 +1,1242 @@
...
+            cs.constrain(Variable::One() - valid_mint);
+        }
+
+        // Validate that the external transfer's mint did show up in exactly one of the balances
+        let transfer_applied = EqGadget::eq(external_transfer_applied, Variable::One().into(), cs);
+        let transfer_applied_or_zero = OrGate::or(transfer_applied, external_transfer_zero, cs);
```
akirillo, line 273: similarly, should this be an `XOR`?

circuits/src/zk_circuits/valid_wallet_update.rs
```diff
@@ -0,0 +1,1242 @@
...
+            external_transfer.direction.into(),
+            cs,
+        );
+
+        // Stores a boolean indicating whether the external transfer appeared in the new balances
+        // it must be applied to at least one balance
+        let mut external_transfer_applied: LinearCombination = Variable::Zero().into();
+
+        // Check the balance state transition for each balance
+        for (old_balance, new_balance) in old_wallet.balances.iter().zip(new_wallet.balances.iter())
```
akirillo, line 201: I'm a little confused by the structure of this iteration.
it seems to rely on a proper ordering of balances to be zipped, i.e. any pair in the zip must be either:
1. (`mint_i`, `mint_i`)
2. (`zero_mint`, `mint_i`)
3. (`mint_i`, `zero_mint`)
but a simple off-by-one could throw off the whole circuit, i.e.:
```
(`zero_mint`, `mint_i`)
(`mint_i`, `zero_mint`)
```
how do we know the iterators are properly sorted / have the right number of zero mints/zero balances in the right places? is this handled in the `Iterator` impl for `wallet.balances`?

circuits/src/zk_circuits/valid_wallet_update.rs
```diff
@@ -0,0 +1,1242 @@
...
+    // ----------
+
+    /// Validates the orders of the new wallet
+    fn validate_order_updates<CS: RandomizableConstraintSystem>(
+        old_wallet: &WalletVar<MAX_BALANCES, MAX_ORDERS, MAX_FEES, LinearCombination>,
+        new_wallet: &WalletVar<MAX_BALANCES, MAX_ORDERS, MAX_FEES, LinearCombination>,
+        new_timestamp: Variable,
+        cs: &mut CS,
+    ) {
+        // Ensure that all order's assert pairs are unique
```
akirillo, line 311:
```suggestion
        // Ensure that all order's asset pairs are unique
```

circuits/src/zk_circuits/valid_wallet_update.rs
```diff
@@ -0,0 +1,1242 @@
...
+    /// For each order, if the order is unchanged from the previous wallet, the timestamp
+    /// should remain constant. Otherwise, the timestamp should be updated to the current
+    /// timestamp passed as a public variable
+    fn constrain_updated_order_timestamps<CS: RandomizableConstraintSystem>(
+        old_wallet: &WalletVar<MAX_BALANCES, MAX_ORDERS, MAX_FEES, LinearCombination>,
+        new_wallet: &WalletVar<MAX_BALANCES, MAX_ORDERS, MAX_FEES, LinearCombination>,
+        new_timestamp: Variable,
+        cs: &mut CS,
+    ) {
+        for (old_order, new_order) in old_wallet.orders.iter().zip(new_wallet.orders.iter()) {
```
akirillo, line 328: similar question about iterator ordering here, how can we be sure that the "logically same" orders are being zipped together here?
e.g. let's say we had an array of orders represented by asset pairs in the following way:
```
[
  (a0, a1),
  (a2, a3),
  (a4, a5),
]
```
and then we remove the second order `(a2, a3)`. is it the prover's responsibility to make sure they encode the new wallet with the following orders:
```
[
  (a0, a1),
  (zero_asset, zero_asset),
  (a4, a5)
]
```
so it zips properly?

circuits/src/zk_circuits/valid_wallet_update.rs
```diff
@@ -0,0 +1,1242 @@
...
+        new_wallet.orders[0] = Order::default();
+
+        assert!(constraints_satisfied_on_wallets(
+            old_wallet,
+            new_wallet,
+            ExternalTransfer::default()
+        ))
+    }
+
+    /// Tests placing an order without updating the timestamp
```
akirillo, line 910: this test feels a little strange to me... if i'm placing a new order, how can you tell if the timestamp has been "updated" or not? there's no old order to compare to.
unless the design here is that all "zero orders" come with a default timestamp that new orders must update from. if so:
1. is that only the case during testing?
2. if it's true in prod too, what is the default timestamp?

circuits/src/zk_circuits/valid_wallet_update.rs
```diff
@@ -0,0 +1,1242 @@
...
+            account_addr: BigUint::from(0u8),
+        };
+
+        assert!(!constraints_satisfied_on_wallets(
+            old_wallet, new_wallet, transfer
+        ));
+    }
+
+    /// Tests an invalid deposit in which the prover adds an unrelated balance
+    /// as well as the updated balance
```
akirillo, line 1195: is it worth doing a withdrawal test case like this as well, i.e. "withdraw from one balance but also update another"?

circuits/src/zk_circuits/valid_wallet_update.rs
```diff
@@ -0,0 +1,1242 @@
...
+    // -------------------------------
+
+    /// Tests the case in which a prover zeros a balance without any external
+    /// transfer corresponding to the update
+    #[test]
+    fn test_malicious_prover__spuriously_zerod_balance() {
+        let old_wallet = INITIAL_WALLET.clone();
+        let mut new_wallet = INITIAL_WALLET.clone();
+
+        // Prover also tries to increment the non-updated balance
```
akirillo, line 1233: is this comment accurate?

akirillo (commented):
generally lgtm but a couple questions about iterator ordering

circuits/src/zk_circuits/valid_wallet_update.rs
```diff
@@ -0,0 +1,1242 @@
...
+        let recovered_blinder = witness.old_wallet_private_shares.blinder.clone()
+            + witness.old_wallet_public_shares.blinder.clone();
+        witness.old_wallet_public_shares.unblind(recovered_blinder);
+
+        let old_wallet = witness.old_wallet_private_shares + witness.old_wallet_public_shares;
+
+        // Verify that the nullifiers of the two secret shares are correctly computed
+        let public_nullifier = NullifierGadget::wallet_shares_nullifier(
+            old_public_shares_comm,
+            old_wallet.blinder.clone(),
```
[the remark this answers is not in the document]
joeykraut: correct

akirillo (approved):
good for merge

circuits/src/zk_circuits/valid_wallet_update.rs
joeykraut: The prover should not shuffle the balances in between old and new wallet, see the test cases below. So the inputs to the circuit should come sorted. The validity checks on the mints and balances effectively enforce that all but one balance is updated, and that this balance update is consistent with the external transfer. So as long as the witness is not shuffled we're good

circuits/src/zk_circuits/valid_wallet_update.rs
joeykraut: Talked in person, copying here for transparency -- the inputs to the `OR` gate are implicitly mutually exclusive, so `OR` and `XOR` are equivalent. `OR` is more efficient

circuits/src/zk_circuits/valid_wallet_update.rs
joeykraut: same as above

circuits/src/zk_circuits/valid_wallet_update.rs
joeykraut: yes, prover's responsibility. We essentially define a canonical way of comparing two lists of balances or orders in the circuit, and it's the prover's responsibility to format the data into the witness such that this verifies.

circuits/src/zk_circuits/valid_wallet_update.rs
joeykraut: No default timestamp, this is testing the case in which a prover lies about the public input timestamp that the contract will verify. I'll update this comment to be more clear, thanks for pointing out

circuits/src/zk_circuits/valid_wallet_update.rs
```diff
@@ -0,0 +1,1242 @@
...
+        external_transfer: ExternalTransferVar,
+        cs: &mut CS,
+    ) {
+        // Zero out the external transfer amount if its mint is zero
+        let external_transfer_zero = EqZeroGadget::eq_zero(external_transfer.mint, cs);
+        let external_transfer_not_zero = NotGate::not(external_transfer_zero, cs);
+        let (_, _, transfer_amount) =
+            cs.multiply(external_transfer.amount.into(), external_transfer_not_zero);
+
+        // The term added to any balance matching the external transfer mint
```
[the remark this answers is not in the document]
joeykraut: fixed

circuits/src/zk_circuits/valid_wallet_update.rs
joeykraut: fixed

circuits/src/zk_circuits/valid_wallet_update.rs
joeykraut: fixed

circuits/src/zk_circuits/valid_wallet_update.rs
joeykraut: added

PR #84, opened 2023-05-02
circuits: zk-circuits: `VALID REBLIND`: Implement circuit

joeykraut:
### Purpose
This PR adds the implementation of the `VALID REBLIND` circuit. This circuit is used to re-blind a wallet ahead of an attempted match to preserve anonymity. The circuit checks that:
- The inputs to the reblinding are valid, non-nullified elements of the state tree
- A commitment to the new private shares is correctly computed
- The prover knows `sk_match`
- The reblinding results from a valid sampling of the wallet's CSPRNGs, then applied to the old blinding

### Testing
- Unit tests pass, tested the following cases among others:
    - Incorrectly sampled blinder
    - Incorrectly sampled secret share
    - Incorrectly applied blinder and secret share
    - Spurious wallet modifications during reblinding
    - Invalid state inputs
    - Unauthorized prover

### DIFF
--- a/circuits/src/mpc_gadgets/poseidon.rs
+++ b/circuits/src/mpc_gadgets/poseidon.rs
@@ -10,7 +10,7 @@
 //! Their poseidon implementation can be found here:
 //!     https://github.com/arkworks-rs/sponge/blob/master/src/poseidon/mod.rs
 
-use std::cell::Ref;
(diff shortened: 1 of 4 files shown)


akirillo (approved):
Much more cursory glance this time but lgtm, implementation lines up w what i expected & test cases seem to cover a lot

PR #85, opened 2023-05-03
circuits: zk-circuits: `VALID COMMITMENTS`: Implement re-designed circuits

joeykraut:
### Purpose
This PR implements the new design of `VALID COMMITMENTS`. This circuit leaks the indices of the wallets send and receive balances for an order, as well as the index of the order itself. This allows the counterparty to update the wallet secret shares at the correct index with the match result.

As well, this circuit proves that the claimed balances, orders, etc that are to be used in the match process are indeed valid elements of the wallet. 

This proof is commitment-linked to a proof of `VALID REBLIND` as a state inclusion check, and a proof of `VALID SETTLE` after match for secret share update validity.

### Testing
- Unit tests pass, tested various adversarial prover cases.


circuits/src/zk_gadgets/wallet_operations.rs
```diff
@@ -0,0 +1,211 @@
...
+
+use crate::{
+    mpc_gadgets::poseidon::PoseidonSpongeParameters,
+    types::{balance::BalanceVar, fee::FeeVar, order::OrderVar, wallet::WalletSecretShareVar},
+};
+
+use super::{comparators::EqVecGadget, poseidon::PoseidonHashGadget};
+
+// ------------------------
+// | Public State Gadgets |
```
joeykraut, line 16: The public state gadgets are unchanged from before, I added the wallet state primitive comparators, but the diff seems to suggest the whole file changed.

circuits/src/zk_circuits/valid_commitments.rs
```diff
@@ -0,0 +1,951 @@
...
+        Self::contains_balance_at_index(
+            statement.balance_receive_index,
+            witness.balance_receive,
+            &augmented_wallet,
+            cs,
+        );
+        cs.constrain(witness.balance_receive.mint - receive_mint);
+
+        // Verify that the fee balance is in the wallet
+        // TODO: Implement fees properly
```
joeykraut, line 115: Fees will come in a follow up, still need to sketch out the details of how they fit into the design

akirillo: Naively, what's so different about fees? Can't we secret-split in the same way?

circuits/src/zk_circuits/valid_commitments.rs
```diff
@@ -0,0 +1,950 @@
...
+        let (mut witness, statement) = create_witness_and_statement(&wallet);
+
+        // Modify a fee in the wallet
+        witness.augmented_public_shares.fees[0].gas_token_amount += Scalar::one();
+        assert!(!constraints_satisfied(witness, statement));
+    }
+
+    /// Test the case in which a prover attempts to modify wallet keys and blinders in augmentation
+    #[test]
+    fn test_invalid_commitment__augmentation_modifies_keys_blinder() {
```
akirillo, line 819: nit: should this be split out into 2 tests?

circuits/src/zk_circuits/valid_commitments.rs
```diff
@@ -0,0 +1,950 @@
...
+
+        // Modify the index of the order
+        statement.order_index += 1;
+
+        assert!(!constraints_satisfied(witness, statement))
+    }
+
+    /// Test the case in which a balance is missing from the wallet
+    #[test]
+    fn test_invalid_commitment__balances_missing() {
```
akirillo, line 872: super nit: should this be broken out into 3 tests?

akirillo (approved):
mostly some small nits that can be addressed once the code stabilizes

LFARTM

circuits/src/zk_circuits/valid_commitments.rs
joeykraut: Decrementing the fee is easy, you're correct. The problem is incrementing it elsewhere (i.e. the relayer and protocol wallets). It's not *hard* per se, it's just something I'm deferring until after the initial implementation to tame the complexity the first go around

circuits/src/zk_circuits/valid_commitments.rs
```diff
@@ -0,0 +1,950 @@
...
+    fn create_witness_and_statement(
+        wallet: &SizedWallet,
+    ) -> (SizedWitness, ValidCommitmentsStatement) {
+        let mut rng = thread_rng();
+
+        // Split the wallet into secret shares
+        let (private_shares, public_shares) = create_wallet_shares(wallet);
+
+        // Choose an order and fee to match on
+        let ind_order = 0;
```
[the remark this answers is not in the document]
joeykraut: Makes sense, I originally had it as random, but it makes it harder to predictably clobber invalid values in the test cases below. I'll leave it as static for now just to avoid complicating all the tests

circuits/src/zk_circuits/valid_commitments.rs
```diff
@@ -0,0 +1,950 @@
...
+        };
+
+        (witness, statement)
+    }
+
+    /// Finds a balance for the given order returning the index and the balance itself
+    ///
+    /// If the balance does not exist the `augment` flag lets the method augment the wallet
+    /// with a zero'd balance
+    fn find_balance(mint: BigUint, wallet: &mut SizedWallet, augment: bool) -> (usize, Balance) {
```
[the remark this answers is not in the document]
joeykraut: good suggestion, done

circuits/src/zk_circuits/valid_commitments.rs
joeykraut: fixed

circuits/src/zk_circuits/valid_commitments.rs
joeykraut: fixed

PR #88, opened 2023-05-06
core: Use `PENDING` state for finality, rather than `ACCEPTED_ON_L2`

ChrisBender:
### Purpose
This PR aims to use `PENDING` state for confirmation, rather than `ACCEPTED_ON_L2`. These two states are equivalent while StarkNet is still single-sequencer, and we get significantly better time-to-confirmation by using `PENDING`.

Main changes:
* Use `PENDING` everywhere, including in events streaming and transaction polling.
* Add support for multi-account rotation when submitting transactions. The StarkNet sequencer has a nonce check bug, such that multiple transactions from a single account are not possible within a single `PENDING` block, so we rotate between accounts for each tx.
* Stream a WalletUpdate event immediately after the transaction is confirmed as `PENDING`, rather than after `VALID_COMMITMENTS` proofs are generated.

### DIFF
--- a/core/src/config.rs
+++ b/core/src/config.rs
@@ -114,12 +114,12 @@ struct Cli {
     /// The HTTP addressable StarkNet JSON-RPC node
     #[clap(long = "starknet-gateway", value_parser)]
     pub starknet_jsonrpc_node: Option<String>,
-    /// The StarkNet address corresponding to the given key
-    #[clap(long = "starknet-account-addr", value_parser, requires = "starknet-private-key")]
-    pub starknet_account_address: Option<String>,
-    /// The StarkNet private key used to send transactions
-    #[clap(long = "starknet-account-pkey", value_parser, requires = "starknet-account-address")]
-    pub starknet_private_key: Option<String>,
+    /// The StarkNet addresses corresponding to the given private keys
+    #[clap(long = "starknet-account-addr", value_parser, requires = "starknet-private-keys")]
+    pub starknet_account_addresses: Option<Vec<String>>,
+    /// The StarkNet private keys used to send transactions
+    #[clap(long = "starknet-account-pkey", value_parser, requires = "starknet-account-addresses")]
+    pub starknet_private_keys: Option<Vec<String>>,
     /// A file holding a json representation of the wallets the local node
     /// should manage
     #[clap(short, long, value_parser)]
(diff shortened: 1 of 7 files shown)


core/src/config.rs
```diff
@@ -114,12 +114,12 @@ struct Cli {
...
-    pub starknet_account_address: Option<String>,
-    /// The StarkNet private key used to send transactions
-    #[clap(long = "starknet-account-pkey", value_parser, requires = "starknet-account-address")]
-    pub starknet_private_key: Option<String>,
+    /// The StarkNet addresses corresponding to the given private keys
+    #[clap(long = "starknet-account-addr", value_parser, requires = "starknet-private-keys")]
+    pub starknet_account_addresses: Option<Vec<String>>,
+    /// The StarkNet private keys used to send transactions
+    #[clap(long = "starknet-account-pkey", value_parser, requires = "starknet-account-addresses")]
+    pub starknet_private_keys: Option<Vec<String>>,
```
joeykraut, line 122: Can we just make these `Vec<String>`? No need for the option anymore, it can just be empty

joeykraut: same everywhere else they're used

core/src/starknet_client/client.rs
```diff
@@ -139,7 +139,7 @@ impl StarknetClientConfig {
     ///
     /// Only when this is enabled may the client write transactions to the sequencer
     pub fn account_enabled(&self) -> bool {
-        self.starknet_pkey.is_some() && self.starknet_account_address.is_some()
+        self.starknet_pkeys.is_some() && self.starknet_account_addresses.is_some()
```
joeykraut, line 142: Related to the above, if we take the account info as `Vec<String>`, we just change this check to `!self.starknet_pkeys.is_empty() && ...`

core/src/starknet_client/client.rs
```diff
@@ -187,28 +189,34 @@ impl StarknetClient {
...
-            let account = SingleOwnerAccount::new(
-                config.new_gateway_client(),
-                signer,
-                account_addr_felt,
-                config.chain.into(),
-            );
-
-            Some(account)
+        // Build the accounts to sign transactions with
+        let accounts = if config.account_enabled() {
```
joeykraut, line 193: We will not need this as a conditional if we make the aforementioned change

core/src/starknet_client/client.rs
```diff
@@ -244,9 +253,21 @@ impl StarknetClient {
...
+    /// Get the underlying account at the given index as an immutable reference
+    pub fn get_account(
+        &self,
+        account_index: usize,
+    ) -> &SingleOwnerAccount<SequencerGatewayProvider, LocalWallet> {
+        self.accounts.as_ref().unwrap().get(account_index).unwrap()
+    }
+
+    /// Query the current nonce and advance it by one
+    pub fn new_account_index(&mut self) -> usize {
```
joeykraut, line 265: nit: Maybe `increment_account_index` or `next_account_index`?

joeykraut (approved):
overall lgtm, just a few small suggestions, thanks for doing this!

Can we also `git rebase` to remove the `rustfmt` bot PR?

PR #89, opened 2023-05-06
core: tasks: create-wallet: Refactor task after encryption redesign

joeykraut:
### Purpose
This PR refactors the `create-wallet` task to fit the new secret share centric design. The changes needed are essentially that we now add secret shares of the wallet to the calldata, and change the interface of `VALID WALLET CREATE` to the updated proof.

### Testing
- Unit tests pass
- I will test the tasks together once they're all complete and fix bugs then, it's easier to chain them together than to isolate any single task.


core/src/chain_events/listener.rs
```diff
@@ -349,7 +349,19 @@ impl OnChainEventListenerExecutor {
                 .fetch_add(1u32, Ordering::Relaxed);
 
             self.update_wallet_merkle_path(
-                locked_wallet.merkle_proof.as_mut().unwrap(),
```
joeykraut, line 352: this change isn't functional, just done to get the binary building

core/src/tasks/helpers.rs
```diff
@@ -0,0 +1,27 @@
+//! Helpers for common functionality across tasks
+
+use crate::{
+    starknet_client::{client::StarknetClient, error::StarknetClientError},
+    state::wallet::{Wallet, WalletAuthenticationPath},
+};
+
+/// Find the merkle authentication path of a wallet
+pub(super) async fn find_merkle_path(
```
akirillo, line 9: do relayers maintain a synced local version of the merkle tree? to avoid only being able to query the contract for inclusion proofs

akirillo (approved):
lgtm, just a smol q

core/src/tasks/helpers.rs
joeykraut: Not currently; we may maintain a sparse representation in the future as an optimization, but for now they just recover the paths they need from contract state on demand.

PR #90, opened 2023-05-06
core: tasks: intialize-state: Refactor task to fit encryption redesign

joeykraut:
### Purpose
This PR refactors the `initialize-state` task to fit the secret share pattern in the encryption redesign.

At a high level, the changes are:
- Find merkle authentication paths for both the public and private wallet shares separately
- Prove both `VALID REBLIND` once for each wallet and `VALID COMMITMENTS` for each order in the wallet. Store both of these (heap allocated) proofs in the global state.

### Testing
- Unit tests pass
- Will defer testing concrete functionality until the rest of the tasks are refactored, easier to test the whole flow end-to-end.

### DIFF
--- a/circuits/src/lib.rs
+++ b/circuits/src/lib.rs
@@ -770,7 +770,7 @@ pub mod native_helpers {
     /// Reblind a wallet given its secret shares
     ///
     /// Returns the reblinded private and public shares
-    pub(crate) fn reblind_wallet<
+    pub fn reblind_wallet<
         const MAX_BALANCES: usize,
         const MAX_ORDERS: usize,
         const MAX_FEES: usize,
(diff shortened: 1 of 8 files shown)


circuits/src/lib.rs
```diff
@@ -753,6 +766,241 @@ pub mod native_helpers {
     ) -> Nullifier {
         compute_poseidon_hash(&[share_commitment, wallet_blinder])
     }
+
```
joeykraut, line 769: Moved these helpers from `test_helpers` to `native_helpers` to export them outside of the `circuits` crate. Other than that, they are the same

akirillo (approved):
lgtm

PR #91, opened 2023-05-06
core: tasks: update-wallet: Refactor to fit encryption redesign

joeykraut:
### Purpose
This PR refactors the `update-wallet` task to fit the secret sharing scheme of the encryption redesign. Specifically, this involves:
- Change the proofs of `VALID WALLET UPDATE` to fit the new design
- Change the post-tx logic to prove both `VALID REBLIND` as well as `VALID COMMITMENTS`
- Change the wallet update to index both the public and private share Merkle openings

### Testing
- Unit tests pass
- Did not test the functionality in goerli; will wait til all tasks are done.

### DIFF
--- a/core/src/tasks/helpers.rs
+++ b/core/src/tasks/helpers.rs
@@ -238,5 +238,9 @@ fn find_order(base_mint: &BigUint, quote_mint: &BigUint, wallet: &SizedWallet) -
         .iter()
         .enumerate()
         .find(|(_ind, order)| order.quote_mint.eq(quote_mint) && order.base_mint.eq(base_mint))
+<<<<<<< HEAD
         .map(|(ind, _order)| ind)
+=======
+        .map(|(ind, _balance)| ind)
+>>>>>>> 181e771 (core: tasks: initialize-state: Refactor task after encryption redesign)
 }
--- a/core/src/tasks/helpers.rs
+++ b/core/src/tasks/helpers.rs
@@ -1,5 +1,10 @@
 //! Helpers for common functionality across tasks
 
+use std::{
+    sync::Arc,
+    time::{SystemTime, UNIX_EPOCH},
+};
+
 use circuits::{
     native_helpers::{
         compute_wallet_share_commitment, create_wallet_shares_from_private, reblind_wallet,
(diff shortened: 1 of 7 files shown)


core/src/tasks/helpers.rs
```diff
@@ -240,3 +268,103 @@ fn find_order(base_mint: &BigUint, quote_mint: &BigUint, wallet: &SizedWallet) -
...
+                .await
+                .attach_validity_proof_witness(
+                    order_id,
+                    OrderValidityWitnessBundle {
+                        reblind_witness: wallet_reblind_witness.clone(),
+                        commitment_witness: order_commitment_witness.clone(),
+                    },
+                )
+                .await;
+        } // order_book lock released
```
akirillo, line 327: This is the same as what's used in your `initialize-state` task PR, right? (https://github.com/renegade-fi/renegade/blob/joey/initialize-state/core/src/tasks/initialize_state.rs#L246-L261)

If so, might be nice to pack this into a helper fn

core/src/tasks/update_wallet.rs
```diff
@@ -323,172 +286,51 @@ impl UpdateWalletTask {
...
-        self.network_sender
-            .send(GossipOutbound::Pubsub {
-                topic: ORDER_BOOK_TOPIC.to_string(),
-                message: PubsubMessage::OrderBookManagement(message),
-            })
-            .map_err(|err| UpdateWalletTaskError::SendMessage(err.to_string()))
+    /// After a wallet update has been submitted on-chain, find its authentication
+    /// path, and re-prove `VALID REBLIND` for the wallet and `VALID COMMITMENTS`
+    /// for all orders in the wallet
+    async fn update_validity_proofs(&self) -> Result<(), UpdateWalletTaskError> {
```
akirillo, line 325: maybe i'm blind, but do we actually call this after submitting the updated wallet on-chain?

akirillo: ah, a function w this name existed before, i take it we do call this it just doesn't show up in the diff

akirillo (approved):
lgtm, smol nit for a helper

core/src/tasks/helpers.rs
joeykraut: I'm not sure I see what a helper would do here? It's a single function call after all

core/src/tasks/update_wallet.rs
joeykraut: that's correct, call happens in state transition [here](https://github.com/renegade-fi/renegade/blob/main/core/src/tasks/update_wallet.rs#L172)

PR #92, opened 2023-05-07
core: tasks: settle-match: Update task after encryption redesign

joeykraut:
### Purpose
This PR refactors the `settle-match` task to fit the new secret-share centric system redesign. Match and settle are now atomic so the flow of this task is now:
- Update the wallet shares for *both* wallets and prove `VALID SETTLE` which validates the update
- Submit the match to the contract via the `match` function and await transaction completion
- Find the new wallet in the Merkle tree and re-prove `VALID REBLIND` for the wallet and `VALID COMMITMENTS` for each order.

### Testing
- Unit tests pass
- I am deferring functionality testing in Goerli until we have completely `lookup-wallet` as this greatly eases the end-to-end testing burden.


akirillo (approved):
lgtm

PR #93, opened 2023-05-08
core: tasks: lookup-wallet: Refactor task wrt encryption redesign

joeykraut:
### Purpose
This PR refactors the `lookup-wallet` task to fit the encryption redesign, the specific changes are:
- Instead of looking up a direct mapping from `pk_view`, we iterate over the blinder shares generated by the csprng and lookup the most recently used `public_blinder_share`.
- We index Merkle entries for both public and private shares
- We prove `VALID REBLIND` at the wallet level, and `VALID COMMITMENT` for each order in the wallet

### Tests
- Unit tests pass
- Will test the whole system end-to-end and open a PR that fixes all bugs then.

### DIFF
--- a/core/src/config.rs
+++ b/core/src/config.rs
@@ -46,7 +46,7 @@ struct Cli {
     #[clap(long)]
     pub chain_id: ChainId,
     /// The address of the darkpool contract, defaults to the Goerli deployment
-    #[clap(long, value_parser, default_value = "0x1e7857cdd3d73838b0e053be1fa068aa15113793fea95ab663501789d3d0b51")]
+    #[clap(long, value_parser, default_value = "0x02179f01358c09410f234f1ece40d235719e05db3babca7bef0babc974333162")]
     pub contract_address: String,
 
     // ----------------------------
--- a/core/src/tasks/lookup_wallet.rs
+++ b/core/src/tasks/lookup_wallet.rs
@@ -93,8 +93,6 @@ pub enum LookupWalletTaskError {
     NotFound(String),
     /// Error generating a proof of `VALID COMMITMENTS`
     ProofGeneration(String),
-    /// Error sending a message to another worker in the local relayer
-    SendMessage(String),
     /// Error interacting with the starknet client
     Starknet(String),
 }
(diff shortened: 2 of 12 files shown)


core/src/config.rs
```diff
@@ -46,7 +46,7 @@ struct Cli {
     #[clap(long)]
     pub chain_id: ChainId,
     /// The address of the darkpool contract, defaults to the Goerli deployment
-    #[clap(long, value_parser, default_value = "0x1e7857cdd3d73838b0e053be1fa068aa15113793fea95ab663501789d3d0b51")]
```
joeykraut, line 49: I deployed a new version of the contract to goerli with an updated interface given the encryption redesign

core/src/tasks/lookup_wallet.rs
```diff
@@ -177,243 +179,128 @@ impl LookupWalletTask {
...
+            let curr_blinder = blinder_csprng.next().unwrap();
+            let curr_blinder_private_share = blinder_csprng.next().unwrap();
+            curr_blinder_public = curr_blinder - curr_blinder_private_share;
 
-        if last_updated == StarknetFieldElement::ZERO {
-            return Err(LookupWalletTaskError::NotFound(
-                ERR_WALLET_NOT_FOUND.to_string(),
-            ));
+            blinder_index += 1;
         }
```
akirillo, line 203: I mentioned this in your writeup as well, but this assumes that the owner of the given `sk_root` never experienced an MPC hangup, which would force them to do another `reblind` even though there was no update / settle transaction for the previous blinders.

You mentioned an approach to this being that we can check both if a blinder is indexed on-chain, and if it's been nullified, and consider the current wallet to be one where the public blinder is indexed on-chain but not nullified.

If this should be implemented after v0 that's fine, but it should be mentioned here as a `TODO`

akirillo (changes requested):

core/src/tasks/lookup_wallet.rs
joeykraut: Added a todo.

I consider this as out of scope for v0, because the cause of the problem is not yet implemented either (i.e. we do not take any privacy preserving precautions after hangup yet). More of a privacy _optimization_ than a competion gating feature.

PR #94, opened 2023-05-11
Joey/proof linking

joeykraut:


core/src/starknet_client/client.rs
```diff
@@ -666,18 +667,19 @@ impl StarknetClient {
...
+        old_shares_nullifier: Nullifier,
         external_transfer: Option<ExternalTransfer>,
         new_public_shares: SizedWalletShare,
         valid_wallet_update: ValidWalletUpdateBundle,
     ) -> Result<TransactionHash, StarknetClientError> {
-        let public_share_commitment = compute_wallet_share_commitment(new_public_shares.clone());
+        let new_wallet_share_commitment = compute_wallet_commitment_from_private(
+            new_public_shares.clone(),
+            new_private_shares_commitment,
+        );
```
akirillo, line 678: If we are calculating here (relayer-side) the commitment to the new public+private shares, how do we verify that this commitment is correctly calculated? In the proof I only see us verifying that the commitment to the new private shares is correct

core/src/starknet_client/client.rs
```diff
@@ -708,10 +710,8 @@ impl StarknetClient {
     #[allow(clippy::too_many_arguments)]
```
akirillo, line 710: maybe can remove this now? :P

akirillo (approved):
I defer to your judgement / understanding of the system so lgtm, but i think i'm misunderstanding something about how the commitment to the full wallet shares (public + private) is verified... see my comment

core/src/starknet_client/client.rs
joeykraut: we do not, this will eventually be removed once the contract computes it, but we need to implement non-native poseidon first.

core/src/starknet_client/client.rs
joeykraut: Still too many argument :(

core/src/starknet_client/client.rs
akirillo: ok, this makes sense, but two things:
1) may be helpful to add a TODO / reminder comment so this is easier to find down the line
2) could we, instead, only post the commitment to the public+private shares to the contract, and verify that commitment in the proof (instead of just the private shares commitment)? The tradeoff here is a bit of extra hashing in the proof + cost of verifying that extra hashing, but we need 1 felt less calldata, vs doing non-native hashing in the contract

joeykraut: Yeah we _can_, a "bit of extra hashing" in the proof is a hard sell though. This doubles the amount of hashing done in the circuit, which already makes up > 90% of the constraints.

PR #95, opened 2023-05-12
core, circuits: Link witness elements between `VALID COMMITMENT`, `VALID REBLIND`, `VALID SETTLE`

joeykraut:
### Purpose
This PR adds commitment linking between elements of the proofs used in the match process -- `VALID REBLIND`, `VALID COMMITMENTS`, and `VALID MATCH MPC`, and `VALID SETTLE`.

The linked elements are:
- re-blinded wallet shares between `VALID REBLIND` and `VALID COMMITMENTS`.
- augmented wallet shares between `VALID COMMITMENTS` and `VALID SETTLE`
- balances, orders, and fees between `VALID COMMITMENTS` and `VALID MATCH MPC`

### Testing
- Unit tests pass

### DIFF
--- a/core/src/gossip/orderbook.rs
+++ b/core/src/gossip/orderbook.rs
@@ -1,7 +1,10 @@
 //! Groups handlers for updating and managing order book state in response to
 //! events elsewhere in the local node or the network
 
-use circuits::{types::wallet::Nullifier, verify_singleprover_proof};
+use circuits::{
+    types::wallet::Nullifier, verify_singleprover_proof,
+    zk_circuits::commitment_links::verify_reblind_commitments_link,
+};
 use futures::executor::block_on;
 use libp2p::request_response::ResponseChannel;
 use tracing::log;
@@ -31,6 +34,9 @@ use super::{
 
 /// Error message emitted when an already-used nullifier is received
 const ERR_NULLIFIER_USED: &str = "invalid nullifier, already used";
+/// Error message emitted when two validity proofs are improperly commitment linked
+const ERR_INVALID_PROOF_LINK: &str =
+    "invalid proof link between VALID REBLIND and VALID COMMITMENTS";
 
 impl GossipProtocolExecutor {
     /// Dispatches messages from the cluster regarding order book management
@@ -317,6 +323,15 @@ impl GossipProtocolExecutor {
         // take ownership
         let reblind_proof = proof_bundle.copy_reblind_proof();
         let commitment_proof = proof_bundle.copy_commitment_proof();
+
+        // Verify the two proofs are correctly commitment linked
+        if !verify_reblind_commitments_link(&reblind_proof.commitment, &commitment_proof.commitment)
+        {
+            log::error!("received validity proof bundle with invalid proof linking");
+            return Err(GossipError::ValidCommitmentVerification(
+                ERR_INVALID_PROOF_LINK.to_string(),
+            ));
+        }
 
         // Check that the proof shares' nullifiers are unused
         self.assert_nullifier_unused(reblind_proof.statement.original_shares_nullifier)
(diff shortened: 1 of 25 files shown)


core/src/gossip/orderbook.rs
```diff
@@ -318,6 +324,15 @@ impl GossipProtocolExecutor {
         let reblind_proof = proof_bundle.copy_reblind_proof();
         let commitment_proof = proof_bundle.copy_commitment_proof();
 
+        // Verify the two proofs are correctly commitment linked
+        if !verify_reblind_commitments_link(&reblind_proof.commitment, &commitment_proof.commitment)
```
akirillo, line 328: down the line we should use this for peer scoring / banning

joeykraut: Yeah good call out; we store priorities at both the cluster and order granularity in the global state object [here](https://github.com/renegade-fi/renegade/blob/main/core/src/state/priority.rs#L54-L59). I was thinking we just add hooks to update these values every time a rule like the one you're mentioning occurs

core/src/state/wallet.rs
```diff
@@ -283,7 +285,34 @@ where
 impl From<Wallet> for SizedCircuitWallet {
     fn from(wallet: Wallet) -> Self {
         // Recover the wallet from its secret shares to maintain ordering for witnesses
```
akirillo, line 287: This comment is no longer really true, right? Now we just recover the wallet from the plaintext elements stored on it

To that end, why did we recover from private/public shares previously, if we already store the elements in plaintext?

integration-helpers/src/mpc_network/mocks.rs
```diff
@@ -88,6 +88,14 @@ impl MpcNetwork for MockMpcNet {
...
     }
 
+    async fn send_bytes(&mut self, _bytes: &[u8]) -> Result<(), MpcNetworkError> {
+        Ok(())
+    }
+
+    async fn receive_bytes(&mut self, _num_expected: usize) -> Result<Vec<u8>, MpcNetworkError> {
+        Err(MpcNetworkError::RecvError)
+    }
+
```
akirillo, line 98: Do we use these anywhere?

akirillo (approved):
just one smol nit and lgtm

core/src/gossip/orderbook.rs
akirillo: yeah i like that - definitely doesn't need to be in this PR, though i hope that hooking in to update the global state doesn't introduce too much locking overhead...
it probably won't so this train of thought may be premature optimization, but we could also move all peer priority updates to a separate (low priority) task

integration-helpers/src/mpc_network/mocks.rs
joeykraut: No we don't use these methods of the mock object anywhere. I updated the `MpcNetwork` trait in the `mpc-ristretto` crate separately [here](https://github.com/renegade-fi/mpc-ristretto/commit/521d50c9d339df527597ca302768d78b30a69a07) and we just need to implement dummy implementations in the mock for interface parity. Should've made this more clear

core/src/state/wallet.rs
joeykraut: Yeah thanks, removed the comment. The reason was to preserve ordering of balances, orders, fees; between the output of this method and the on-chain state. 

We need to preserve ordering so that `VALID REBLIND` proves correctly (elements of the reblinded wallet shouldn't be re-ordered). 

The indexed state filters out default orders, balances, etc for API convenience, but this pushes all the non-default orders to the front of the recovered wallet, which doesn't necessarily reflect the in-contract secret share state. Practically, it's simpler to implement `From` this way and just reconstruct the witness to `VALID REBLIND` directly from the known shares when we need ordering.

PR #96, opened 2023-05-12
core: starknet-client, chain-events: Add devnet sequencer support

ChrisBender:
### Purpose

Add support for a `starknet-devnet` sequencer node located at `localhost:5050`. Includes some modifications to event streaming logic to support sync from genesis.


akirillo (approved):
looks like build is failing in CI but it doesn't seem to come from this diff so this lgtm

PR #99, opened 2023-05-18
circuits: zk_circuits, zk_gadgets: add `apply_constraints` method to `SingleProverCircuit` trait

akirillo:
Adds a method `apply_constraints` to the `SingleProverCircuit` trait, which encapsulates the logic of building the actual constraints in the system given the allocated witness and statement variables.

This will be used in the constraint intermediate representation / export tool to build circuit weight matrices in a manner that is generic over the circuit itself

### DIFF
--- a/circuits/src/lib.rs
+++ b/circuits/src/lib.rs
@@ -452,25 +452,15 @@ impl<N: MpcNetwork + Send, S: SharedValueSource<Scalar>> Open<N, S>
 pub trait SingleProverCircuit {
     /// The witness type, given only to the prover, which generates a blinding commitment
     /// that can be given to the verifier
-    type Witness;
+    type Witness: CommitWitness;
     /// The statement type, given to both the prover and verifier, parameterizes the underlying
     /// NP statement being proven
-    type Statement: Clone;
+    type Statement: Clone + CommitPublic;
     /// The data type of the output commitment from the prover.
     ///
     /// The prover commits to the witness and sends this commitment to the verifier, this type
     /// is the structure in which that commitment is sent
     type WitnessCommitment;
-    /// The type of a witness that has been allocated into a constraint system
-    ///
-    /// This is created by both the prover and the verifier in preparation for
-    /// applying constraints
-    type WitnessVar;
-    /// The type of a statement that has been allocated into a constraint system
-    ///
-    /// This is created by both the prover and the verifier in preparation for
-    /// applying constraints
-    type StatementVar;
 
     /// The size of the bulletproof generators that must be allocated
     /// to fully compute a proof or verification of the statement
@@ -482,8 +472,8 @@ pub trait SingleProverCircuit {
     /// Applies all the constraints embodied by the circuit over the allocated witness & statement,
     /// building them into the given constraint system
     fn apply_constraints<CS: RandomizableConstraintSystem>(
-        witness_var: Self::WitnessVar,
-        statement_var: Self::StatementVar,
+        witness_var: <Self::Witness as CommitWitness>::VarType,
+        statement_var: <Self::Statement as CommitPublic>::VarType,
         cs: &mut CS,
     ) -> Result<(), R1CSError>;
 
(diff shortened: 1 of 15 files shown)


circuits/src/lib.rs
```diff
@@ -461,6 +461,16 @@ pub trait SingleProverCircuit {
...
+    /// The type of a witness that has been allocated into a constraint system
+    ///
+    /// This is created by both the prover and the verifier in preparation for
+    /// applying constraints
+    type WitnessVar;
+    /// The type of a statement that has been allocated into a constraint system
+    ///
+    /// This is created by both the prover and the verifier in preparation for
+    /// applying constraints
+    type StatementVar;
```
joeykraut, line 473: I don't think we need to add two more associated types here. I'd suggest a different approach that more closely aligns with the existing traits:
1. Restrict the above types as `type Witness: CommitWitness` and `type Statement: Clone + CommitPublic`
2. In `apply_constraints` we type `witness_var: <Self::Witness as CommitWitness>::VarType` and `statement_var: <Self::Statement as CommitPublic>::VarType`

This more directly expresses the relationship between statement and witness vars, and enforces a trait bound that should really exist anyways.

joeykraut: I'll also make a more generic change if this is done (outside of this PR), to abstract prove/verify using your new `apply_constraints` method

joeykraut (approved):
Lgtm overall, my suggested refactor seems like it might have a somewhat larger blast radius so this is fine for now if you'd prefer not to go through the effort of implementing `CommitWitness` and `CommitPublic` for each type you've defined. We can always get around to that later when we clean this crate up

circuits/src/lib.rs
akirillo: I like this, gonna go ahead and take a stab!

akirillo:
Looking ahead a bit... my plan for exporting the weight vectors is to, in a manner that is generic / agnostic over the underlying circuit, make a dummy prover, witness, and statement, and run `apply_constraints(witness, statement, prover)`. But to make a dummy witness / statement, it looks like I'll need to enforce that `{CommitWitness, CommitStatement}::VarType: Default`. Should I press on here, or do any other approaches come to mind @joeykraut?

joeykraut:
> Looking ahead a bit... my plan for exporting the weight vectors is to, in a manner that is generic / agnostic over the underlying circuit, make a dummy prover, witness, and statement, and run apply_constraints(witness, statement, prover). But to make a dummy witness / statement, it looks like I'll need to enforce that {CommitWitness, CommitStatement}::VarType: Default. Should I press on here, or do any other approaches come to mind @joeykraut?


I don't think we need to implement `Default`. I wouldn't tailor the `circuits` package to fit these scripts, it's not an axis of extensibility we care much about. Instead, why don't we push the "default" behavior to the caller (i.e. the export scripts).

I think more generally, the abstraction boundary of this crate should not be concerned with constructing witnesses and statements (even default), and any caller should have to handle this (i.e. core, your tool, etc)

circuits/src/lib.rs
```diff
@@ -578,6 +587,30 @@ impl CommitPublic for Scalar {
...
+    type ErrorType = ();
+
+    fn commit_public<CS: RandomizableConstraintSystem>(
+        &self,
+        _cs: &mut CS,
+    ) -> Result<Self::VarType, Self::ErrorType> {
+        Ok(())
+    }
+}
+
```
akirillo, line 613: ngl this feels pretty gross, but it felt overkill to make statement types to wrap `bool` and `()` in the places where these are used as statements

akirillo:
Implemented the suggested change, albeit lazily: there are multiple cases where the `VarType` or `CommitType` for a given circuit is a pretty basic type like just a `Variable` or a `Vec<CompressedRistretto>`. In these cases, I didn't create new wrapper types for `{Circuit}StatementVar` or `{Circuit}WitnessCommitment`, etc. (even though if we're being pedantic I should), because of the pending types refactor

circuits/src/zk_gadgets/nonnative.rs
```diff
@@ -1288,14 +1291,51 @@ mod nonnative_tests {
...
+
+            let statement_word_lcs: Vec<LinearCombination> = statement_word_vars
+                .into_iter()
+                .map(Into::into)
+                .collect_vec();
+
+            Ok(NonNativeElementVar::new(statement_word_lcs, self.1.clone()))
+        }
+    }
+
```
akirillo, line 1316: Similarly to what I did in `lib.rs`, I could make a statement type wrapping a `BigUint` and a `FieldMod` (as it seems that `FieldMod` does belong in the statement, no?) but I opted to just keep it as a `(BigUint, FieldMod)` because of the pending refactor

joeykraut (approved):
A small change around how we handle `NonNativeElement` statement types, feel free to merge once that's fixed

PR #100, opened 2023-05-19
circuits: zk_circuits, zk_gadgets: add `apply_constraints` method to `MultiProverCircuit` trait

akirillo:
Similar to #99, this adds an `apply_constraints` method, now to the `MultiProverCircuit` trait, to encapsulate the logic of building the constraint system given the allocated witness (and counterparty witness shares) and statement

### DIFF
--- a/circuits/src/lib.rs
+++ b/circuits/src/lib.rs
@@ -548,11 +548,11 @@ where
     /// Applies all the verifier-side constraints embodied by the circuit over
     /// the entire committed witness & statement,
     /// building them into the given constraint system
-    // fn apply_constraints_single_prover<CS: RandomizableConstraintSystem>(
-    //     witness_var: <Self::Witness as CommitWitness>::VarType,
-    //     statement_var: <Self::Statement as CommitPublic>::VarType,
-    //     cs: &mut CS,
-    // ) -> Result<(), R1CSError>;
+    fn apply_constraints_single_prover<CS: RandomizableConstraintSystem>(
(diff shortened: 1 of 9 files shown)


circuits/src/lib.rs
```diff
@@ -519,6 +535,25 @@ pub trait MultiProverCircuit<'a, N: 'a + MpcNetwork + Send, S: 'a + SharedValueS
...
+        prover: &mut MpcProver<'a, '_, '_, N, S>,
+        fabric: SharedFabric<N, S>,
+    ) -> Result<(), ProverError>;
+
+    /// Applies all the verifier-side constraints embodied by the circuit over
+    /// the entire committed witness & statement,
+    /// building them into the given constraint system
+    fn apply_constraints_single_prover<CS: RandomizableConstraintSystem>(
+        witness_var: <<Self::WitnessCommitment as Open<N, S>>::OpenOutput as CommitVerifier>::VarType,
+        statement: Self::Statement,
```
akirillo, line 553: It's not pretty, but for now I'm just having this method accept a `Self::Statement` parameter. A bit of a leaky abstraction, since the logic this should encapsulate is "given a witness/statement that has been allocated into the constraint system, build the constraints", but now in this method, we first allocate the statement variables and then apply constraints.

The reason for this is that the various statement types don't consistently impl some given trait in a multi-prover context, and given the pending type refactor, I felt it wasn't worthwhile to wrangle all the statement types into a trait implementation.

akirillo:
This will fail to compile until https://github.com/renegade-fi/mpc-bulletproof/pull/2 is merged

akirillo:
this was implemented as part of the types refactor

PR #106, opened 2023-06-01
circuits: zk-circuits: Add midpoint orders to `VALID MATCH MPC`

joeykraut:
### Purpose
This PR adds the initial changes to the `match` MPC and ZK-MPC circuits for midpoint orders. Specifically, the changes are:
- Remove `execution_price` from the `MatchResult` type and take in a price as part of the witness from each party. The parties should agree upon these prices outside of the circuit.
- Remove the computation of the midpoint price in the circuit, the execution price is the agreed upon price.
- Add in checks that the balance covers the matched amount at the executed price.
- Add in checks that the matched amount is less than those specified in the order

### Todo
The following are left for another PR:
- Price protection; i.e. buy side can specify a maximum price, sell side can specify a minimum price
- Integration and unit tests

### Testing
- None; will follow up in another PR for testing specifically
- `circuits` unit tests pass, though CI here will fail to build

### DIFF
--- a/circuits/src/mpc_circuits/match.rs
+++ b/circuits/src/mpc_circuits/match.rs
@@ -1,13 +1,15 @@
 //! Groups logic related to the match computation circuit
 
 use curve25519_dalek::scalar::Scalar;
-use mpc_ristretto::{beaver::SharedValueSource, network::MpcNetwork};
+use mpc_ristretto::{
+    authenticated_scalar::AuthenticatedScalar, beaver::SharedValueSource, network::MpcNetwork,
+};
 
 use crate::{
     errors::MpcError,
     mpc::SharedFabric,
     mpc_gadgets::comparators::min,
-    types::{order::AuthenticatedOrder, r#match::AuthenticatedMatchResult},
+    types::{order::AuthenticatedOrder, r#match::AuthenticatedMatchResult, AMOUNT_BITS},
     zk_gadgets::fixed_point::AuthenticatedFixedPoint,
 };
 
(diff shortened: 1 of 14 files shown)


circuits/src/zk_circuits/valid_match_mpc.rs
```diff
@@ -415,10 +437,18 @@ pub struct ValidMatchMpcWitness {
     pub order1: LinkableOrder,
     /// The first party's balance
     pub balance1: LinkableBalance,
+    /// The price that the first party agreed to execute at for their asset
+    pub price1: FixedPoint,
+    /// The maximum amount that the first party may match
+    pub amount1: Scalar,
```
akirillo, line 443: sorry, I'm a bit confused by the difference between `amount1`/`amount2` and `order1.amount`/`order2.amount`.
the only places where I see `amount1`/`amount2` be used is in the `max_minus_min` checks, which previously used `order1.amount`/`order2.amount`.
but it seems the `amount` field on the `order` struct has not been completely moved out to the `ValidMatchMpcWitness` as I see it still getting used in `validate_volume_constraints`

circuits/src/mpc_circuits/match.rs
```diff
@@ -23,23 +25,22 @@ use crate::{
...
     order2: &AuthenticatedOrder<N, S>,
+    amount1: &AuthenticatedScalar<N, S>,
+    amount2: &AuthenticatedScalar<N, S>,
+    price: &AuthenticatedFixedPoint<N, S>,
     fabric: SharedFabric<N, S>,
 ) -> Result<AuthenticatedMatchResult<N, S>, MpcError> {
     // Compute the amount and execution price that will be swapped if the orders match
     let (min_index, min_base_amount) =
-        min::<32, _, _>(&order1.amount, &order2.amount, fabric.clone())?;
+        min::<32 /* bits */, _, _>(amount1, amount2, fabric.clone())?;
```
akirillo, line 35: nit: should this be constantized?

akirillo (approved):
circuits lgtm but want to double-check some confusion, left a comment

circuits/src/zk_circuits/valid_match_mpc.rs
joeykraut: The `amount` here is referring to a maximum possible match amount with the given, agreed upon price that is constrained either by the order volume or the balance. This change allows orders to be matched with undercapitalized balances. In testing I caught an issue with using `order.amount` still when constraining `max_minus_min_amount` so I'll change that.

PR #108, opened 2023-06-02
circuits: zk-circuits: `VALID MATCH MPC`: Add price protection constraints

joeykraut:
### Purpose
This PR adds price protection to the `Order` struct and enforces the price protection in `VALID MATCH MPC`.

Price protection is implemented as a `worst_case_price` that represents a maximum price for buy side orders and a minimum price for sell side orders.

### Testing
- Unit tests in `circuits` pass, though CI will fail here because `core` has not consumed the midpoint changes
- Integration tests in `circuits` pass

### DIFF
--- a/circuits/integration/mpc_circuits/match.rs
+++ b/circuits/integration/mpc_circuits/match.rs
@@ -2,16 +2,15 @@
 
 use circuits::{
     mpc_circuits::r#match::compute_match,
-    traits::{BaseType, LinkableBaseType, MpcBaseType, MpcType, MultiproverCircuitBaseType},
+    traits::{LinkableBaseType, MpcBaseType, MpcType, MultiproverCircuitBaseType},
     types::{
         balance::Balance,
-        order::{AuthenticatedOrder, Order},
+        order::{AuthenticatedOrder, Order, OrderSide},
         r#match::MatchResult,
     },
     zk_circuits::valid_match_mpc::{AuthenticatedValidMatchMpcWitness, ValidMatchMpcCircuit},
     zk_gadgets::fixed_point::FixedPoint,
 };
-use curve25519_dalek::scalar::Scalar;
 use integration_helpers::types::IntegrationTest;
 use merlin::Transcript;
 use mpc_bulletproof::{r1cs_mpc::MpcProver, PedersenGens};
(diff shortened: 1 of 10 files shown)


circuits/integration/mpc_circuits/match.rs
```diff
@@ -60,56 +59,65 @@ fn test_match_no_match(test_args: &IntegrationTestArgs) -> Result<(), String> {
...
+            },
+            10, /* execution_price */
+        ),
+        // Orders on the same side (buy side)
+        (
+            Order {
+                quote_mint: 1u8.into(),
+                base_mint: 2u8.into(),
+                side: OrderSide::Buy,
+                worst_case_price: FixedPoint::from_integer(15),
```
akirillo, line 93: Should the `worst_case_price` be set differently for each relayer here, i.e. `FixedPoint::from_integer(sel!(5, 15))`?

Just wanna make sure this test fails because of the orders being on the same side and not because of a price protection violation

circuits/integration/mpc_circuits/match.rs
```diff
@@ -60,56 +59,65 @@ fn test_match_no_match(test_args: &IntegrationTestArgs) -> Result<(), String> {
         .map_err(|err| format!("Error allocating balance2 in the network: {:?}", err))?;
 
     // Build the test cases for different invalid match pairs
-    let mut test_cases: Vec<Vec<u64>> = vec![
+    let test_cases: Vec<(Order, u64)> = vec![
```
akirillo, line 62: still not super familiar w/ the difference between `mpc_circuits` and `zk_circuits` testing suites, but should we be adding test cases here that check for price protection violations?

akirillo (approved):
overall lgtm, clean - just wanna confirm a couple smol things

circuits/integration/mpc_circuits/match.rs
joeykraut: This test doesn't interact with this field, plus it makes sense for the worst case prices to be directionally the same since the orders are on the same side of the market.

circuits/integration/mpc_circuits/match.rs
joeykraut: We run a raw MPC before the ZK-MPC that effectively serves to compute the witness for the multiprover ZKP. This test suite tests *only* the raw MPC, so it does not interact with price protection

PR #109, opened 2023-06-03
core: handshake: Refactor handshake process to use midpoint prices

joeykraut:
### Purpose
This PR makes the initial changes to the `handshake::manager` logic to support midpoint prices. Specifically, this PR changes the communication structure such that the proposer proposes a price vector and each party selects the midpoint price of their asset pair. The execution is then done at the midpoint as specified by these vectors.

This PR also makes the smaller, necessary API and state management changes to bring `core` into consistency with the `circuits` changes for midpoints.

### Todo
- Incorporate price feeds from the `price_reporter`

### Testing 
- Ad hoc tested with the current dummy prices from the handshake manager


akirillo (approved):
just one miniscule nit, lfartm

PR #113, opened 2023-06-10
core: handshake: Consumer prices from price reporter for handshake

joeykraut:
### Purpose
This PR integrates the price streams from the `price-reporter` module with the handshake price proposal logic. Specifically, the proposer peeks at the median price for a set of assets, one of which it intends to match on, and proposes these prices to the peer. The peer then peeks their own observed midpoints and validates that the proposed prices are within some tolerance.

Two simplifying assumptions are made in this implementation that will be revisited later:
- We propose a static subset of common asset pairs to mask the desired pair. A better strategy would be to propose a random subset of asset pairs, also including the desired pair, and possibly weighted by trading volume.
- When we validate prices, we only look at max deviation. We'd really like to look at other things as well: for example the deviation on the asset pair we intend to match may need to have a tighter threshold.

### Testing
- Unit and integration tests
- Ad hoc match tests

### DIFF
--- a/core/src/handshake/manager/price_agreement.rs
+++ b/core/src/handshake/manager/price_agreement.rs
@@ -1,0 +1,117 @@
+//! Groups logic for sampling and agreeing upon price vectors during a handshake
+
+use lazy_static::lazy_static;
+use tokio::sync::mpsc::UnboundedSender as TokioSender;
+
+use crate::{
+    handshake::error::HandshakeManagerError,
+    price_reporter::{jobs::PriceReporterManagerJob, tokens::Token},
+};
+
+// ----------------
+// | Quote Tokens |
+// ----------------
+
+/// USDC ticker
+///
+/// For now we only stream prices quoted in USDC
+const USDC_TICKER: &str = "USDC";
+
+// ---------------
+// | Base Tokens |
+// ---------------
+
+/// BTC ticker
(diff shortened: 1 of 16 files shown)


core/src/handshake/manager/price_agreement.rs
```diff
@@ -0,0 +1,228 @@
...
+        if proposed_price.is_none() {
+            return Ok(None);
+        }
+        let (_, _, proposed_price) = proposed_price.unwrap(); /* (base, quote, price) */
+
+        // Validate that the maximum deviation between the proposed prices and the locally
+        // observed prices is within the acceptable range
+        let my_prices: HashMap<(Token, Token), Price> = self.fetch_price_vector().await?.into();
+        let peer_prices: HashMap<(Token, Token), Price> = proposed_prices.clone().into();
+        assert!(my_prices.contains_key(&(base.clone(), quote.clone())));
```
akirillo, line 211: are we sure we want to assert here as opposed to returning a `Result::Err`?

akirillo (approved):
truly sensational stuff 🤝 💼

core/src/handshake/manager/price_agreement.rs
joeykraut: updated

PR #116, opened 2023-06-12
core: tasks: update-wallet: Fix merkle path used in `VALID REBLIND`

joeykraut:
### Purpose
The `update-wallet` task previously was using the old Merkle authentication path when re-proving `VALID REBLIND` (post update). Instead, the task should find the updated wallet's Merkle path and use this in the proof instead.

As well, this PR fixes a bug in which an invalid order validity proof is broadcast where:
1. A node receives the invalid proof and fails to verify it
2. The node then continues requesting and attempting to verify the invalid proof indefinitely
The fix is to index the order in the `Received` state before the proof is verified and only transition to `Verified` after verification is complete. For invalid proofs, this means the order is stuck in the `Received` state until a managing node pushes a valid proof via `UpdateWalletValidityProof`.

### Testing
- Tested with an invalid proof
- Tested the case in which a validity proof is generated out of an `update-wallet` task

### DIFF
--- a/core/src/gossip/orderbook.rs
+++ b/core/src/gossip/orderbook.rs
@@ -22,7 +22,7 @@ use crate::{
         OrderValidityProofBundle, OrderValidityWitnessBundle, SizedValidCommitments,
         SizedValidReblind,
     },
-    state::{NetworkOrder, OrderIdentifier},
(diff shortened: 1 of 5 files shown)


akirillo (approved):
clean, lgtm

PR #117, opened 2023-06-13
core: handshake: price-agreement: Convert token addrs to ETH mainnet

joeykraut:
### Purpose
For now, we specify order asset pairs in terms of their Starknet addresses, but measure prices in terms of Eth mainnet addresses. This may change when we implement a custom bridging scheme, but for now we must translate addresses.

### Testing
- Unit and integration tests
- Ad hoc match tests


akirillo (approved):
makes sense, lgtm

PR #118, opened 2023-06-13
circuits: zk-circuits: valid-match-mpc: Floor div quote amount 

joeykraut:
### Purpose
Previously in `VALID MATCH MPC`, the computed quote amount was constrained to equal the expected quote amount by directly shifting the expected quote amount into a `FixedPoint` representation. This is not valid for non-integral prices as the quote amount has its fractional component truncated during the MPC.

This PR changes the constraint to be such that the difference between the computed quote amount and the expected quote amount (shifted into a `FixedPoint` representation) is less than `DEFAULT_PRECISION` bits in width. This is equivalent to the constraint that the difference is purely fractional; i.e. that the expected amount is the correct integral truncation of the computed amount.

### Testing
- Added an integration test for random non-integral prices in the price protection range
- Ad-hoc tested matches on real price feeds (which are non-integral)

### DIFF
--- a/circuits/integration/zk_circuits/valid_match_mpc.rs
+++ b/circuits/integration/zk_circuits/valid_match_mpc.rs
@@ -29,6 +29,7 @@ use mpc_ristretto::{
     mpc_scalar::scalar_to_u64,
     network::{MpcNetwork, QuicTwoPartyNet},
 };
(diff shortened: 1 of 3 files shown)


akirillo (approved):
lgtm

PR #119, opened 2023-06-13
core: handshake: Better handle zero'd orders after complete fill

joeykraut:
### Purpose
After an order is filled, it is not removed from the secret share state (the settling party may not own the wallet and thus cannot zero filled orders). This PR makes two improvements to how these orders are handled:
- Do not create validity proofs for, gossip about, or index zero-amount orders
- Add a `/wallet/:id/orders/:id/update` endpoint so that a now-zero order can be updated directly to a new order without cancelling and re-creating.

### Testing
- Tested the match flow, verified that the zero'd order was not re-indexed in the network order book
- Tested the update order endpoint, verified that it correctly overwrote the zero'd order

### DIFF
--- a/circuits/src/types/order.rs
+++ b/circuits/src/types/order.rs
@@ -69,6 +69,14 @@ impl Order {
     pub fn is_default(&self) -> bool {
         self.eq(&Self::default())
     }
+
+    /// Whether or not this order is for zero volume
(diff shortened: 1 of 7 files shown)


akirillo (approved):
lgtm ty for explanation

PR #120, opened 2023-06-14
core: tasks: settle-match: Fix bugs preventing chained matches

joeykraut:
### Purpose
This PR fixes a couple bugs that prevent chained matches:
1. In the `settle-match` task, we previously were not properly re-blinding the wallet and recomputing its Merkle inclusion proof. This PR fixes that
2. The `/wallet/:id/orders/:id/update` endpoint needs to re-assign `OrderIdentifier`s to updated orders. This is because stale identifiers are used as match engine cache keys, so if an updated order maintains its identifier, it may never be matched by counterparties that believe they have already matched it.

### Testing
- Tested the end-to-end flow of matching an order, topping off the order on the filled counterparty's side, and then re-matching to fill more of the unfilled counterparty's side.


akirillo (approved):
lgtm

PR #124, opened 2023-06-15
core: price-reporter: properly disable price reporter with flag

joeykraut:
### Purpose
This PR changes the way we disable the price reporter to simply ignore jobs, instead of trying to cancel the worker. Long term we can do something more efficient that deallocates the thread, but for now this is sufficient.

### Testing
- Disabled the price reporter, verified that the rest of the relayer worked properly


akirillo (approved):
sure lgtm

PR #125, opened 2023-06-16
core: config: Add `bind_addr` argument for specific interface binding

joeykraut:
### Purpose
By default we bind to `0.0.0.0` which listens and sends on all network interfaces. However, we sometimes want to specify a specific network interface to send/receive on. 

One such example is the bootstrap cluster; it is provisioned with a static IP so that it may be reached at a known address, but this IP is attached to a non-default network interface on the node. If the bootstrap node binds to all network interfaces, it may receive traffic on the interface of the static IP and send traffic on a different interface. This will be blocked by symmetric NATs as the response will seem to come from a different hostport. For this reason, we bind the bootstrap nodes to only the network interface on which we provision the static IP.

### Testing
- Tested bootstrapping into the network from my local node, which is behind a symmetric NAT (as per a STUN test)

### DIFF
--- a/core/src/config.rs
+++ b/core/src/config.rs
@@ -8,7 +8,7 @@ use serde::{Deserialize, Serialize};
 use std::{
     env::{self},
     fs,
-    net::SocketAddr,
+    net::{IpAddr, SocketAddr},
 };
 use toml::{value::Map, Value};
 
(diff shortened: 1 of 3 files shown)


akirillo (approved):
lgtm, thank u for explaining

PR #126, opened 2023-06-17
core: pricer-reporter: reporter: Use token bucket for exchange failures

joeykraut:
### Purpose
This PR changes the way we handle exchange failures. Specifically, we previously only allowed `MAX_CONN_RETRIES` over the course of a connection's lifetime; now we allow `MAX_CONN_RETRIES` every `MAX_CONN_RETRY_WINDOW_MS` milliseconds. This detects consecutive failures only, not simple recoverable failures that accrue over time.

### Testing
- Tested a few quick failures with Binance


akirillo (approved):
lgtm, good catch

PR #128, opened 2023-06-20
core: config, api-server: Add `disable_fee_validation` flag, and serialize an empty request/response as `null`

ChrisBender:
### Purpose

Adds the `disable_fee_validation` flag, which disables the check that a wallet's fee is nonzero. This is useful for either a testnet or for a private cluster.

### Testing

Tested with both 1) the full testing suite of `renegade-js` version `3.0.1`, and 2) an order placement in the browser.


joeykraut (approved):
Lgtm

PR #129, opened 2023-06-20
core: gossip: Broadcast wallet updates to peers

joeykraut:
### Purpose
We previously relied on cluster join to replicate wallets, but we also need to replicate wallets when a new wallet is created or updated. This PR introduces this by:
1. publish all wallet updates to the system bus
2. consume this on the network manager's side
3. broadcast these wallet updates to peers

I also refactored the network manager to aid discoverability. This constitutes most of the diff, but is essentially just moving code around.

### Testing
- Tested the case in which a new wallet is created on one cluster peer, verified that other cluster peers had indexed it thereafter.


core/src/network_manager/manager/control_directives.rs
```diff
@@ -0,0 +1,192 @@
...
+            ConnectionRole::Dialer => {
+                // Attempt to dial the peer on all known addresses
+                let mut brokered_net = None;
+                for peer_addr in known_peer_addrs
+                    .into_iter()
+                    .filter_map(|addr| multiaddr_to_socketaddr(&addr, peer_port))
+                    .filter(|addr| is_dialable_addr(addr, allow_local))
+                {
+                    let local_addr: SocketAddr =
+                        format!("127.0.0.1:{:?}", local_port).parse().unwrap();
```
akirillo, line 136: maybe a bit out-of-scope, but just want to confirm that this is still the right way to get the local address after the bug fix last week regarding fetching network interfaces?
same goes for the `local_addr` and `peer_addr` in the `ConnectionRole::Listener` match arm below

akirillo (approved):
just a smol q about setting local/peer addresses but lgtm

core/src/network_manager/manager/control_directives.rs
joeykraut: The local addr does not matter for the `Dialer` and the peer_addr does not matter for the `Listener`, I'll change it though so it is clearer to read

PR #131, opened 2023-06-21
core: handshake: internal-engine: Add internal matching engine

joeykraut:
### Purpose
This PR completes the first half of the internal order book feature; leaving settlement for a follow up. The changes are:
- Implement `SingleProverCircuit` for `VALID MATCH MPC`
- Add a job to run the internal matching engine on an order in the handshake manager
- Enqueue this job whenever an order becomes first available for match scheduling

### Todo
- Forward successful matches to the task driver for settlement
- Recursive matches: if we partially fill an order via the internal order book, re-enqueue an `InternalMatchingEngine` job to the handshake manager so that we may try to fill more of the order internally
- We may want to periodically run the `InternalMatchingEngine` job to pick up any orders that get left behind as a result of bugs

### Testing
- Tested the plaintext matching engine thoroughly through unit tests

### DIFF
--- a/circuits/src/zk_circuits/valid_match_mpc.rs
+++ b/circuits/src/zk_circuits/valid_match_mpc.rs
@@ -534,12 +534,12 @@ impl<'a, N: 'a + MpcNetwork + Send, S: SharedValueSource<Scalar>> MultiProverCir
                 <<<Self::Witness as MultiproverCircuitBaseType<N, S>>::MultiproverCommType
                     as MultiproverCircuitCommitmentType<N, S>>::BaseCommitType
                     as CircuitCommitmentType>::VarType,
-        _statement:
+        statement:
                 <<Self::Statement as MultiproverCircuitBaseType<N, S>>::BaseType
                     as CircuitBaseType>::VarType<Variable>,
         cs: &mut CS,
     ) -> Result<(), R1CSError> {
-        ValidMatchMpcSingleProver::matching_engine_check_single_prover(witness, cs)
+        ValidMatchMpcSingleProver::apply_constraints(witness, statement, cs)
     }
 }
 
--- a/circuits/src/zk_circuits/valid_match_mpc.rs
+++ b/circuits/src/zk_circuits/valid_match_mpc.rs
@@ -12,7 +12,7 @@ use crate::{
     traits::{
         BaseType, CircuitBaseType, CircuitCommitmentType, CircuitVarType, LinearCombinationLike,
         MpcBaseType, MpcLinearCombinationLike, MpcType, MultiproverCircuitBaseType,
-        MultiproverCircuitCommitmentType, MultiproverCircuitVariableType,
+        MultiproverCircuitCommitmentType, MultiproverCircuitVariableType, SingleProverCircuit,
     },
     types::{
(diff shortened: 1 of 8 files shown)


circuits/src/zk_circuits/valid_match_mpc.rs
```diff
@@ -536,6 +539,21 @@ impl<'a, N: 'a + MpcNetwork + Send, S: SharedValueSource<Scalar>> MultiProverCir
                     as CircuitBaseType>::VarType<Variable>,
         cs: &mut CS,
     ) -> Result<(), R1CSError> {
-        Self::matching_engine_check_single_prover(witness, cs)
+        ValidMatchMpcSingleProver::matching_engine_check_single_prover(witness, cs)
```
akirillo, line 542: ok giga nit but can we do:
```suggestion
        ValidMatchMpcSingleProver::apply_constraints(witness, statement, cs)
```
just feels like a proper use of the abstraction...

feel free to tell me to stfu

core/src/handshake/manager/internal_engine.rs
```diff
@@ -0,0 +1,270 @@
...
+        // Sample a price to match the order at
+        let (base, quote) = self.token_pair_for_order(my_order);
+        let price = self
+            .fetch_price_vector()
+            .await?
+            .find_pair(&base, &quote)
+            .ok_or_else(|| HandshakeManagerError::NoPriceData(ERR_NO_PRICE_DATA.to_string()))?
+            .2; /* (base, quote, price) */
+
+        // Shuffle the ordering of the other orders for fairness
```
akirillo, line 57: Since we can see the full contents of other locally managed orders, should we maybe order by order volume? So that we can minimize the number of fills?

core/src/handshake/manager/internal_engine.rs
```diff
@@ -0,0 +1,270 @@
...
+        let mut rng = thread_rng();
+        let mut shuffled_indices = (0..other_orders.len()).collect_vec();
+        shuffled_indices.shuffle(&mut rng);
+
+        // Match against each other order in the local book
+        for order_id in shuffled_indices.into_iter().map(|ind| &other_orders[ind]) {
+            // Same order
+            if network_order.id == *order_id {
```
akirillo, line 65: Should we also filter for orders coming from different wallets?

akirillo (approved):
couple non-blocking suggestions, otherwise lgtm

circuits/src/zk_circuits/valid_match_mpc.rs
joeykraut: yeah that's clean

core/src/handshake/manager/internal_engine.rs
joeykraut: good catch

core/src/handshake/manager/internal_engine.rs
joeykraut: It's slightly a pain to do this because of the way the state is setup. It should probably be done eventually I agree, but I'm planning on refactoring the state access at some point as well to support more intricate access patterns like the one you describe. For now, I'd prefer to keep it simple. We will also have a `min_fill_size` field on each order at some point which will aid this issue.

PR #132, opened 2023-06-22
core: tasks: settle-match-internal: Implement initial task steps

joeykraut:
### Purpose
This PR defines the necessary code structure and skeleton to settle a match that was generated by the internal matching engine. This includes:
1. Adding a proof manager job to prove `VALID MATCH MPC` as a single prover
2. Wrangle witness elements when setting up the matching engine
3. Prove this in the first step of the task

As well, this PR adds match volume computation gated by volume to allow for undercapitalized matches.

### Testing


core/src/tasks/settle_match_internal.rs
```diff
@@ -0,0 +1,272 @@
...
+        write!(f, "{self:?}")
+    }
+}
+
+#[async_trait]
+impl Task for SettleMatchInternalTask {
+    type State = SettleMatchInternalTaskState;
+    type Error = SettleMatchInternalTaskError;
+
+    async fn step(&mut self) -> Result<(), Self::Error> {
```
akirillo, line 123: just a q: is this called repeatedly, i.e. poll-style? want to make sure we don't e.g. call `self.prove_match_mpc()` multiple times

core/src/tasks/settle_match_internal.rs
```diff
@@ -0,0 +1,272 @@
...
+            SettleMatchInternalTaskError::EnqueuingJob(ERR_AWAITING_PROOF.to_string())
+        })?;
+        self.valid_match_mpc = Some(proof.into());
+
+        Ok(())
+    }
+
+    /// Prove `VALID SETTLE` on the order pair
+    async fn prove_settle(&self) -> Result<(), SettleMatchInternalTaskError> {
+        todo!()
```
akirillo, line 255: thought for the future: after we successfully create a `VALID SETTLE` proof, we should shoot down in-flight MPCs for both of the matched orders.

or, does this happen automatically when the wallet/order states are updated?

core/src/handshake/manager/internal_engine.rs
```diff
@@ -77,22 +101,73 @@ impl HandshakeExecutor {
                 continue;
             }
 
+            // Lookup the witness used for this order
+            let (other_proof, other_witness) =
```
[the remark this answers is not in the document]
joeykraut: These are not a counterparty's validity proofs, they're generated by the local peer or a cluster peer, so they can be trusted!

core/src/handshake/manager/internal_engine.rs
```diff
@@ -153,15 +234,40 @@ fn match_orders(
...
+        min_amount_order_index: if party0_max_amount <= party1_max_amount {
+            0
+        } else {
+            1
+        },
     }))
 }
 
+/// Compute the maximum matchable amount for an order and balance
+fn compute_max_amount(price: &FixedPoint, order: &Order, balance: &Balance) -> u64 {
```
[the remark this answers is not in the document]
joeykraut: I thought about it, but that's a testing helper and therefore a much less stable interface to depend on. We also can't put it here and import in `circuits` because that would create a circular dependency. I think the long term best solution is to factor all of this out into crates, one of which may expose primitives like a plaintext match.

core/src/tasks/settle_match_internal.rs
joeykraut: Yeah this is a good point, it does happen automatically when the nullifier is spent on-chain, but we can do this optimistically since it's an internal match. I'll add a TODO

core/src/tasks/settle_match_internal.rs
joeykraut: The `TaskDriver` calls this method in a loop, either to retry a failed step or to continue onto the next step. So the only way in which we'd call `prove_match_mpc()` multiple times is when it fails. See [`run_task_to_completion`](https://github.com/renegade-fi/renegade/blob/main/core/src/tasks/driver.rs#L141) for the driver code.

PR #134, opened 2023-06-23
core: tasks: settle-match-internal: Fix execution bugs and recursive match

joeykraut:
### Purpose
This PR fixes a few bugs with the internal matching engine:
- Orders should be nullified during settlement ahead of when their new validity proofs are created. This prevents the internal engine from attempting to double match before new validity proofs are created
- Orders that have zero volume should be ignored by the engine

### Testing
- Tested a recursive internal match in which one buy side order is recursively matched by two separate sell side orders, both of which it fills entirely.


akirillo (approved):
lgtm

PR #135, opened 2023-07-22
core: starknet_client: proof & witness calldata serialization

akirillo:


akirillo:
Closing b/c this has been overwritten in the `starknet-client` crate already

PR #136, opened 2023-07-22
crypto: Refactor implementation over Stark curve

joeykraut:
### Purpose
This PR makes the initial steps to refactor `renegade` over the Stark curve by refactoring the `crypto` crate. This crate defines `crypto` utilities such as:
- Poseidon hashes
- Field conversions
- Elgamal encryptions

### Testing
- Crate tests pass, though CI will be red as the rest of the crates are not refactored

### DIFF
--- a/crypto/benches/elgamal.rs
+++ b/crypto/benches/elgamal.rs
@@ -1,0 +1,54 @@
+use criterion::{criterion_group, criterion_main, BatchSize, BenchmarkId, Criterion, Throughput};
+use crypto::elgamal::{decrypt_scalar, encrypt_scalar};
+use mpc_stark::algebra::scalar::Scalar;
+use rand::thread_rng;
+
+/// Run a benchmark on the ElGamal encryption implementation
+fn bench_encrypt(c: &mut Criterion) {
+    let mut rng = thread_rng();
+
+    let mut group = c.benchmark_group("ElGamal encryption");
+    group.throughput(Throughput::Elements(1));
+
+    group.bench_function(BenchmarkId::from_parameter(1), |b| {
+        b.iter_batched(
+            || {
+                (
+                    Scalar::random(&mut rng),
+                    Scalar::random(&mut rng).to_biguint(),
+                )
+            },
+            |(plaintext, public_key)| encrypt_scalar(plaintext, &public_key),
+            BatchSize::SmallInput,
+        );
+    });
+}
+
+/// Run a decryption benchmark
+fn bench_decrypt(c: &mut Criterion) {
+    let mut rng = thread_rng();
+
+    let mut group = c.benchmark_group("ElGamal decryption");
+    group.throughput(Throughput::Elements(1));
+
+    group.bench_function(BenchmarkId::from_parameter(1), |b| {
+        b.iter_batched(
+            || {
+                let (cipher, _) = encrypt_scalar(
+                    Scalar::random(&mut rng),
+                    &Scalar::random(&mut rng).to_biguint(),
+                );
+                (cipher, Scalar::random(&mut rng).to_biguint())
+            },
+            |(cipher, secret_key)| decrypt_scalar(cipher, &secret_key),
+            BatchSize::SmallInput,
+        );
+    });
+}
+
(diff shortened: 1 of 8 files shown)


crypto/benches/elgamal.rs
```diff
@@ -0,0 +1,54 @@
...
+    group.bench_function(BenchmarkId::from_parameter(1), |b| {
+        b.iter_batched(
+            || {
+                let (cipher, _) = encrypt_scalar(
+                    Scalar::random(&mut rng),
+                    &Scalar::random(&mut rng).to_biguint(),
+                );
+                (cipher, Scalar::random(&mut rng).to_biguint())
+            },
+            |(cipher, secret_key)| decrypt_scalar(cipher, &secret_key),
```
akirillo, line 43: not very familiar w criterion, but wanna make sure that the encryption in this first closure won't affect the benchmarking?

joeykraut: Yes, the criterion interface uses the first closure as setup for this exact reason

crypto/src/constants.rs
```diff
@@ -21,7 +20,7 @@ use crate::fields::DalekRistrettoField;
 ///     python3 calc_round_numbers.py
 /// from the scripts above and taking the output for t = 3, \alpha = 5
 #[memoize]
-pub fn POSEIDON_MDS_MATRIX_T_3() -> Vec<Vec<DalekRistrettoField>> {
+pub fn POSEIDON_MDS_MATRIX_T_3() -> Vec<Vec<Scalar::Field>> {
```
akirillo, line 23: maybe I'm blind, but I tried checking main, v0.1.0, and joey/bp-integration in mpc-stark and I couldn't find a field `Field` on `Scalar` - am I just missing something obvious?

crypto/src/fields.rs
```diff
@@ -56,18 +38,18 @@ pub fn scalar_to_bigdecimal(a: &Scalar) -> BigDecimal {
     BigDecimal::from(bigint)
 }
 
-/// Converts a dalek scalar to an arkworks ff element
-pub fn scalar_to_prime_field(a: &Scalar) -> DalekRistrettoField {
-    Fp256::from(scalar_to_biguint(a))
+pub fn scalar_to_u64(a: &Scalar) -> u64 {
```
akirillo, line 41: nit:
could we call this `reduce_scalar_to_u64`, or leave a doc comment indicating that this takes the 8 least significant bytes of the scalar?

crypto/src/hash.rs
```diff
@@ -19,28 +16,20 @@ use crate::{
...
 /// match the hashes defined in the ZK circuitry
-pub fn poseidon_hash_default_params<T: Into<Vec<u64>>>(val: T) -> DalekRistrettoField {
-    let hashable_values: Vec<u64> = val.into();
-    let arkworks_input = hashable_values
-        .into_iter()
-        .map(DalekRistrettoField::from)
-        .collect_vec();
-
+pub fn poseidon_hash_default_params<T: Into<Scalar>>(val: Vec<T>) -> Scalar {
+    let input = val.into_iter().map(|x| x.into().inner()).collect_vec();
```
akirillo, line 20: not able to find an `inner` method on the `Scalar` type either - is the branch of `mpc-stark` you're using here pushed up to github?

joeykraut: correct, it won't be pushed yet, it'll get merged in as a patch version bump. Not worth creating an analogous PR for every small method I need to add

crypto/src/constants.rs
joeykraut: see below

akirillo (approved):
~approving this but i'm not sure which branch of mpc-stark this uses, maybe im dum~
edit: saw ur comment, lfartm

crypto/src/fields.rs
joeykraut: Added a doc comment, want to keep the method signature the same because it's used a bunch of places. Also added a lint denial for missing docs on this library so that we don't have any more methods without docs.

PR #148, opened 2023-08-07
core: stark_client, chain_events: Rebase onto most recent starknet-rs

ChrisBender:
### Purpose

Rebase our StarkNet client onto the most recent version of `starknet-rs`, as it has upstream changes that are necessary for new RPC specs.

### Testing

* Rust unit tests pass.
* Integration tests with `renegade-js` pass. This particular branch does not pass, but a rebased `testnet` branch on top of these changes works.

### TODO

We use certain deprecated methods, since even though they're "deprecated", there is no equivalent RPC-compliant version that does not directly contact the sequencer. We will remove this once the RPC is up-to-date.

### DIFF
--- a/core/src/starknet_client/client.rs
+++ b/core/src/starknet_client/client.rs
@@ -20,30 +20,25 @@ use crypto::fields::{
 };
 use curve25519_dalek::scalar::Scalar;
 use num_bigint::BigUint;
-use reqwest::Url;
+use reqwest::Url as ReqwestUrl;
 use starknet::{
     accounts::{Account, Call, SingleOwnerAccount},
     core::types::{
-        BlockId as CoreBlockId, FieldElement as StarknetFieldElement, TransactionInfo,
-        TransactionStatus,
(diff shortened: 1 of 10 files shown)


core/src/starknet_client/client.rs
```diff
@@ -159,8 +158,9 @@ impl StarknetClientConfig {
             return None;
         }
 
-        let transport =
-            HttpTransport::new(Url::parse(&self.starknet_json_rpc_addr.clone().unwrap()).ok()?);
+        let transport = HttpTransport::new(
+            ReqwestUrl::parse(&self.starknet_json_rpc_addr.clone().unwrap()).ok()?,
```
joeykraut, line 162: nit: don't need to `clone` here, just do `self.starknet_json_rpc_addr.as_ref().unwrap()`

joeykraut (approved):
lgtm, a couple small things but nothing blocking. Maybe squash the commits together as well

PR #151, opened 2023-08-07
circuits: benches: match: Add `VALID MATCH MPC` benchmarks

joeykraut:
### Purpose
This PR makes the following changes:
- Refactors  the `commitment_links` module over the Stark curve -- this only required a one-line change
- Adds benchmarks for the `match` MPC and the `VALID MATCH MPC` collaborative proof (constraint generation, prover latency, and verification latency)

### Testing
- CI will remain red until the refactor is complete

### DIFF
--- a/circuits/benches/match.rs
+++ b/circuits/benches/match.rs
@@ -1,26 +1,77 @@
 //! Groups integration tests for matching an order and proving `VALID MATCH MPC` collaboratively
+//!
+//! TODO: Benchmark with various simulated latencies
 
-//! Tests the process of proving and verifying a `VALID REBLIND` circuit
-#![allow(incomplete_features)]
 #![feature(generic_const_exprs)]
+#![allow(incomplete_features)]
+
+use std::time::{Duration, Instant};
 
 use circuit_types::{
+    balance::Balance,
     fixed_point::FixedPoint,
     order::Order,
-    traits::{MpcBaseType, MpcType},
+    r#match::LinkableMatchResult,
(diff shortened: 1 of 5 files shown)


circuits/benches/match.rs
```diff
@@ -0,0 +1,247 @@
...
+                // Execute an MPC to generate the constraints
+                let (party0_time, party1_time) = execute_mock_mpc(|fabric| async move {
+                    // Create a witness to the proof
+                    let witness = create_dummy_witness(&fabric);
+
+                    // Create a constraint system to allocate the constraints within
+                    let pc_gens = PedersenGens::default();
+                    let transcript = HashChainTranscript::new(b"test");
+                    let mut prover =
+                        MpcProver::new_with_fabric(fabric.clone(), transcript, pc_gens);
```
akirillo, line 136:
```suggestion
                        MpcProver::new_with_fabric(fabric.clone(), transcript, &pc_gens);
```

akirillo: here and elsewhere - at least, my IDE is saying this needs to be a reference

akirillo (approved):
i think there's a small bug here, but generally lgtm, i like the multiple levels of bench granularity

circuits/benches/match.rs
joeykraut: Run `cargo update` and it should work, that was part of [this change](https://github.com/renegade-fi/mpc-bulletproof/pull/8)

PR #153, opened 2023-08-08
misc: Refactor non-circuits crates to operate over the Stark curve

joeykraut:
### Purpose
This PR migrates the non-circuits crates to the Stark curve. This is largely changing types but involves a few interface changes here and there, as well as changing the `HandshakeManager::match` implementation to more effectively use the executor's concurrency. 

### Todo
- Ad hoc relayer functionality tests
- Regenerate poseidon parameters
- Audit the circuits for places that depend on field bitlength, since this has changed

### Tests
- Unit and integration test pass repo-wide


renegade-crypto/src/fields.rs
```diff
@@ -50,7 +50,8 @@ pub fn scalar_to_u64(a: &Scalar) -> u64 {
 
 /// Converts a dalek scalar to a StarkNet field element
 pub fn scalar_to_starknet_felt(a: &Scalar) -> StarknetFieldElement {
-    StarknetFieldElement::from_byte_slice_be(&a.to_bytes_be()).unwrap()
+    StarknetFieldElement::from_byte_slice_be(&a.to_bytes_be())
+        .expect("`Scalar` field size is smaller than `StarknetFieldElement`")
```
akirillo, line 54:
```suggestion
        .expect("`Scalar` field size is larger than `StarknetFieldElement`")
```

akirillo (approved):
just one tiny nit, lookin clean

PR #155, opened 2023-08-08
core, state: Fix debug TUI

joeykraut:
### Purpose
The debug TUI was broken during the crate refactor because it is gated behind a feature flag that was removed. This PR removes this conditional compilation and simply depends on `cli_args.debug` to switch on/off the TUI mode

### Testing
- Tested the TUI, verified its functionality

### DIFF
--- a/state/src/tui.rs
+++ b/state/src/tui.rs
@@ -1,6 +1,4 @@
 //! Groups definitions of the debug text interface
-#![cfg(feature = "debug-tui")]
-
 use crossterm::{
     event::{self, Event, KeyCode},
     execute,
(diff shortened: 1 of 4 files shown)


state/src/tui.rs
```diff
@@ -9,6 +7,7 @@ use crossterm::{
 
 use futures::executor::block_on;
 use itertools::Itertools;
+use lazy_static::lazy_static;
```
akirillo, line 10: is this used anywhere?

akirillo (approved):
lgtm, potentially unused import tho?

state/src/tui.rs
joeykraut: yes down below, it just wasn't erroring before because the feature flag was disabling its compilation

PR #156, opened 2023-08-09
task-driver: Setup integration test harness

joeykraut:
### Purpose
This PR sets up the testing harness for the `task-driver` integration tests. This involves:
- Generalizing some of the helpers used for integration test setup in the `circuits` crate (the only other crate with formal integration tests)
- Setting up a docker compose style stack, currently just with a copy of the relayer running the tests
- Adding an integration test harness using the harness defined in the `test-helpers` crate.

### Todo
- Add a devnet sequencer to the `docker-compose`
- Add integration tests to the crate

### Testing
- Ran the integration tests, verified that the dummy test passed

### DIFF
--- a/task-driver/docker-compose.yml
+++ b/task-driver/docker-compose.yml
@@ -1,0 +1,14 @@
+services:
+  party0:
+    image: task-driver-integration-test:latest
+    build: .
+    ports:
+      - "8000:8000"
+    command: >
+      cargo test --test integration --
+        --party 0 
+        --port1 8000 
+        --port2 9000 
+        --docker 
+        --verbose
+    tty: true
--- a/task-driver/docker-compose.yml
+++ b/task-driver/docker-compose.yml
@@ -6,9 +6,5 @@ services:
       - "8000:8000"
     command: >
       cargo test --test integration --
-        --party 0 
-        --port1 8000 
-        --port2 9000 
-        --docker 
         --verbose
     tty: true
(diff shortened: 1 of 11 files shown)


task-driver/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
...
+  party0:
+    image: task-driver-integration-test:latest
+    build: .
+    ports:
+      - "8000:8000"
+    command: >
+      cargo test --test integration --
+        --party 0 
+        --port1 8000 
+        --port2 9000
```
akirillo, line 11: does this mean we also need to publish port 9000?

task-driver/integration/main.rs
```diff
@@ -0,0 +1,50 @@
...
+    fn from(_: CliArgs) -> Self {
+        Self
+    }
+}
+
+/// Setup code for the integration tests
+fn setup_integration_tests(_test_args: &CliArgs) {
+    // Configure logging
+    util::logging::setup_system_logger(LevelFilter::Info);
+}
```
akirillo, line 35: just wanna confirm that this is stub code? otherwise seems like it doesn't really do everything it's meant to?

akirillo (approved):
lack a bit of context here so have a couple clarifying qs, but defer to your judgement here and lgtm

task-driver/docker-compose.yml
joeykraut: good catch, copied over the configuration in `circuits` and forgot to change it, updated

task-driver/integration/main.rs
joeykraut: that is correct, I'm writing a follow up that does the interesting things like deploying our contract in the setup

PR #157, opened 2023-08-09
task-driver: integration: Setup devnet sequencer for integration tests

joeykraut:
### Purpose
This PR sets up a devnet sequencer for integration testing the `tasks` crate. Much of this infra is re-usuable and will be used across various crates. The setup is as follows:
1. A docker image the builds our contract sources from `main:HEAD` of `renegade-contracts` so that encapsulating containers may pull pre-built sources in.
2. A docker container that wraps `shardlabs/devnet-sequencer` and installs the `starknet-sierra-compile` binary at the version we use to compile contracts.
3. A docker container that runs the integration tests for the `task-driver` with access to the compiled contract artifacts.

### Testing
- Integration tests against devnet with forked state are passing

### DIFF
--- a/docker/contracts/Dockerfile
+++ b/docker/contracts/Dockerfile
@@ -1,0 +1,24 @@
+# This Dockerfile compiles contract sources into .sierra and .casm artifacts so
+# that they me be consumed by other Dockerfiles in the relayer for e.g. integration testing
+FROM ubuntu AS sources
+
+# Install git and curl
+RUN apt-get update && \
+    apt-get install -y git && \
+    apt-get install -y curl
+
+# Install Scarb
+ARG SCARB_VERSION=0.5.1
+ENV SHELL /bin/bash
+RUN curl --proto '=https' --tlsv1.2 -sSf https://docs.swmansion.com/scarb/install.sh | /bin/sh -s -- -v $SCARB_VERSION 
+ENV PATH="/root/.local/bin:${PATH}"
+
+# Clone the contracts source and build the contracts
+WORKDIR /sources
+RUN git clone https://github.com/renegade-fi/renegade-contracts.git
+
+WORKDIR /sources/renegade-contracts
+RUN scarb build
+
+# Move the compiler artifacts to an output direction
+RUN mv target/dev /artifacts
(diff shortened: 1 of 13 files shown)


docker/contracts/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+# Clone the contracts source and build the contracts
+WORKDIR /sources
+RUN git clone https://github.com/renegade-fi/renegade-contracts.git
+
+WORKDIR /sources/renegade-contracts
+RUN scarb build
+
+# Move the compiler artifacts to an output direction
+RUN mv target/dev /artifacts
```
akirillo, line 24: nit:
```suggestion
RUN scarb -P release build

# Move the compiler artifacts to an output direction
RUN mv target/release /artifacts
```
should be a little faster - for example they don't replace sierra method IDs w/ the function names as defined in source

task-driver/Dockerfile
```diff
@@ -1,31 +1,38 @@
...
+# directory. This means that changes to `task-driver` sources will invalidate the
+# cache for this stage only, as the end state is the same
+FROM alpine AS sources
+COPY . .
+RUN rm -rf task-driver
 
-# Create a build dir and add local dependencies
-WORKDIR /build
+# Add a stage for the contract sources, we pull compiled artifacts out of this stage
+FROM renegade-contracts as contract-sources
```
akirillo, line 10: my docker lingo is a bit rusty 😅 where is the `renegade-contracts` image getting pulled from? just wanna make sure it's using the old version of the contracts for now as you mentioned

test-helpers/src/contracts.rs
akirillo, line 1: confirming that in the integration tests themselves, you'll need sequencer / gateway api access? so far here i'm only seeing json-rpc usage which means u could use katana & all of the utils in starknet-scripts from the contracts repo

akirillo (approved):
lgtm, just a couple qs - am still sad that we can't really use the `starknet-scripts` utilities out of the box 😭

test-helpers/src/contracts.rs
joeykraut: Discussed offline, we will move to Katana or equivalent when:
1. We have migrated the `StarknetClient` to use the JSON-RPC api only -- katana does not support the sequencer API
2. We are ready to use the full fledged Cairo 1.0 contracts in testing -- katana cannot fork goerli state so we would need to deploy new contracts

task-driver/Dockerfile
joeykraut: This gets built and cached locally in the `cargo-integrate` command

PR #158, opened 2023-08-11
state, config: Mock `RelayerState` object for integration tests

joeykraut:
### Purpose
This PR adds a mock for the `RelayerState` object so that it may be injected into integration tests. This also adds a more explicit `Default` behavior to the `RelayerConfig` object.

The mock follows a builder pattern to expose the flags that the `RelayerState` constructor consumes.

### Testing


akirillo (approved):
clean, lgtm

PR #159, opened 2023-08-11
workers: proof-manager: Define mock manager that skips proving

joeykraut:
### Purpose
This PR creates a mock for the `ProofManager` that immediately returns a dummy (invalid) proof. This will be used in integration tests to remove proof generation latency from the test critical path.

### Testing
- Added a test to verify the data path of the mock


akirillo (approved):
This looks great, would love to use these mocks for the constraint-ir utils as well

PR #161, opened 2023-08-12
circuits: benches: valid-wallet-create: Add full sized benchmarks

joeykraut:
### Purpose
This PR adds benchmarks on `VALID WALLET CREATE` prover, verifier, and constraint generation latency on the full sized circuits, with parameters taken from the system wide `constants` crate.

### Testing
- Ran all benchmarks


akirillo (approved):
splendid

PR #163, opened 2023-08-13
circuits: benches: `VALID REBLIND`: Add full sized benchmarks

joeykraut:
### Purpose
This PR adds full sized benchmarks for the `VALID REBLIND` circuit using the constants defined in the `constants` crate.

### Testing
- Ran benchmarks


akirillo (approved):
lgtm

PR #165, opened 2023-08-14
circuits: benches: `VALID SETTLE`: Add full size benchmarks

joeykraut:
### Purpose
This PR adds full sized benchmarks for the `VALID SETTLE` circuit. Sizing constants are taken from the system wide constants defined in the `constants` crate.

### Testing
- Ran all benchmarks


akirillo (approved):
lgtm

PR #166, opened 2023-08-14
test-helpers: mpc-network: mock-with-delay: Add delayed network mock

joeykraut:
### Purpose
This PR adds a `MockNetworkWithDelay` that mocks a non-zero point-to-point latency between nodes. This will be used in MPC benchmarks to simulate various connection latencies.

### Todo
- Possibly add bandwidth as an axis to mock on

### Testing

### DIFF
--- a/test-helpers/src/mpc_network/mock_with_delay.rs
+++ b/test-helpers/src/mpc_network/mock_with_delay.rs
@@ -1,0 +1,216 @@
+//! Defines a network mock that introduces a constant connection delay
+
+use std::{
+    future::Future,
+    pin::Pin,
+    task::{Context, Poll},
+    time::{Duration, Instant},
+};
+
+use async_trait::async_trait;
+use futures::{ready, FutureExt, Sink, Stream};
+use mpc_stark::{
+    error::MpcNetworkError,
+    network::{MpcNetwork, NetworkOutbound, PartyId},
+};
+
+use super::mocks::UnboundedDuplexStream;
+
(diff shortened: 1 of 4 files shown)


test-helpers/src/mpc_network/mock_with_delay.rs
```diff
@@ -0,0 +1,216 @@
...
+        if duration_since_send >= self.delay {
+            Poll::Ready(Some(Ok(message)))
+        } else {
+            self.recv_buffer = Some((instant, message));
+
+            let timer_delay = self.delay - duration_since_send;
+            self.register_timer(timer_delay, cx);
+            Poll::Pending
+        }
+    }
```
akirillo, line 107: I defer to your expertise here, but do we ever actually rely on the `timer` object here? Seems like the "delay" functionality actually comes from the fact that we return `Poll::Pending` if `duration_since_send >= self.delay`, and not from ever `.await`ing the timer (we only poll it once)

akirillo (approved):
don't fully understand use of the `tokio::sleep` timer, but not blocking for merge

test-helpers/src/mpc_network/mock_with_delay.rs
joeykraut: Spoke in person, copying here for transparency:

We effectively defer the context management to the timer. Although we don't directly return anything from the `Sleep::poll` call, this call will internally register the context to be re-polled when the timer deadline is met.

PR #167, opened 2023-08-15
circuits: benches: `match`: Add benchmarks with connection delay

joeykraut:
### Purpose
This PR adds benchmarks for the raw match MPC (generates the shared witness), constraint generation latency, and shared prover latency with a non-zero connection latency. The connection latencies tested are:
- `1ms`: A healthy overestimate of the point-to-point latency within an AWS AZ
- `10ms`: A healthy overestimate of the point-to-point latency within an AWS region but not in the same AZ
- `100ms`: A somewhat realistic estimate on point-to-point latency from a network outside of the AWS backbone, perhaps in a different region on the same continent.

### Testing
- Ran all benchmarks

### DIFF
--- a/circuits/benches/match.rs
+++ b/circuits/benches/match.rs
@@ -24,7 +24,7 @@ use circuits::{
         ValidMatchMpcWitness,
     },
 };
-use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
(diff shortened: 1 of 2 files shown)


akirillo (approved):
lgtm

PR #168, opened 2023-08-15
starknet-client: Remove the use of the sequencer gateway API

joeykraut:
### Purpose
This PR migrates the `StarknetClient` to use only the JSON-RPC api, and not the sequencer gateway. The gateway will be deprecated as Starknet decentralizes, so this change is necessary eventually. As well, this change allows us to test against more recent sequencer versions and implementations, like Katana.

### Testing


akirillo (approved):
lgtm

PR #169, opened 2023-08-16
circuits: benches: Add compile-time flag to enable large benchmarks

joeykraut:
### Purpose
This PR adds the `large_benchmarks` feature flag to the `circuits` crate to conditionally enable the larger sized benchmarks on the circuits. These benchmarks can be very long running and are not very useful for performance analysis (as compared to their smaller counterparts), so they are disabled by default.

### Testing
- Ran all benchmarks to verify the feature flag works

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ Cargo.lock
 .vscode
 *.sw[a-p]
 secrets.env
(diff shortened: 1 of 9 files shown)


akirillo (approved):
lgtm

PR #172, opened 2023-08-21
docker: devnet: Run Katana node with pre-deployed contracts for tests

joeykraut:
### Purpose
This PR sets up a [katana](https://book.dojoengine.org/toolchain/katana/overview.html) node in CI for integration tests to run against. This setup depends on a locally available image with the contract sources compiled within them. The katana sequencer copies the contract sources from this image, deploys them to the sequencer, then dumps the state of the sequencer so that it may be started back up with pre-deployed contracts.

### Testing
- Tested a nullifier check integration test in the `task-driver` integration tests against the sequencer stack

### DIFF
--- a/docker/devnet/setup-sequencer.sh
+++ b/docker/devnet/setup-sequencer.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+
+# Exit on error
+set -e
+
+cd renegade-contracts/starknet_scripts
+export RUST_BACKTRACE=1
+
+# Run a sequencer with state dump enabled
+katana \
+    --seed 0 \
+    --dump-state /katana-state.out \
+    --disable-fee \
+    --chain-id KATANA &
+katana_pid=$!
+sleep 5
+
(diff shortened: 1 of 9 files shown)


docker/devnet/setup-sequencer.sh
```diff
@@ -0,0 +1,31 @@
...
+
+# Deploy and initialize the contracts
+cargo run -- \
+    deploy \
+    --initialize \
+    --dump-deployments \
+    --artifacts-path /artifacts \
+    -c darkpool \
+    --network localhost \
+    --address 0x3ee9e18edc71a6df30ac3aca2e0b02a198fbce19b7480a63a0d71cbd76652e0 \
```
akirillo, line 26: nit: add a comment indicating that this is the predeployed account

akirillo, line 26: nit: add a comment indicating that this is the predeployed account

akirillo (approved):
nice! I love to see the deploy scripts getting used. just one small nit, not blocking

PR #173, opened 2023-08-22
starknet-client: Update client to match Cairo 1.0 contract interface

joeykraut:
### Purpose
This PR updates the `starknet-client` to match the Cairo 1 contracts that are now deployed. This involves:
- Updating the `starknet` dependency to target `0.5` and inheriting all the library changes that come along with that. Notably the removal of the pending block from most interfaces.
- Updating selectors for changed names in the new contracts
- Refactoring the `starknet-client` to aid discoverability now that the code size has grown

### Testing
- Existing tests pass
- I will add integration tests for the client in a follow up

### DIFF
--- a/common/src/types/chain_id.rs
+++ b/common/src/types/chain_id.rs
@@ -3,7 +3,9 @@
 use std::str::FromStr;
 
 use serde::{Deserialize, Serialize};
-use starknet::core::types::FieldElement as StarknetFieldElement;
+use starknet::core::{
+    types::FieldElement as StarknetFieldElement, utils::cairo_short_string_to_felt,
+};
 
 /// Starknet mainnet chain-id
 /// TODO: use `starknet-rs` implementation once we upgrade versions
@@ -34,9 +36,12 @@ pub enum ChainId {
     /// Starknet mainnet
     #[serde(rename = "mainnet")]
     Mainnet,
-    /// Devnet at localhost:5050
+    /// A locally hosted devnet node
     #[serde(rename = "devnet")]
     Devnet,
+    /// A Katana devnet node at `localhost:5050`
+    #[serde(rename = "katana")]
+    Katana,
 }
 
 impl From<ChainId> for StarknetFieldElement {
@@ -45,6 +50,7 @@ impl From<ChainId> for StarknetFieldElement {
             ChainId::AlphaGoerli => STARKNET_TESTNET_ID,
             ChainId::Mainnet => STARKNET_MAINNET_ID,
             ChainId::Devnet => STARKNET_DEVNET_ID,
+            ChainId::Katana => cairo_short_string_to_felt("KATANA").unwrap(),
         }
     }
 }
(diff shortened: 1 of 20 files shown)


common/src/types/chain_id.rs
```diff
@@ -34,9 +36,12 @@ pub enum ChainId {
     /// Starknet mainnet
     #[serde(rename = "mainnet")]
     Mainnet,
-    /// Devnet at localhost:5050
+    /// A locally hosted devnet node
     #[serde(rename = "devnet")]
     Devnet,
+    /// A Katana devnet node at `localhost:5050`
+    #[serde(rename = "katana")]
+    Katana,
```
akirillo, line 44: do we need both of these options?

starknet-client/src/client/contract_interaction.rs
```diff
@@ -0,0 +1,266 @@
...
+impl StarknetClient {
+    // --- Getters ---
+
+    /// Check whether the given Merkle root is valid
+    pub async fn check_merkle_root_valid(
+        &self,
+        root: MerkleRoot,
+    ) -> Result<bool, StarknetClientError> {
+        // TODO: Implement BigUint on the contract
+        let reduced_root = scalar_to_starknet_felt(&root);
```
akirillo, line 43: nit: not sure if this should be called `reduced` here (and elsewhere), since there's no "reduction" from scalar to felt (as the field is smaller)

starknet-client/src/types.rs
```diff
@@ -78,6 +84,28 @@ impl From<ExternalTransfer> for Vec<StarknetFieldElement> {
     }
 }
 
+/// A single party's state validity proof bundle used as input to the matching engine
+/// and forwarded to the contract for verification
+#[derive(Clone)]
+pub struct MatchPayload {
```
akirillo, line 90: nit: would feel nice to match the type definition in the contract, which has the `VALID COMMITMENTS` and `VALID REBLIND` statements as nested struct fields, matching the definitions of the `ValidCommitmentsStatement` and `ValidReblindStatement` types in the relayer, for which I've already implemented `CalldataSerializable` in the contracts e2e tests.

wonder if you have strong opinions on this? makes the code read a bit better and feel more modular

akirillo (approved):
couple nits, nothing blocking

common/src/types/chain_id.rs
joeykraut: We're still using the `devnet-sequencer` elsewhere (i.e. for testing our FE) so we will support it for the time being

starknet-client/src/client/contract_interaction.rs
joeykraut: oops yes forgot to change this after the curve refactor

starknet-client/src/types.rs
joeykraut: The branch we're integrating with does not have this structure, we can add it to the relayer after the changes are made upstream: [see here](https://github.com/renegade-fi/renegade-contracts/blob/cairo2-no-poseidon-or-verifier/src/darkpool/types.cairo#L21)

PR #175, opened 2023-08-23
circuits: mpc-circuits: match: Take commitment linked orders directly

joeykraut:
### Purpose
This PR restructures the `match` MPC circuit to take `AuthenticatedLinkableOrder` types directly instead of the un-linked `AuthenticatedOrder` types. this allows us to avoid re-allocation at MPC time (once for the linkable type and once for the un-linked type) and instead use the same, pre-allocated types across both the `match` MPC and the collaborative proof of `VALID MATCH MPC`.

### Testing


akirillo (approved):
lgtm

PR #177, opened 2023-08-24
rust-toolchain: Bump to `nightly 08.19.23`

joeykraut:
### Purpose
This PR bumps the rust toolchain to the 08/19/23 nightly version to inherit the [`std::io::IsTerminal`](https://doc.rust-lang.org/stable/std/io/trait.IsTerminal.html) changes used upstream in the `anstream` dependency. This also involves making explicit the use of the `inherent_associated_types` feature in some crates already using the feature, and fixing some lints that are enforced it the new toolchain.


akirillo (approved):
lgtm

PR #178, opened 2023-08-24
starknet-client: integration: Add integration test skeleton

joeykraut:
### Purpose
This PR adds the skeleton for the `starknet-client` integration tests. This includes a docker compose setup for compiling the contracts, deploying them to a devnet, and then running the integration tests in a node targeting this devnet. Included is a sample PR for checking a dummy nullifier is not used immediately after the contract is initialized.

### Testing
- Unit and integration tests pass, including the nullifier test added with this PR

### DIFF
--- a/starknet-client/integration/main.rs
+++ b/starknet-client/integration/main.rs
@@ -1,0 +1,127 @@
+//! Defines integration tests for
+#![deny(unsafe_code)]
+#![deny(missing_docs)]
+#![deny(clippy::missing_docs_in_private_items)]
+#![allow(incomplete_features)]
+#![feature(generic_const_exprs)]
+
+use clap::{ArgGroup, Parser};
+use common::types::chain_id::ChainId;
+use eyre::Result;
+use mpc_stark::algebra::scalar::Scalar;
+use rand::thread_rng;
+use starknet_client::client::{StarknetClient, StarknetClientConfig};
+use test_helpers::{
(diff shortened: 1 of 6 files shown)


starknet-client/integration/main.rs
```diff
@@ -0,0 +1,128 @@
...
+    starknet_account_addr: String,
+    /// The address of the darkpool deployed on Starknet at the time the test is started
+    ///
+    /// If not provided, the test will deploy a new darkpool contract
+    #[arg(long)]
+    darkpool_addr: Option<String>,
+    /// The location of a `deployments.json` file that contains the addresses of the
+    /// deployed contracts
+    #[arg(long)]
+    deployments_path: Option<String>,
```
akirillo, line 57: do we need both of these arguments? i would assume that if the darkpool is deployed, it would be included in `deployments.json`

akirillo: i see below that we expect one of the two of these to be set.
1) we could use the clap `group` attribute to ensure exactly one is set, if this is the desired behavior
2) this is confusing to me because the doc comment here seems to imply that its okay for both of these flags to be unset, as if this is the case then the test will deploy the darkpool contract

akirillo (approved):
nothing blocking, but a bit confused about test args

starknet-client/integration/main.rs
joeykraut: I'll use clap group and update the doc comment,  copied these args from old tests

PR #179, opened 2023-08-25
starknet-client: Add chain state integration tests

joeykraut:
### Purpose
This PR adds integration tests for:
- Fining a wallet commitment in the contract state
- Finding a Merkle path in the contract state
- Recovering wallet shares from a transaction

This PR also restructures the way in which calldata is parsed from `InvokeV1` transactions.

### Testing

### DIFF
--- a/starknet-client/src/helpers.rs
+++ b/starknet-client/src/helpers.rs
@@ -2,19 +2,14 @@
 
 use std::convert::TryInto;
 
-use circuit_types::SizedWalletShare;
-use serde::de::DeserializeOwned;
+use circuit_types::{traits::BaseType, SizedWalletShare};
+use renegade_crypto::fields::{starknet_felt_to_scalar, starknet_felt_to_usize};
 use starknet::core::types::FieldElement as StarknetFieldElement;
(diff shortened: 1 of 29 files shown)


starknet-client/src/helpers.rs
```diff
@@ -29,11 +24,97 @@ const UPDATE_WALLET_EXTERNAL_TRANSFER_LEN: usize = 4;
...
+
+    let (selector, data_offset, data_len) = find_first_darkpool_transaction(
+        &tx_calldata[metadata_start..metadata_end],
+        darkpool_contract_addr,
+    )?;
+
+    // Parse the calldata of the transaction
+    let calldata_start = metadata_end;
+    let calldata = tx_calldata[calldata_start..].to_vec();
+    let tx_calldata = calldata[data_offset..(data_offset + data_len)].to_vec();
```
akirillo, line 82: nit: bouncing between `let calldata = tx_calldata[...]` and `let tx_calldata = calldata[...]` is a bit tough to read, could we rename the second var 😅

PR #180, opened 2023-08-25
starknet-client: integration: Add contract interaction integration tests

joeykraut:
### Purpose
This PR adds integration test cases for:
- Checking Merkle root validity
- Checking nullifier use
- Submitting a `new_wallet` tx
- Submitting an `update_wallet` tx
- Submitting a `process_match` tx

As well, this PR makes the following changes to fix bugs caught by the tests:
- Fix calldata parsing bugs for secret share recovery
- Add a `CalldataDeserializable` trait and implement it on basic calldata types

### Testing

### DIFF
--- a/starknet-client/integration/contract_interaction.rs
+++ b/starknet-client/integration/contract_interaction.rs
@@ -1,0 +1,178 @@
+//! Defines integration tests for `StarknetClient` methods that interact directly
+//! with darkpool contract methods
+
+use circuit_types::{
+    native_helpers::compute_wallet_share_commitment, traits::BaseType, SizedWalletShare,
+};
+use eyre::Result;
+use mpc_stark::algebra::scalar::Scalar;
+use rand::thread_rng;
+use test_helpers::{assert_eq_result, assert_true_result, integration_test_async};
+
+use crate::{
+    helpers::{
+        deploy_new_wallet, dummy_valid_match_mpc_bundle, dummy_valid_settle_bundle,
+        dummy_valid_wallet_update_bundle, dummy_validity_proof_bundle, dummy_wallet_share,
+    },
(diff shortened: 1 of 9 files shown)


starknet-client/integration/contract_interaction.rs
```diff
@@ -0,0 +1,176 @@
...
+        .await?;
+    client.poll_transaction_completed(tx_hash).await?;
+
+    // Check that the nullifier is used
+    let nullifier_unused = client.check_nullifier_unused(nullifier).await?;
+    assert_true_result!(!nullifier_unused)?;
+
+    // Find the transaction that updated the wallet
+    let tx_hash = client
+        .get_public_blinder_tx(public_shares.blinder)
```
akirillo, line 111: wouldn't the transaction that updated the wallet be indexed by `new_shares.blinder`?
and subsequently, we should check that `recovered_public_shares == new_shares`, right

starknet-client/integration/contract_interaction.rs
```diff
@@ -0,0 +1,176 @@
...
+
+    // Deploy a new wallet
+    let (commitment, public_shares) = deploy_new_wallet(client).await?;
+    let nullifier = Scalar::random(&mut rng);
+    let new_shares = dummy_wallet_share();
+
+    // Update the wallet
+    let tx_hash = client
+        .update_wallet(
+            commitment,
```
akirillo, line 96: nit: although this doesn't affect the correctness of the test, shouldn't this commitment be the commitment to `new_shares`? using that here and in `test_match` would help with the test reading a bit more as expected

akirillo (approved):
one test correctness change and one small nit, will approve so that u can merge after correcting

PR #181, opened 2023-08-26
task-driver: integration: lookup-wallet: Add integration tests

joeykraut:
### Purpose
This PR adds integration tests for the `LookupWalletTask` including:
- An invalid lookup on a wallet that does not exist in contract state
- A valid lookup on a wallet that has has a variable number of `update_wallet` transactions applied to it

### Testing


common/src/types/wallet.rs
```diff
@@ -216,7 +216,7 @@ impl Wallet {
         new_private_shares.push(new_blinder_private_share);
 
         let (new_private_share, new_public_share) = create_wallet_shares_from_private(
-            self.clone().into(),
+            &self.clone().into(),
```
akirillo, line 219: nit: can we remove the `clone().into()` here?

renegade-crypto/src/hash.rs
```diff
@@ -12,6 +12,31 @@ use crate::constants::{POSEIDON_MDS_MATRIX_T_3, POSEIDON_ROUND_CONSTANTS_T_3};
 /// A type alias for the arkworks Poseidon sponge over the Stark field
 pub type PoseidonParams = PoseidonConfig<Scalar::Field>;
 
+/// A hash chain from a seed used to compute CSPRNG values
+pub struct PoseidonCSPRNG {
+    /// The seed of the CSPRNG, this is chained into a hash function
+    /// to give pseudorandom values
+    seed: Scalar,
```
akirillo, line 19: nit: can we call this `state`, and update the doc comment to mention that it is initially set to the seed? makes the iterator functionality read a bit more like its expected to

task-driver/integration/helpers.rs
```diff
@@ -1,9 +1,62 @@
...
 use system_bus::SystemBus;
-use task_driver::driver::TaskDriver;
+use task_driver::driver::{TaskDriver, TaskDriverConfig};
+
+// ------------------------
+// | Contract Interaction |
+// ------------------------
+
+/// Create a wallet in the contract state
+pub async fn allocate_wallet_in_darkpool(wallet: &Wallet, client: &StarknetClient) -> Result<()> {
```
akirillo, line 25: thoughts on consolidating this with `deploy_new_wallet` in `starknet-client/integration/helpers.rs`? don't have a strong opinion here, just seems like these two helpers do allllmost the same thing

task-driver/integration/main.rs
```diff
@@ -84,6 +84,13 @@ struct CliArgs {
 struct IntegrationTestArgs {
     /// The starknet client that resolves to a locally running devnet node
     starknet_client: StarknetClient,
+    /// A sender to the network manager's work queue
+    network_sender: TokioSender<GossipOutbound>,
+    /// A receiver for the network manager's work queue, held here to avoid dropping
+    /// the channel
+    ///
+    /// Held here to avoid closing the channel on `Drop`
```
akirillo, line 92: nit: last line in doc comment here is redundant

akirillo (approved):
a couple vain nits, nothing blocking

common/src/types/wallet.rs
joeykraut: No, the type conversion is necessary and `Into::into` needs ownership

task-driver/integration/helpers.rs
joeykraut: Their implied interface is different enough for them to be separate; I think combining them would be over-optimizing code-reuse at the expense of complicating the interaction with the helper in both cases.

PR #183, opened 2023-08-26
task-driver: integration: update-wallet: Add integration tests

joeykraut:
### Purpose
This PR adds initial integration tests for the `UpdateWalletTask`. More tests will be added but for brevity, the scope of this PR is limited to:
- Setup and common helper code
- A simple re-blinding update
- An update that places a new order
In each case, wallet recovery is tested after the task completes.

### Testing


task-driver/integration/helpers.rs
```diff
@@ -9,13 +9,66 @@ use common::types::{
...
+
+/// Lookup a wallet in the contract state and verify that it matches the expected wallet
+pub(crate) async fn lookup_wallet_and_check_result(
+    expected_wallet: &Wallet,
+    blinder_seed: Scalar,
+    share_seed: Scalar,
+    test_args: IntegrationTestArgs,
+) -> Result<()> {
+    // Start a lookup task for the new wallet
+    let new_wallet_id = Uuid::new_v4();
```
akirillo, line 39: shouldn't we be using the `expected_wallet.id` to look it up?

akirillo (approved):
(i think) there needs to be a small change in the `lookup_wallet_and_check_result` helper, o/w lgtm

task-driver/integration/helpers.rs
joeykraut: No, this ID doesn't matter and doing it this way makes sure that we don't overwrite any state elements on lookup -- which is desired if the caller allocated state elements.

PR #184, opened 2023-08-26
common: types: proof-bundles: Allocate all proof bundles on heap

joeykraut:
### Purpose
This PR changes the runtime representation of all proof types to be heap-allocated. This avoids stack overflows (especially in Tokio contexts) and crowding the stack with infrequently accessed values.

### Testing

### DIFF
--- a/common/src/types/proof_bundles.rs
+++ b/common/src/types/proof_bundles.rs
@@ -1,7 +1,5 @@
 //! Defines proof bundles that are passed across worker boundaries
 
-use std::sync::Arc;
-
 use circuits::zk_circuits::{
     valid_commitments::{
         SizedValidCommitmentsWitness, ValidCommitmentsStatement, ValidCommitmentsWitnessCommitment,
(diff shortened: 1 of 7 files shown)


common/src/types/proof_bundles.rs
```diff
@@ -107,11 +105,12 @@ pub struct GenericValidCommitmentsBundle<
...
 
 /// A type alias that specifies the default generics for `GenericValidCommitmentsBundle`
-pub type ValidCommitmentsBundle = GenericValidCommitmentsBundle<MAX_BALANCES, MAX_ORDERS, MAX_FEES>;
+pub type ValidCommitmentsBundle =
+    Box<GenericValidCommitmentsBundle<MAX_BALANCES, MAX_ORDERS, MAX_FEES>>;
 
 /// A bundle of the statement, witness commitment, and proof of `VALID MATCH MPC`
 #[derive(Clone, Debug, Serialize, Deserialize)]
-pub struct ValidMatchMpcBundle {
+pub struct GenericValidMatchMpcBundle {
```
akirillo, line 113: nit, but should this be prefixed `Generic` when it doesn't take any generic parameters?

akirillo (approved):
tiny nit, lgtm

common/src/types/proof_bundles.rs
joeykraut: what name would read better? I went with `Generic` to maintain consistency in the relationship between this and analogous types

PR #186, opened 2023-08-28
starknet-client: pathfinder: Use pathfinder API to poll transaction finality

joeykraut:
### Purpose
This PR changes the `starknet-client` to the use the `pathfinder_getTransactionStatus` API to poll transaction finality instead of the `starknet_getTransactionReceipt` API. This API is more stable, and more consistently reports finality information. As well, we do not need any element on a transaction's receipt except its finality status.

### Testing
- Integration tests for `starknet-client` pass
- Integration tests for `task-driver` pass

### DIFF
--- a/starknet-client/src/client/chain_state.rs
+++ b/starknet-client/src/client/chain_state.rs
@@ -6,7 +6,10 @@ use std::{
 };
 
 use circuit_types::SizedWalletShare;
-use common::types::merkle::{MerkleAuthenticationPath, MerkleTreeCoords};
+use common::types::{
+    chain_id::ChainId,
+    merkle::{MerkleAuthenticationPath, MerkleTreeCoords},
+};
 use constants::MERKLE_HEIGHT;
 use mpc_stark::algebra::scalar::Scalar;
 use num_bigint::BigUint;
@@ -18,8 +21,7 @@ use starknet::{
     accounts::{Account, Call},
     core::types::{
         BlockId, BlockTag, EmittedEvent, EventFilter, FieldElement as StarknetFieldElement,
-        FunctionCall, InvokeTransaction, MaybePendingTransactionReceipt, Transaction,
-        TransactionReceipt,
+        FunctionCall, InvokeTransaction, Transaction,
     },
     providers::Provider,
 };
@@ -34,8 +36,8 @@ use crate::{
 };
 
 use super::{
-    StarknetClient, TransactionHash, BLOCK_PAGINATION_WINDOW, ERR_UNEXPECTED_TX_TYPE,
-    MAX_FEE_MULTIPLIER, TX_STATUS_POLL_INTERVAL_MS,
+    pathfinder::TransactionStatus, StarknetClient, TransactionHash, BLOCK_PAGINATION_WINDOW,
+    ERR_UNEXPECTED_TX_TYPE, MAX_FEE_MULTIPLIER, TX_STATUS_POLL_INTERVAL_MS,
 };
 impl StarknetClient {
     /// Helper to make a view call to the contract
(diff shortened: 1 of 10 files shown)


starknet-client/src/client/chain_state.rs
```diff
@@ -105,25 +107,24 @@ impl StarknetClient {
     pub async fn poll_transaction_completed(
         &self,
         tx_hash: StarknetFieldElement,
-    ) -> Result<TransactionReceipt, StarknetClientError> {
+    ) -> Result<TransactionStatus, StarknetClientError> {
+        // On devnet, we do not generally have access to the pathfinder API, and transaction
+        // finality is essentially instant, so we skip the polling
+        match self.config.chain {
```
akirillo, line 113: this looks fine to me, but can we add a debug log saying that polling is being skipped on devnet?

starknet-client/src/client/chain_state.rs
```diff
@@ -235,14 +236,16 @@ impl StarknetClient {
...
+        let start_block_limit = current_block.saturating_sub(BLOCK_PAGINATION_WINDOW as u64);
+        'outer: for start_block in (earliest_block..=start_block_limit)
             .rev()
             .step_by(BLOCK_PAGINATION_WINDOW)
         {
             // Exhaust events from the start block to the end block
-            let start_block = end_block.saturating_sub(BLOCK_PAGINATION_WINDOW as u64);
             filter.from_block = Some(BlockId::Number(start_block));
-            filter.to_block = Some(BlockId::Number(end_block));
+            filter.to_block = Some(end_block);
```
akirillo, line 248: on the first iteration of this loop, we may fetch more than `BLOCK_PAGINATION_WINDOW` blocks worth of events. we have `from_block = current_block - BLOCK_PAGINATION_WINDOW` and `end_block = LATEST_BLOCK`. is this expected? feels like we should be initializing `end_block = current_block`?

starknet-client/src/client/pathfinder.rs
```diff
@@ -0,0 +1,148 @@
...
+// | Client Implementation |
+// -------------------------
+
+impl StarknetClient {
+    /// Get the status of a transaction from the pathfinder API
+    pub async fn get_tx_status(
+        &self,
+        tx_hash: FieldElement,
+    ) -> Result<TransactionStatus, StarknetClientError> {
+        let tx_hash_str = format!("0x{tx_hash:x}");
```
akirillo, line 114: i know tests pass, but still want to confirm that the Pathfinder API doesn't expect a specific padding on the tx hash?

starknet-client/src/client/pathfinder.rs
```diff
@@ -0,0 +1,148 @@
...
+        &self,
+        method: String,
+        params: Vec<String>,
+    ) -> Result<T, StarknetClientError> {
+        let res = self
+            .http_client
+            .post(self.config.starknet_json_rpc_addr.clone())
+            .json(&RawJsonRpcRequest {
+                jsonrpc: JSON_RPC_VERSION.to_string(),
+                id: 1,
```
akirillo, line 130: should we make the `id` an incrementing counter?

task-driver/src/settle_match_internal.rs
```diff
@@ -385,9 +385,19 @@ impl SettleMatchInternalTask {
...
-        log::info!("tx hash: 0x{tx_hash:x}");
 
-        // TODO: Fix the polling method and await finality here
+        log::info!("tx hash: 0x{tx_hash:x}");
+        let status = self
+            .starknet_client
+            .poll_transaction_completed(tx_hash)
+            .await
+            .map_err(|err| SettleMatchInternalTaskError::Starknet(err.to_string()))?;
+        if status == TransactionStatus::Reverted || status == TransactionStatus::Aborted {
```
akirillo, line 395: could we use `status.into_result().map_err(...)` here?

task-driver/src/settle_match.rs
```diff
@@ -304,8 +304,18 @@ impl SettleMatchTask {
...
             tx_submit_res.map_err(|err| SettleMatchTaskError::StarknetClient(err.to_string()))?;
 
-        // TODO: Fix the polling method and await finality here
         log::info!("tx hash: 0x{:x}", starknet_felt_to_biguint(&tx_hash));
+        let status = self
+            .starknet_client
+            .poll_transaction_completed(tx_hash)
+            .await
+            .map_err(|err| SettleMatchTaskError::StarknetClient(err.to_string()))?;
+        if status == TransactionStatus::Reverted || status == TransactionStatus::Aborted {
```
akirillo, line 313: here as well, could we use `status.into_result().map_err(...)`?

akirillo (approved):
couple questions / suggestions

starknet-client/src/client/chain_state.rs
joeykraut: Yes this is expected, we don't particularly care if we're off by a block, recall that we paginate within this window anyways.

starknet-client/src/client/pathfinder.rs
joeykraut: correct

starknet-client/src/client/pathfinder.rs
joeykraut: These are synchronous calls, I don't see what we gain by maintaining more state?

task-driver/src/settle_match_internal.rs
joeykraut: No, we do not want to error on `Rejected` in the match tasks. The transaction rejects for one of the parties if both try to settle, this case is handled above.

PR #187, opened 2023-08-29
circuit-types: replaced signing key types with StarkPoint & Scalar

akirillo:
This PR updates the `PublicSigningKey` type to be a representation of a `StarkPoint` (namely, the `Scalar`s needed to represent a point's affine coordinates), and the `SecretSigningKey` type to be a `Scalar`, as expected for ECDSA over the STARK curve in accordance with the [IETF spec](datatracker.ietf.org/doc/html/rfc6979).

This PR also stubs out some changes to the API, namely it temporarily removes request signature verification logic, pending a future PR where STARK ECDSA is implemented.

All unit tests pass (including an updated serialization / deserialization test for the `PublicSigningKey`), and the affected integration tests (those in the `circuits` crate) pass.

### DIFF
--- a/circuit-types/src/keychain.rs
+++ b/circuit-types/src/keychain.rs
@@ -3,8 +3,10 @@
 
 use std::ops::Add;
 
+use ark_ec::models::short_weierstrass::Affine;
+use ark_ff::fields::PrimeField;
 use circuit_macros::circuit_type;
-use ed25519_dalek::PublicKey as DalekKey;
+use lazy_static::lazy_static;
 use mpc_bulletproof::r1cs::{LinearCombination, Variable};
 use mpc_stark::{
     algebra::{
@@ -16,7 +18,6 @@ use mpc_stark::{
 };
 use num_bigint::BigUint;
 use rand::{CryptoRng, RngCore};
-use renegade_crypto::fields::get_scalar_field_modulus;
 use serde::{Deserialize, Serialize};
 
 use crate::{
@@ -29,14 +30,15 @@ use crate::{
     },
 };
 
-use super::{biguint_from_hex_string, biguint_to_hex_string};
-
 /// The number of keys held in a wallet's keychain
 pub const NUM_KEYS: usize = 4;
-/// The number of bytes used in a single scalar to represent a key
-pub const SCALAR_MAX_BYTES: usize = 31;
-/// The number of words needed to represent a non-native root key
-pub const ROOT_KEY_WORDS: usize = 2;
+/// The number of scalar words needed to represent STARK base field element
+pub const FELT_WORDS: usize = 2;
+
+lazy_static! {
+    /// The bit mask for the lower 128 bits of a U256
+    static ref LOW_BIT_MASK: BigUint = (BigUint::from(1u8) << 128) - 1u8;
+}
 
 // -------------
 // | Key Types |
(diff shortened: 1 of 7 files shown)


circuit-types/src/keychain.rs
```diff
@@ -29,14 +30,15 @@ use crate::{
...
-use super::{biguint_from_hex_string, biguint_to_hex_string};
-
 /// The number of keys held in a wallet's keychain
 pub const NUM_KEYS: usize = 4;
-/// The number of bytes used in a single scalar to represent a key
-pub const SCALAR_MAX_BYTES: usize = 31;
-/// The number of words needed to represent a non-native root key
-pub const ROOT_KEY_WORDS: usize = 2;
+/// The number of scalar words needed to represent STARK base field element
+pub const FELT_WORDS: usize = 2;
```
joeykraut, line 36: maybe `FELT_WORDS_PER_BASE`? "felt words" alone does not describe a relation

circuit-types/src/keychain.rs
```diff
@@ -29,14 +30,15 @@ use crate::{
...
 
-use super::{biguint_from_hex_string, biguint_to_hex_string};
-
 /// The number of keys held in a wallet's keychain
 pub const NUM_KEYS: usize = 4;
-/// The number of bytes used in a single scalar to represent a key
-pub const SCALAR_MAX_BYTES: usize = 31;
-/// The number of words needed to represent a non-native root key
-pub const ROOT_KEY_WORDS: usize = 2;
+/// The number of scalar words needed to represent STARK base field element
```
joeykraut, line 35: nit: `STARK` -> `STARK curve` and elsewhere

circuit-types/src/keychain.rs
```diff
@@ -129,105 +140,63 @@ impl From<SecretIdentificationKey> for Scalar {
...
+impl PublicSigningKey {
+    /// Split a biguint into 128-bit scalar words
+    fn split_biguint_into_words(val: &BigUint) -> [Scalar; FELT_WORDS] {
+        let low_mask = val & &*LOW_BIT_MASK;
+        let low: u128 = low_mask.try_into().unwrap();
+        let high: u128 = (val >> 128u32).try_into().unwrap();
 
-        res.try_into().unwrap()
+        [Scalar::from(low), Scalar::from(high)]
     }
```
joeykraut, line 157: A couple of suggestions on improving this method:
1. It's much cleaner to define a modulus and use `div_rem`
2. We don't need to go through `u128` to get to `Scalar`
```rust
use renegade_crypto::fields::biguint_to_scalar;

lazy_static {
     static ref TWO_TO_128: BigUint = BigUint::from(1u8) << 128;
}

fn split_biguint_into_words(val: &BigUint) -> [Scalar; FELT_WORDS] {
    let (high, low) = BigUint::div_rem(val, &TWO_TO_128);

    [
        biguint_to_scalar(&low),
        biguint_to_scalar(&high),
    ]
}
```

circuit-types/src/keychain.rs
```diff
@@ -129,105 +140,63 @@ impl From<SecretIdentificationKey> for Scalar {
...
-    }
-}
+impl From<StarkPoint> for PublicSigningKey {
+    fn from(value: StarkPoint) -> Self {
+        let Affine { x, y, infinity } = value.to_affine();
+        if infinity {
+            Self {
+                x: [Scalar::zero(); 2],
+                y: [Scalar::zero(); 2],
+            }
```
joeykraut, line 175: I believe the IETF spec disallows the additive identity as the public key? If so, we should `assert` here

joeykraut (approved):

circuit-types/src/keychain.rs
akirillo: oh yes great call

PR #192, opened 2023-08-30
renegade-crypto: ecdsa: reverse keccak byte output

akirillo:
This PR fixes the reduction of a keccak hash output to a scalar in the `ecdsa` module to match the Cairo keccak implementation, which expects little-endian `u256`s as input but returns a big-endian `u256`.

All of the `ecdsa` unit tests pass.


joeykraut (approved):
looks good, let's add a line to the doc comment explaining that the input is expected to be little endian (considering it's an unstructured byte array)

akirillo:
> looks good, let's add a line to the doc comment explaining that the input is expected to be little endian (considering it's an unstructured byte array)

we already have a doc comment explaining this in the `compute_message_hash` function, here to be clear we're reversing the _output_ of the keccak hash as well so wouldn't make sense to call the unstructured byte input to keccak "little-endian"

PR #195, opened 2023-09-10
statev2: storage: db: Initialize database interface

joeykraut:
### Purpose
This PR beings the storage refactor by exposing a simple key-value interface over an `mdbx` embedded instance. Currently, this interface only exposes `set` and `get` methods, but we will add transactional semantics in a follow up PR

### Testing
- Simple unit tests pass

### DIFF
--- a/statev2/Cargo.toml
+++ b/statev2/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "statev2"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+
+# === Storage === #
+flexbuffers = "2.0"
+libmdbx = "0.3"
+serde = { workspace = true, features = ["derive"] }
+
+[dev-dependencies]
+tempfile = "3.8"
--- a/statev2/src/storage/db.rs
+++ b/statev2/src/storage/db.rs
@@ -1,0 +1,206 @@
+//! Defines the interface to the `mdbx` instance
+//!
+//! We serialize types using the `flexbuffers` format (a schema-less version of
+//! `flatbuffers`): https://flatbuffers.dev/flexbuffers.html
+
+use std::{borrow::Cow, path::Path};
+
+use libmdbx::{Database, TableFlags, Transaction, WriteFlags, WriteMap, RO, RW};
+use serde::{Deserialize, Serialize};
+
+use super::{
+    error::StorageError,
+    traits::{Key, Value},
+};
+
+/// The number of tables to open in the database
+const NUM_TABLES: usize = 2;
+
+/// The database config
+pub struct DbConfig {
+    /// The path to open the database at
+    path: String,
+}
+
+/// The persistent storage layer for the relayer's state machine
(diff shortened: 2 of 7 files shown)


statev2/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "statev2"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+
+# === Storage === #
+flexbuffers = "2.0"
+libmdbx = "0.3"
```
akirillo, line 10: idk if it's relevant to us, but reth forked an earlier version of this crate b/c it switched to GPL license: https://github.com/paradigmxyz/reth/tree/main/crates/storage/libmdbx-rs

statev2/src/storage/db.rs
```diff
@@ -0,0 +1,207 @@
...
+            .map_err(StorageError::OpenTable)?;
+
+        // Set the value
+        tx.put(&table, key_bytes, value_bytes, WriteFlags::default())
+            .map_err(StorageError::TxOp)
+            .and_then(|_| tx.commit().map_err(StorageError::Commit).map(|_| ()))
+    }
+
+    /// Get a key from the database
+    pub fn get_table<K: Key, V: Value>(
```
akirillo, line 82: nit: `get_table` is a slightly confusing name, my initial reaction would be that this returns an ORM-style handle on a table rather than look up a single value in a table (same for `set_table` above).

i think the function signature documents itself pretty well with the `table_name` parameter, but if just `get` and `set` feel too vague, maybe we could do `get_value` and `set_value`?

statev2/src/storage/db.rs
```diff
@@ -0,0 +1,207 @@
...
+            .unwrap();
+        let val: Option<TestValue> = db.get_table(TABLE_NAME, &key_name).unwrap();
+
+        assert_eq!(val.unwrap(), TestValue::dummy());
+    }
+
+    /// Tests retrieving a value that doesn't exist
+    #[test]
+    fn test_get_nonexistent() {
+        // Add a value to the DB then read it back
```
akirillo, line 196: nit: remove copied comment

akirillo (approved):
couple nits but lgtm!
perhaps a question for later: does `flexbuffers` do any compression? down the road i could see us wanting to persist data in compressed form, though also it feels like a relayer won't need to maintain toooo much state

statev2/Cargo.toml
joeykraut: yeah planning to look into all that later

joeykraut:
> couple nits but lgtm! perhaps a question for later: does `flexbuffers` do any compression? down the road i could see us wanting to persist data in compressed form, though also it feels like a relayer won't need to maintain toooo much state

No, it's a serialization format first and foremost. We likely would not want compression even if it were available, our dataset is tiny and the memory footprint is already low. Adding compression just means higher read/write latency

PR #196, opened 2023-09-11
statev2: db: Add transaction semantics and full functionality test cases

joeykraut:
### Purpose
This PR adds a set of transactional semantics over the `mdbx` database wrapper. Together these wrappers effectively provide a cleaner (fewer flags) interface to the db, and automatic serialization/deserialization to the caller's expected type. This thin wrapper will be composed into higher layers of abstraction by the replication layer and application logic.

As well, this PR adds a number of tests for various basic cases as well as transaction isolation, crash recovery, etc.

### Testing

### DIFF
--- a/statev2/src/storage/db.rs
+++ b/statev2/src/storage/db.rs
@@ -5,7 +5,9 @@
 
 use std::{borrow::Cow, path::Path};
 
-use libmdbx::{Database, TableFlags, Transaction, WriteFlags, WriteMap, RO, RW};
+use libmdbx::{
+    Database, Table, TableFlags, Transaction, TransactionKind, WriteFlags, WriteMap, RO, RW,
+};
 use serde::{Deserialize, Serialize};
 
 use super::{
@@ -15,6 +17,10 @@ use super::{
 
 /// The number of tables to open in the database
 const NUM_TABLES: usize = 2;
+
+// ------------
+// | Database |
+// ------------
 
 /// The database config
 pub struct DbConfig {
@@ -49,10 +55,9 @@ impl DB {
     pub fn create_table(&self, table_name: &str) -> Result<(), StorageError> {
         // Begin a tx
         let tx = self.new_write_tx()?;
-        tx.create_table(Some(table_name), TableFlags::default())
-            .map_err(StorageError::TxOp)
-            .map(|_| ())
-            .and_then(|_| tx.commit().map_err(StorageError::Commit).map(|_| ()))
+        tx.create_table(table_name)?;
+
+        tx.commit()
     }
 
     /// Set a key in the database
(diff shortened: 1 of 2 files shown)


statev2/src/storage/db.rs
```diff
@@ -195,12 +287,177 @@ mod test {
...
-
         db.create_table(TABLE_NAME).unwrap();
         let val: Option<TestValue> = db.read(TABLE_NAME, &key_name).unwrap();
 
         assert_eq!(val, None);
     }
+
+    /// Tests a read only tx to a table
+    #[test]
+    fn test_ro_tx_simple() {
```
akirillo, line 299: thoughts on having a test case for being unable to write in a RO tx?

statev2/src/storage/db.rs
```diff
@@ -195,12 +287,177 @@ mod test {
...
+        let v2: Option<TestValue> = tx.read(TABLE_NAME2, &key2).unwrap();
+        tx.commit().unwrap();
+
+        assert_eq!(v1.unwrap(), value1);
+        assert_eq!(v2.unwrap(), value2);
+    }
+
+    /// Tests concurrent transaction isolation
+    #[test]
+    fn test_concurrent_updates() {
```
akirillo, line 386: gud test, should we run it multiple times or w/ more threads to have higher confidence that some lock contention occurred?

statev2/src/storage/db.rs
```diff
@@ -195,12 +287,177 @@ mod test {
...
+            ..Default::default()
+        };
+
+        let tx = db.new_write_tx().unwrap();
+        tx.create_table(TABLE_NAME).unwrap();
+        tx.write(TABLE_NAME, &key, &value).unwrap();
+        tx.commit().unwrap();
+
+        // Drop the db to simulate a crash
+        db.sync().unwrap();
```
akirillo, line 451: more out of curiosity, is persistence to disk only ensured by calling `sync`? if there's no periodic syncing going on, then it's on us to make sure we call this often enough / in the right areas, which may cause some annoying heisenbugs

akirillo (approved):
lgtm, couple questions

statev2/src/storage/db.rs
joeykraut: In a `RO` tx the method to write isn't even in scope, so I'm not sure what that would test?

statev2/src/storage/db.rs
joeykraut: I prefer to keep it simple, the test will be run many times anyways over the course of its lifetime.

Plus, practically speaking, `mdbx` has a global write lock, so there isn't *actually* any chance that these two transactions run concurrently

statev2/src/storage/db.rs
joeykraut: No this isn't the only that we write back to disk, it does periodically do so on its own. The `sync` method is just a way to force it

PR #197, opened 2023-09-12
statev2: storage: cursor: Add `Cursor` implementation

joeykraut:
### Purpose
This PR adds an initial `Cursor` implementation on top of the database. This wraps an internal cursor and provides a serialization/deserialization layer on top of the cursor.

We could have used the `libdmbx::orm` module to do this ([link](https://docs.rs/libmdbx/latest/libmdbx/orm/index.html)), but this is a bit restrictive in that *all* tables must have a model at database creation time. Some tables will store heterogenous data (e.g. metadata tables), so the ORM doesn't quite give us what we want.

### Testing
- Planning to add more tests in a follow up PR


akirillo (approved):
clean, lgtm

PR #198, opened 2023-09-12
storage: cursor: Add cursor unit tests

joeykraut:
### Purpose
This PR adds a full unit test suite for the `Cursor` interface. Of particular interest is sorted iteration: we currently sort numeric keys by casting them to a string. This is okay for now, but should we want more advanced sorting we can implement a hook into the [`MDBX_DUPSORT`](https://erthink.github.io/libmdbx/group__c__dbi.html#ggafe3bddb297b3ab0d828a487c5726f76aaed33d978b4e23b7c1659e8106e9c7acd:~:text=comparison%20for%20keys.-,MDBX_DUPSORT,-Use%20sorted%20duplicates) interface.

### DIFF
--- a/statev2/Cargo.toml
+++ b/statev2/Cargo.toml
@@ -12,3 +12,4 @@ serde = { workspace = true, features = ["derive"] }
 
 [dev-dependencies]
 tempfile = "3.8"
+rand = { workspace = true }
--- a/statev2/src/storage/cursor.rs
+++ b/statev2/src/storage/cursor.rs
@@ -138,16 +138,38 @@ impl<'txn, T: TransactionKind, K: Key, V: Value> Iterator for DbCursor<'txn, T, 
 
 #[cfg(test)]
 mod test {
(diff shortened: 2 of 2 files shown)


akirillo (approved):
nice, lgtm - when I was reviewing the previous PRs, I didn't really notice any explicit sorting, we just call out to `libmdbx`'s implementations for `read`/`write`, and the cursor methods, etc. So is the lexicographic sorting implicit from MDBX & the fact that we serialize keys/values to bytes?

joeykraut:
> nice, lgtm - when I was reviewing the previous PRs, I didn't really notice any explicit sorting, we just call out to `libmdbx`'s implementations for `read`/`write`, and the cursor methods, etc. So is the lexicographic sorting implicit from MDBX & the fact that we serialize keys/values to bytes?

Discussed in person, copying here: that is correct, `mdbx` orders the keys for us

PR #199, opened 2023-09-13
statev2: replication: log-store: Add `LogStore` implementation

joeykraut:
### Purpose
The `raft` needs a [`Storage`](https://docs.rs/raft/latest/raft/storage/trait.Storage.html) impl to read its log, metadata, snapshots, etc from. Concretely we use the embedded `mdbx` instance in the storage layer as the underlying persistence mechanism. 

This means that the DB serves two purposes:
- Store the state machine of the relayer and its cluster
- Store the raft log and metadata used to form consensus over the relayer's state machine

This PR implements the `Storage` trait for `LogStore` which functions as the plumbing between the storage and replication layers.

### Testing
- Will add tests in a follow up PR

### DIFF
--- a/statev2/src/replication/log_store.rs
+++ b/statev2/src/replication/log_store.rs
@@ -1,0 +1,233 @@
+//! Defines the storage layer for the `raft` implementation. We store logs, snapshots,
+//! metadata, etc in the storage layer -- concretely an embedded KV store
+
+use std::sync::Arc;
+
+use libmdbx::{TransactionKind, RO};
+use protobuf::Message;
+use raft::{
+    prelude::{
+        ConfState, Entry as RaftEntry, HardState, Snapshot as RaftSnapshot, SnapshotMetadata,
+    },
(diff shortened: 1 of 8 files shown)


statev2/src/replication/log_store.rs
```diff
@@ -0,0 +1,233 @@
...
+pub struct LogStore {
+    /// The underlying database reference
+    db: Arc<DB>,
+}
+
+impl LogStore {
+    /// Constructor
+    pub fn new(db: Arc<DB>) -> Result<Self, ReplicationError> {
+        // Create the logs table in the db
+        db.create_table(RAFT_METADATA_TABLE)
```
akirillo, line 69: Should we have a way to get a handle on the `LogStore` in the runtime without creating the logs table? E.g. if recovering from a crash, the table should already exist, right?

akirillo (approved):
lgtm, one small question

statev2/src/replication/log_store.rs
joeykraut: That's what `create_table` does, if the table already exists it's a no-op

PR #200, opened 2023-09-13
statev2: replication: log-store: Add `LogStore` unit tests

joeykraut:
### Purpose
This PR adds unit tests to the `LogStore` implementation testing a variety of functionality, including:
- Metadata access at initialized and non-initialized states
- Full logs access
- Partial log access
- Log access with a size bound in place

As well, this PR chances the semantics of the `seek` interfaces on the `DbCursor` type. Previously, the `seek_<x>` methods would return the kv pair under the cursor after seeking, thereby moving the cursor to the next element. It is more convenient to higher level modules if the `seek` interfaces simply reposition the cursor, and the next access to `get_current` or `Iterator::next` on the `DbCursor` will return the value. This PR achieves this by buffering a result after a `seek`.

### Testing

### DIFF
--- a/statev2/src/replication/log_store.rs
+++ b/statev2/src/replication/log_store.rs
@@ -68,6 +68,8 @@ impl LogStore {
         // Create the logs table in the db
         db.create_table(RAFT_METADATA_TABLE)
             .map_err(ReplicationError::Storage)?;
+        db.create_table(RAFT_LOGS_TABLE)
+            .map_err(ReplicationError::Storage)?;
 
         Ok(Self { db })
     }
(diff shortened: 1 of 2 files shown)


statev2/src/replication/log_store.rs
```diff
@@ -216,18 +220,315 @@ impl Storage for LogStore {
...
+    #[test]
+    fn test_empty_log() {
+        let store = mock_log_store();
+
+        let first = store.first_index().unwrap();
+        let last = store.last_index().unwrap();
+        let entry_term = store.term(1 /* index */);
+
+        assert_eq!(first, 0);
+        assert_eq!(last, 0);
```
akirillo, line 434: curious as to why the desired behavior is return `0` when a log at the given index doesn't exist, why don't we return a `RaftStorageError::Unavailable`?

akirillo (approved):
lgtm, just a smol q

statev2/src/replication/log_store.rs
joeykraut: The consumer of this interface is the raft consensus engine, if we return error that signals a fault to the engine whereas if we return `0`, the engine will know the log to be empty

PR #201, opened 2023-09-13
statev2: replication: raft-node: Add initial raft node implementation

joeykraut:
### Purpose
This PR takes the first steps towards including the `raft` consensus engine in our `statev2` replication layer. This involves a few changes to the `LogStore` implementation to bring it into compatibility with the `RaftNode`'s use. Specifically we:
- Apply a default snapshot when the `LogStore` is initialized
- Fall back to this default when other values are not available in storage

### Testing

### DIFF
--- a/statev2/src/replication/log_store.rs
+++ b/statev2/src/replication/log_store.rs
@@ -1,7 +1,7 @@
 //! Defines the storage layer for the `raft` implementation. We store logs, snapshots,
 //! metadata, etc in the storage layer -- concretely an embedded KV store
 
-use std::sync::Arc;
+use std::{cmp, sync::Arc};
 
 use libmdbx::{TransactionKind, RO};
 use protobuf::Message;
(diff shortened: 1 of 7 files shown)


statev2/src/replication/log_store.rs
```diff
@@ -37,6 +37,9 @@ pub const CONF_STATE_KEY: &str = "conf-state";
 /// The name of the snapshot metadata key in the KV store
 pub const SNAPSHOT_METADATA_KEY: &str = "snapshot-metadata";
 
+/// A marker for a placeholder in an unused variable
+pub const UNUSED: u64 = 0;
```
akirillo, line 41: nit: it looks like this constant is only used as a placeholder for the `to` field in the `snapshot` method. could we update either the variable name or docstring to specify this? the current description is a bit confusing and might lead to this constant being used in the wrong place

akirillo (approved):
lgtm, tiny nit

statev2/src/replication/log_store.rs
joeykraut: sure good call, I'll update the docstring because I think `UNUSED` reads better inline when this variable is used

PR #202, opened 2023-09-14
statev2: replication: raft-node: Add raft loop handler implmentation

joeykraut:
### Purpose
This PR implements the main driver loop for the raft consensus engine. This event loop effectively serves two purposes:
- Poll for inbound messages from the network
- Process the ready state from the consensus engine, this involves:
    - Handling outbound messages to peers, generated by the consensus engine
    - Committing log entries and applying them to the state machine
    - Appending new (uncommitted) log entries to the persisted log
    - Applying config and hard state changes

This process connects the raft consensus engine to the storage and networking layers.

### Testing
- Testing of the event loop will be done via the functional unit imposed by defining *our* state transitions, so for now it is expected that this implementation is buggy

### DIFF
--- a/statev2/src/replication/log_store.rs
+++ b/statev2/src/replication/log_store.rs
@@ -79,7 +79,7 @@ impl LogStore {
 
         // Write a default snapshot to the metadata table
         let self_ = Self { db };
-        self_.apply_snapshot(RaftSnapshot::new());
+        self_.apply_snapshot(&RaftSnapshot::new());
 
         Ok(self_)
     }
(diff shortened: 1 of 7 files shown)


statev2/src/replication/log_store.rs
```diff
@@ -112,8 +112,44 @@ impl LogStore {
...
+    }
+
+    /// Append entries to the raft log
+    pub fn append_log_entries(&self, entries: Vec<RaftEntry>) -> Result<(), ReplicationError> {
+        let tx = self.db.new_write_tx().map_err(ReplicationError::Storage)?;
+        for entry in entries.into_iter() {
+            let key = lsn_to_key(entry.index);
+            let value = ProtoStorageWrapper(entry);
+
+            tx.write(RAFT_LOGS_TABLE, &key, &value)
```
akirillo, line 144: thoughts on exposing a batch-write method on the `DbTxn`?

statev2/src/replication/network.rs
```diff
@@ -0,0 +1,88 @@
+//! The network abstraction of the replication layer; defines the interface for
+//! sending and receiving raft messages from cluster peers
+
+use raft::prelude::Message as RaftMessage;
+
+use super::error::ReplicationError;
+
+/// The central trait that must be implemented by any networking layer used
+/// by the replication layer
+pub trait Network {
```
akirillo, line 10: nit: `Network` feels a bit vague, thoughts on renaming this to e.g. `ReplicationNetwork`?

statev2/src/replication/network.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    /// Send a raft message to a peer
+    ///
+    /// This method MAY NOT not block
+    fn send(&mut self, message: RaftMessage) -> Result<(), Self::Error>;
+
+    /// Receive a raft message from the network
+    ///
+    /// This method MAY block
+    fn recv(&mut self) -> Result<Option<RaftMessage>, Self::Error>;
```
akirillo, line 22: thoughts on making this `async`?

akirillo (approved):
lgtm, couple small suggestions

statev2/src/replication/log_store.rs
joeykraut: Prefer the lean interface, we don't need a batch method

statev2/src/replication/network.rs
joeykraut: The consensus engine isn't async so we get nothing out of laying an asynchronous runtime on top

PR #203, opened 2023-09-14
circuit-types, circuits: Update `Wallet` and `Balance` structs for fee impl

joeykraut:
### Purpose
This PR changes the structure of the `Wallet` and `Balance` types to begin implementation of the fee design. These structural changes simplify fees to a single fixed point take that a relayer may capture. As well, we introduce the notion of a `managing_cluster` to the `Wallet`.

This PR also removes many of the fee-centric constraints in the circuit so that the crates compile. These circuits will be updated in follow up PRs

### Testing
- Unit and integration tests pass for `circuits` and `circuit-types`
- CI will remain red while the code is in an intermediate state

### DIFF
--- a/circuit-types/src/balance.rs
+++ b/circuit-types/src/balance.rs
@@ -48,6 +48,10 @@ pub struct Balance {
     pub mint: BigUint,
     /// The amount of the given token stored in this balance
     pub amount: u64,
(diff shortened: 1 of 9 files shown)


akirillo (approved):
lgtm

PR #206, opened 2023-09-15
circuits: mpc-circuits: match: Compute protocol fee takes in MPC circuit

joeykraut:
### Purpose
This PR computes the protocol fee takes in the `match` MPC as a publically known fixed point constant multiplied by the volumes of each asset matched.

Relayer fees and the integration of both fee types into the `VALID MATCH MPC` and `VALID SETTLE` circuits will come in a follow up.

### Testing
- Added unit tests for the fee take

### DIFF
--- a/circuits/integration/zk_circuits/valid_match_mpc.rs
+++ b/circuits/integration/zk_circuits/valid_match_mpc.rs
@@ -127,6 +127,8 @@ async fn match_orders(
         base_amount: min_base_amount,
         quote_amount,
         direction: party0_order.side.into(),
+        protocol_base_fee_amount: 0,
+        protocol_quote_fee_amount: 0,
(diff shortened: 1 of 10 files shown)


circuits/integration/zk_circuits/valid_match_mpc.rs
```diff
@@ -127,6 +127,8 @@ async fn match_orders(
         base_amount: min_base_amount,
         quote_amount,
         direction: party0_order.side.into(),
+        protocol_base_fee_amount: 0,
+        protocol_quote_fee_amount: 0,
```
joeykraut, line 131: Dummy values, planning to update when the constraints are written (tests will fail if not updated then)

akirillo (approved):

PR #207, opened 2023-09-15
circuits: mpc-circuits: match: Compute relayer fee takes in MPC circuit

joeykraut:
### Purpose
This PR adds the computation of the relayer fee take to the match mpc. This is concretely done as follows: the relayers each input a fee term to the MPC which is multiplied by the base amount in the match. The implicit assumption is that the buy side relayer (sells the quote token to buy the base) has pre-multiplied the fee take by the execution price to change units into the quote token. This assumption is constrained in the match proof that follows.

Integration of this fee computation and the protocol fee computation into the proofs of `VALID MATCH MPC` and `VALID SETTLE` will be done in a follow up.

### Testing
- CI will fail until the fee implementation is further along

### DIFF
--- a/circuits/src/mpc_circuits/match.rs
+++ b/circuits/src/mpc_circuits/match.rs
@@ -23,12 +23,20 @@ use crate::mpc_gadgets::{comparators::min, fixed_point::FixedPointMpcGadget};
 /// So, if the match result is invalid, the orders don't overlap, etc; the result will
 /// never be opened, and the information never leaked. Therefore, we do not need to zero
 /// out any values in the circuit.
+///
+/// The fee terms are calculated outside of the circuit and passed in directly by the relayers.
+/// For a relayer that intends to sell the base asset, this term is directly the fee authorized by
+/// the user. For a relayer that intends to sell the quote asset, a unit change is implicitly applied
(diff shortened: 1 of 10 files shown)


circuits/src/mpc_circuits/match.rs
```diff
@@ -23,12 +23,20 @@ use crate::mpc_gadgets::{comparators::min, fixed_point::FixedPointMpcGadget};
 /// So, if the match result is invalid, the orders don't overlap, etc; the result will
 /// never be opened, and the information never leaked. Therefore, we do not need to zero
 /// out any values in the circuit.
+///
+/// The fee terms are calculated outside of the circuit and passed in directly by the relayers.
+/// For a relayer that intends to sell the base asset, this term is directly the fee authorized by
+/// the user. For a relayer that intends to sell the quote asset, a unit change is implicitly applied
```
akirillo, line 29: nit: it may aid readability to phrase this is the unit change being "assumed to be applied" as opposed to "implicitly applied", and mention that this assumption is constrained in the match proof

(this would also be consistent with the great doc comment below)

akirillo (approved):
one nit, lgtm

PR #208, opened 2023-09-19
docker: frontend-test-stack: Add docker compose setup for FE testing

joeykraut:
### Purpose
This PR adds a docker compose setup for the frontend SDK integration tests. This setup is a Katana devnet node and a relayer configured to use the katana node.

Left to do are:
- Deploy a USDC ERC20 contract to the devnet node so that the SDK may test transfers and trades
- Add a script layer over the top of the compose that allows the tester to supply config information

### Testing
- Ran the compose setup, verified that basic relayer <--> katana functionality worked


docker/frontend-test-stack/docker-compose.yml
```diff
@@ -0,0 +1,37 @@
...
+    entrypoint: sleep infinity
+  sequencer:
+    build:
+      context: ../devnet
+    volumes:
+      - deployments:/deployments
+    ports:
+      - "5050:5050"
+    environment:
+      - STARKNET_DEVNET_CAIRO_VM=rust
```
akirillo, line 15: I don't think this env var does anything for Katana, it may be an artifact from when we were using the Python-based `starknet-devnet`

akirillo (approved):
lgtm, tiny nit

docker/frontend-test-stack/docker-compose.yml
joeykraut: oops yes, artifact from when we were using the devnet

PR #209, opened 2023-09-19
docker: devnet: Deploy USDC erc20 with devnet setup

joeykraut:
### Purpose
This PR deploys an ERC20 contract for a USDC token to the Katana devnet. 

### Testing
- Ran the devnet node and ran a few ERC20 calls against the Katana node using `starkli`


akirillo (approved):
try keeping PRs smaller in the future, please

PR #210, opened 2023-09-20
docker: frontend-test-stack: Build relayer in release mode

joeykraut:
### Purpose
This PR builds the relayer in release mode and switches the chain ID to katana in the config


akirillo (approved):
lfartm

PR #211, opened 2023-09-20
statev2: proto: Add state transition protobuf definition

joeykraut:
### Purpose
This PR introduces the basic protobuf setup for state transitions. Specifically we build the protos into the `target` folder using a build script and then re-import them in the library at library compile time.

### Todo
- Define the state transitions necessary for our application

### Testing
- Tested a simple serialization/deserialization test case


statev2/proto/build.rs
akirillo, line 1: getting clippy warnings about missing docs for this crate

akirillo (approved):
lgtm, just some clippy warnings

PR #212, opened 2023-09-21
statev2: proto: Add proto definitions for all existing state transitions

joeykraut:
### Purpose
This PR adds protobuf definitions for all state types that currently have setter methods in the existing state implementation.

As well, this PR derives a builder pattern for all generated protobuf types.

### Testing
- Tested the builder pattern

### DIFF
--- a/statev2/proto/proto/orders.proto
+++ b/statev2/proto/proto/orders.proto
@@ -1,0 +1,67 @@
+/// Defines orderbook metadata message types
+syntax = "proto3";
+package state;
+
+import "common.proto";
+
+// ---------
+// | Types |
+// ---------
+
+/// The state of the order in the network
+enum NetworkOrderState {
+    /// The received state indicates that the local node knows about the order, but
+    /// has not received a proof of `VALID COMMITMENTS` to indicate that this order
+    /// is a valid member of the state tree
+    ///
+    /// Orders in the received state cannot yet be matched against
+    Received = 0;
+    /// The verified state indicates that a proof of `VALID COMMITMENTS` has been received
+    /// and verified by the local node
+    ///
+    /// Orders in the Verified state are ready to be matched
+    Verified = 1;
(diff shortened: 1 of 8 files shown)


statev2/proto/src/lib.rs
```diff
@@ -1,28 +1,94 @@
...
 #[cfg(test)]
 mod tests {
-    use super::NewPeer;
+    use crate::{AddOrderBuilder, AddPeersBuilder, NetworkOrderBuilder, PeerInfoBuilder};
+
+    use super::{AddOrder, AddPeers};
     use prost::Message;
 
     /// Tests the add new peer message
     #[test]
```
akirillo, line 64: is the plan to add serde tests for all proto definitions? feels weird to have it just be for some types, but can also just leave the serde code path to be covered by future unit tests of the state module

akirillo (approved):
lgtm, 2 small qs

statev2/proto/proto/orders.proto
```diff
@@ -0,0 +1,67 @@
...
+    ProtoUuid id = 1;
+    /// The public shares nullifier of the wallet holding this order
+    ProtoScalar nullifier = 2;
+    /// The cluster managing the order
+    ClusterId cluster = 3;
+    /// The state of the order in the network
+    NetworkOrderState state = 4;
+    /// The proof payload of `VALID COMMITMENTS` for this order
+    /// and `VALID REBLIND` for the containing wallet
+    bytes proof = 5;
```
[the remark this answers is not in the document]
joeykraut: not as far as i see :)

statev2/proto/src/lib.rs
joeykraut: No, at that point we're just testing the library, these are more sanity checks that the `prost` configuration ends up with what we want. No need to over-test

PR #213, opened 2023-09-22
statev2: applicator: Create state applicator and add peer index setter methods

joeykraut:
### Purpose
The `StateApplicator` is in charge of appling state transitions to the storage layer. If we think of our state primitive as a distributed log that forms consensus over a sequence of state transition RPCs, the `StateApplicator` is then the RPC server executing these requests.

This PR also adds setters for the `PeerIndex` methods that store peer information and cluster membership information

### Testing
- Tested peer addition and removal

### DIFF
--- a/statev2/state/src/applicator/error.rs
+++ b/statev2/state/src/applicator/error.rs
@@ -1,0 +1,32 @@
+//! Error types emitted by the state applicator
+
+use std::{error::Error, fmt::Display};
+
+use state_proto::error::StateProtoError;
+
+use crate::storage::error::StorageError;
+
+/// The error type emitted by the replication layer
+#[derive(Debug)]
+pub enum StateApplicatorError {
+    /// An error interacting with storage
+    Storage(StorageError),
+    /// An error parsing a message separately from proto errors
+    Parse(String),
+    /// An error processing a proto
+    Proto(StateProtoError),
+}
+
+impl Display for StateApplicatorError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{self:?}")
+    }
+}
+
+impl Error for StateApplicatorError {}
+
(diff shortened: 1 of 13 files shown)


statev2/state/src/applicator/error.rs
```diff
@@ -0,0 +1,32 @@
+//! Error types emitted by the state applicator
+
+use std::{error::Error, fmt::Display};
+
+use state_proto::error::StateProtoError;
+
+use crate::storage::error::StorageError;
+
+/// The error type emitted by the replication layer
```
akirillo, line 9:
```suggestion
/// The error type emitted by the storage applicator
```

statev2/state/src/applicator/peer_index.rs
```diff
@@ -0,0 +1,242 @@
...
+        let tx = self
+            .db()
+            .new_write_tx()
+            .map_err(StateApplicatorError::Storage)?;
+
+        // Index each peer
+        for peer in msg.peers.into_iter() {
+            // Parse the peer info and mark a successful heartbeat
+            let peer_info = PeerInfo::try_from(peer).map_err(StateApplicatorError::Proto)?;
+            peer_info.successful_heartbeat();
```
akirillo, line 32: is it accurate to mark a successful heartbeat here? not sure where a raft node may receive the `AddPeersMsg` from, but I could see situations where it syncs a set of known peers from one such peer, this wouldn't really be the same as receiving a heartbeat from each of these peers

akirillo (approved):
lgtm, smol nit + smol raft q

statev2/state/src/applicator/peer_index.rs
joeykraut: Yeah this effectively resets the timeout so that the latency in the connection layer is not observed as a peer timeout. This gives the local peer time to heartbeat the discovered peer.

PR #214, opened 2023-09-22
statev2: applicator: order-book: Add order book state transition handlers

joeykraut:
### Purpose
This PR adds state transition handler for order book types. This includes:
- Adding new orders to the book
- Attaching validity proofs for an order
- Nullifying orders when their nullifiers are witnessed on-chain

### Testing

### DIFF
--- a/statev2/state/src/applicator/order_book.rs
+++ b/statev2/state/src/applicator/order_book.rs
@@ -3,9 +3,15 @@
 //! TODO: For the order book in particular, it is likely to our advantage to index orders outside
 //! of the DB as well in an in-memory data structure for efficient lookup
 
-use common::types::{gossip::ClusterId, network_order::NetworkOrder, wallet::OrderIdentifier};
+use common::types::{
+    gossip::ClusterId,
+    network_order::{NetworkOrder, NetworkOrderState},
+    proof_bundles::OrderValidityProofBundle,
+    wallet::OrderIdentifier,
+};
 use constants::ORDER_STATE_CHANGE_TOPIC;
 use external_api::bus_message::SystemBusMessage;
+use itertools::Itertools;
 use libmdbx::{TransactionKind, RW};
 use mpc_stark::algebra::scalar::Scalar;
 use serde::{Deserialize, Serialize};
@@ -13,6 +19,7 @@ use state_proto::{
     AddOrder as AddOrderMsg, AddOrderValidityProof as AddOrderValidityProofMsg,
     NullifyOrders as NullifyOrdersMsg,
 };
+use uuid::Error as UuidError;
 
 use crate::{
     applicator::{error::StateApplicatorError, ORDERS_TABLE, PRIORITIES_TABLE},
@@ -96,7 +103,22 @@ impl StateApplicator {
 
     /// Add a validity proof for an order
     pub fn add_order_validity_proof(&self, msg: AddOrderValidityProofMsg) -> Result<()> {
-        unimplemented!()
+        // Deserialize the proof bundle
+        let id: OrderIdentifier = msg
+            .order_id
+            .unwrap_or_default()
+            .try_into()
+            .map_err(|e: UuidError| StateApplicatorError::Parse(e.to_string()))?;
+
+        let bundle: OrderValidityProofBundle = serde_json::from_slice(&msg.proof)
+            .map_err(|e| StateApplicatorError::Parse(e.to_string()))?;
+
+        let tx = self
+            .db()
+            .new_write_tx()
+            .map_err(StateApplicatorError::Storage)?;
+        Self::attach_validity_proof_with_tx(&id, bundle, &tx)?;
+        tx.commit().map_err(StateApplicatorError::Storage)
     }
 
     /// Nullify orders indexed by a given wallet share nullifier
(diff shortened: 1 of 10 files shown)


statev2/state/src/applicator/order_book.rs
```diff
@@ -0,0 +1,488 @@
...
+
+        // Re-index based on the proof's nullifier
+        let prev_nullifier = order_info.public_share_nullifier;
+        let new_nullifier = proof.reblind_proof.statement.original_shares_nullifier;
+        if prev_nullifier != new_nullifier {
+            Self::update_order_nullifier(order_id, prev_nullifier, new_nullifier, tx)?;
+        }
+
+        // Update the order's info
+        order_info.state = NetworkOrderState::Verified;
```
akirillo, line 174: Does this mean `attach_validity_proof_with_tx` is assumed to be called after the validity proof has been verified? I don't see that happening in `add_order_validity_proof`

statev2/state/src/applicator/order_book.rs
```diff
@@ -0,0 +1,488 @@
...
+    /// Cancel an order
+    fn cancel_order_with_tx(&self, order_id: &OrderIdentifier, tx: &DbTxn<'_, RW>) -> Result<()> {
+        let mut order = Self::read_order_info(order_id, tx)?;
+        order.state = NetworkOrderState::Cancelled;
+        order.validity_proof_witnesses = None;
+        order.validity_proofs = None;
+
+        Self::write_order_info(&order, tx)?;
+        self.system_bus().publish(
+            ORDER_STATE_CHANGE_TOPIC.to_string(),
```
akirillo, line 189: I think this is fine for now, but moving forward I'm a bit wary of properly tracking where/when we publish to the system bus. What I would want to avoid is flooding the system bus with messages redundant messages when multiple helpers get called / some helper gets called repeatedly, e.g. if we do multiple different updates on an order in a single state transition and publish to the system bus for each update. Would be nice to publish the minimum set of necessary messages to the system bus in one fell swoop at the end of a state transition / public function on this interface, so we can be sure that no redundant publishing is being done elsewhere in the helpers.

(to be clear, I understand that there are currently no redundant messages being publish to the system bus, I just want to be careful about this moving forward)

statev2/state/src/applicator/order_book.rs
```diff
@@ -0,0 +1,488 @@
...
+            .map(|priority| priority.unwrap_or(CLUSTER_DEFAULT_PRIORITY))
+    }
+
+    /// Write an order priority to the DB
+    fn write_order_priority_with_tx(order: &NetworkOrder, tx: &DbTxn<'_, RW>) -> Result<()> {
+        // Lookup the cluster priority and write the order's priority
+        let cluster_priority = Self::get_cluster_priority_with_tx(&order.cluster, tx)?;
+        let priority = OrderPriority {
+            cluster_priority,
+            order_priority: ORDER_DEFAULT_PRIORITY,
```
akirillo, line 240: It feels a bit strange to me that this function overwrites the `OrderPriority.order_priority` field. The function signature / docstring doesn't really indicate this "reset to default" behavior, nor does it indicate that the `NetworkOrder` is expected to represent an order with an uninitialized priority. Perhaps it would make sense to add an `order_priority` field to the `NetworkOrder` type?

statev2/state/src/applicator/order_book.rs
```diff
@@ -0,0 +1,488 @@
...
+        // Read the nullifier set for the old nullifier and remove the order
+        let old_set = Self::read_nullifier_set(old_nullifier, tx)?;
+        let updated_old_set = old_set
+            .into_iter()
+            .filter(|id| id != order_id)
+            .collect_vec();
+        Self::write_nullifier_set(old_nullifier, updated_old_set, tx)?;
+
+        // Add the order to the new nullifier set
+        Self::append_to_nullifier_set(new_nullifier, *order_id, tx)
```
akirillo, line 277: This feels unnecessarily DRY to me. It makes sense to use `append_to_nullifier_set` in `add_order_with_tx`, but using it here makes `update_order_nullifier` incur 2 reads and 2 writes where it only needs to incur one of each, and I don't think this helps readability all that much more than the alternative.

akirillo (approved):
nothing blocking, but do have some qs / suggestions

statev2/state/src/applicator/order_book.rs
joeykraut: It still needs 2 reads and 2 writes; and this is way nicer to work with than the raw deserialization and error handling that the method encapsulates

statev2/state/src/applicator/order_book.rs
joeykraut: This method is only called when a new order is added, so no priority events have occurred that would change the order's priority.

Plus I'd rather not form consensus over priorities, I'd rather optimistically gossip (if anything at all) about protocol faults that would lower a priority. This is far less complex

statev2/state/src/applicator/order_book.rs
joeykraut: Yes that's correct, this assumption isn't mandated by the interface (I see no simple way to do so), but instead follows the pattern that the rest of the relayer lays out: authentication happens at the network and gossip layers which then dispatch to the storage layer

statev2/state/src/applicator/order_book.rs
joeykraut: Sure, any ideas on how to guarantee this?

PR #218, opened 2023-09-28
statev2: applicator, replication: Accept inbound proposals, and apply state transitions

joeykraut:
### Purpose
This PR takes the first step in pulling together the replication, storage, and applicator layers by accepting state transition proposals from outside the state management module. The flow is:
1. An inbound proposal reaches the raft event loop via the `proposals_queue` channel
2. The raft node proposes the state transition to the cluster
3. The raft group appends the state transition to the log, forms consensus over the entry, then commits the log entry
4. The raft nodes individually forward the now-committed state transition to the `StateApplicator` to apply the state transition

### Testing
- Tested state application of a new wallet to test the communication flow

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,10 +57,13 @@ libp2p-identity = { version = "0.1" }
 libp2p-swarm = { version = "0.42" }
 libp2p-swarm-derive = { version = "0.32" }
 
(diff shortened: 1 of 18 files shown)


akirillo (approved):
lgtm

PR #219, opened 2023-10-06
handshake-manager: Add Katana token mappings

sehyunc:
This PR adds a token mapping for the Katana devnet to allow for internal matches.


common/src/types/token.rs
```diff
@@ -626,6 +626,28 @@ lazy_static! {
             ),
         ].into_iter().collect()
     };
+
+    /// Remaps Katana ERC20 addresses to their Ethereum mainnet counterparts
+    /// so that a price can be determined
+    ///
+    /// deployments.json generated by renegade-contracts
```
joeykraut, line 633: nit: What does this comment mean? Is this added by mistake?

common/src/types/token.rs
```diff
@@ -626,6 +626,28 @@ lazy_static! {
...
+    ///
+    /// TODO: This will be removed when we implement our custom bridging solution
+    #[derive(Debug)]
```
joeykraut, line 636: shouldn't need this derive here, hashmap [already implements `Debug`](https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html#impl-Debug-for-HashMap%3CK,+V,+S%3E)

common/src/types/token.rs
```diff
@@ -668,6 +697,11 @@ impl Token {
         Self::from_starknet_goerli_addr(&format!("0x{}", addr.to_str_radix(16 /* radix */)))
     }
 
+    /// Given an ERC-20 contract address on Katana (represented as a `BigUint`) returns a Token
+    pub fn from_katana_addr_biguint(addr: &BigUint) -> Self {
+        Self::from_katana_addr(&format!("0x{}", addr.to_str_radix(16 /* radix */)))
```
joeykraut, line 702: can we break the bigint -> string out into a private helper now that a few methods use it, specifically:
```rust
fn biguint_to_str_addr(x: &BigUint) -> String {
    let addr = x.to_str_radix(16 /* radix */);
    format!("0x{addr}")
}
```

joeykraut (approved):
Lgtm, a few small requests, feel free to merge when they are addressed

common/src/types/token.rs
sehyunc: was trying to mimic citing the source of token addresses like starknet token map:
```      
/// Remaps Starknet Goerli ERC20 addresses to their Ethereum mainnet counterparts
/// so that a price can be determined
///
/// https://github.com/starknet-io/starknet-addresses/blob/master/bridged_tokens/goerli.json
///
/// TODO: This will be removed when we implement our custom bridging solution
```
will go ahead and remove it since it's just hard coded

PR #221, opened 2023-10-12
handshake-manager: internal-engine: Reutrn no match with zero volume

joeykraut:
### Purpose
Previously, the handshake manager would allow matches of zero volume as part of the logic that allows for undercapitalized orders. This leads to infinite recursive loops in internal matches, and makes no sense, so we explicitly disallow it.

### Testing
- Tested the zero volume case before the fix, verified that the test passed afterwards


sehyunc:
lgtmysbwagpr🚀

akirillo (approved):
what he said

PR #222, opened 2023-10-16
price-reporter: Update Coinbase API WebSocket auth

sehyunc:
This PR updates the payload used for authentication with the Coinbase API to use timestamps with 10 digits instead of 7.


joeykraut (approved):
nice find

PR #223, opened 2023-10-16
statev2: replication: raft-node: Implement cluster join logic

joeykraut:
### Purpose
This PR adds logic to add and remove peers and learners from a raft cluster. As well, this PR fixes a few `LogStore` bugs that popped up during implementation.

Testing coverage will be increased in a future PR.

### Testing
- Tested the case in which a peer joins the cluster and receives replication messages

### DIFF
--- a/statev2/state/src/replication/log_store.rs
+++ b/statev2/state/src/replication/log_store.rs
@@ -1,7 +1,10 @@
 //! Defines the storage layer for the `raft` implementation. We store logs, snapshots,
 //! metadata, etc in the storage layer -- concretely an embedded KV store
 
-use std::{cmp, sync::Arc};
+use std::{
+    cmp::{self, Ordering},
+    sync::Arc,
+};
 
 use libmdbx::{TransactionKind, RO};
 use protobuf::Message;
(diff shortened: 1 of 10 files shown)


statev2/state/src/replication/log_store.rs
```diff
@@ -195,18 +224,12 @@ impl Storage for LogStore {
...
         let tx = self.db.new_read_tx().map_err(RaftError::from)?;
-        let hard_state: ProtoStorageWrapper<HardState> = tx
-            .read(RAFT_METADATA_TABLE, &HARD_STATE_KEY.to_string())
-            .map_err(RaftError::from)?
-            .unwrap_or_default();
-        let conf_state: ProtoStorageWrapper<ConfState> = tx
-            .read(RAFT_METADATA_TABLE, &CONF_STATE_KEY.to_string())
-            .map_err(RaftError::from)?
-            .unwrap_or_default();
+        let hard_state = self.read_hard_state_with_tx(&tx)?;
```
akirillo, line 227: looks like this no longer uses a default `conf_state` or `hard_state` if it's not already written to the DB, should we ensure that a default is written before this method gets called?

akirillo (approved):
lgtm, one small comment that i may be misunderstanding

akirillo:
I am seeing existing tests fail:

```
failures:
    replication::log_store::test::test_empty_log
    replication::log_store::test::test_out_of_date_snapshot
```

seems like `Err` variants may have changed somewhere

PR #224, opened 2023-10-17
statev2: state: replication: Add tests for large cluster join/leave semantics

joeykraut:
### Purpose
This PR adds additional tests for raft cluster join/leave semantics. Specifically for conducting leader election and forming consensus over larger peer-sets, and in the event of node failure.

### Testing


akirillo (approved):
lgtm

PR #225, opened 2023-10-19
common: wallet: Add wallet-level lock to prevent concurrent updates

joeykraut:
### Purpose
Concurrent updates result in confusing state from the client's perspective when one of the updates fails. For now we prevent concurrent updates alltogether. We will introduce a more sophisticated update sequencing and queuing primitive when our state is over a raft.

### Tesing

### DIFF
--- a/common/src/types/wallet.rs
+++ b/common/src/types/wallet.rs
@@ -1,6 +1,14 @@
 //! Defines wallet types useful throughout the workspace
 
-use std::{collections::HashSet, hash::Hash, iter};
+use std::{
+    collections::HashSet,
+    hash::Hash,
+    iter,
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        Arc,
+    },
+};
 
 use circuit_types::{
     balance::Balance,
@@ -16,12 +24,14 @@ use circuit_types::{
     SizedWallet as SizedCircuitWallet, SizedWalletShare,
 };
 use constants::{MAX_BALANCES, MAX_FEES, MAX_ORDERS};
+use derivative::Derivative;
 use indexmap::IndexMap;
 use itertools::Itertools;
 use mpc_stark::algebra::scalar::Scalar;
 use num_bigint::BigUint;
 use renegade_crypto::hash::evaluate_hash_chain;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use tracing::log;
 use uuid::Uuid;
 
 use super::{gossip::WrappedPeerId, merkle::MerkleAuthenticationPath};
@@ -32,7 +42,8 @@ pub type WalletIdentifier = Uuid;
 pub type OrderIdentifier = Uuid;
 
 /// Represents the private keys a relayer has access to for a given wallet
-#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Clone, Debug, Derivative, Serialize, Deserialize)]
+#[derivative(PartialEq, Eq)]
 pub struct PrivateKeyChain {
     /// Optionally the relayer holds sk_root, in which case the relayer has
     /// heightened permissions than the standard case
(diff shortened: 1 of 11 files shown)


common/src/types/wallet.rs
```diff
@@ -149,7 +178,42 @@ impl From<Wallet> for SizedCircuitWallet {
...
+    /// Try to lock the wallet for an update
+    ///
+    /// Returns `true` if the update succeeded
+    pub fn try_lock_wallet(&self) -> bool {
+        log::debug!("locking wallet: {}", self.wallet_id,);
+
+        self.update_locked
+            .compare_exchange(
+                false, /* current */
+                true,  /* current */
```
akirillo, line 192:
```suggestion
                true,  /* new */
```

common/src/types/wallet.rs
```diff
@@ -149,7 +178,42 @@ impl From<Wallet> for SizedCircuitWallet {
...
+            .is_ok()
+    }
+
+    /// Unlock the wallet
+    pub fn unlock_wallet(&self) -> bool {
+        log::debug!("unlocking wallet: {}", self.wallet_id);
+        self.update_locked
+            .compare_exchange(
+                true,  /* current */
+                false, /* current */
```
akirillo, line 205:
```suggestion
                false, /* new */
```

task-driver/src/settle_match_internal.rs
```diff
@@ -244,14 +258,46 @@ impl SettleMatchInternalTask {
...
             proof_manager_work_queue,
             task_state: SettleMatchInternalTaskState::Pending,
+        };
+
+        // Try to lock both wallets, if they cannot be locked then the task cannot be run
+        // and the internal matching engine will re-run next time the proofs are updated
+        let wallet1 = self_.find_wallet_for_order(&order1).await?;
+        let wallet2 = self_.find_wallet_for_order(&order2).await?;
+
+        if !wallet1.try_lock_wallet() {
```
akirillo, line 268: not sure how to think about this, but is this the only place where we'll have to lock a wallet? i'm sure there's many other places in the code where we find a wallet

workers/api-server/src/http/wallet.rs
```diff
@@ -57,6 +58,34 @@ pub(super) fn get_current_timestamp() -> u64 {
...
+        .read_wallet_index()
+        .await
+        .get_wallet(&wallet_id)
+        .await
+        .ok_or_else(|| {
+            ApiServerError::HttpStatusCode(StatusCode::NOT_FOUND, ERR_WALLET_NOT_FOUND.to_string())
+        })?;
+
+    // Acquire the lock for the wallet
+    if wallet.is_locked() {
```
akirillo, line 79: looks like we just check if the wallet is locked here, do we also want to `try_lock_wallet`?

akirillo (approved):
lgtm, this approach def requires a lot of care so i have some qs around if we're sure we're always locking/unlocking in all the right places

workers/api-server/src/http/wallet.rs
joeykraut: no, we want to centralize the actual locking mechanism

task-driver/src/settle_match_internal.rs
joeykraut: well the only places you'd do so are where you update a wallet, which is in the tasks.

task-driver/src/settle_match_internal.rs
```diff
@@ -194,6 +197,17 @@ impl Task for SettleMatchInternalTask {
         Ok(())
     }
 
+    async fn cleanup(&mut self) -> Result<(), Self::Error> {
+        self.find_wallet_for_order(&self.order_id1)
+            .await?
+            .unlock_wallet();
+        self.find_wallet_for_order(&self.order_id1)
```
sehyunc, line 204: should be order2?

sehyunc (changes requested):

sehyunc (approved):
darn good pr

PR #227, opened 2023-10-27
workers: Remove key mutation when updating orders

sehyunc:
This PR fixes a bug where internal matches would not be found for orders that had been updated using the update orders route, due to an issue with mutating keys of IndexMap.


sehyunc:
@joeykraut Let me know if I should change `UpdateOrderResponse` and the comment block on line 591 to reflect the fact that we no longer generate a new order id.

joeykraut (approved):

PR #230, opened 2023-11-01
task-driver: Cleanup if SettleMatchInternal task errors outside of task-driver

sehyunc:
This PR addresses the bug in which if the SettleMatchInternal task errors during instantiation, the cleanup function is not called, resulting in deadlock.

### DIFF
--- a/task-driver/src/settle_match_internal.rs
+++ b/task-driver/src/settle_match_internal.rs
@@ -243,7 +243,7 @@ impl SettleMatchInternalTask {
         global_state: RelayerState,
         proof_manager_work_queue: CrossbeamSender<ProofManagerJob>,
     ) -> Result<Self, SettleMatchInternalTaskError> {
-        let self_ = Self {
+        let mut self_ = Self {
             execution_price,
             order_id1: order1,
             order_id2: order2,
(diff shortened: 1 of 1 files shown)


task-driver/src/settle_match_internal.rs
```diff
@@ -273,6 +273,8 @@ impl SettleMatchInternalTask {
         }
 
         if !wallet2.try_lock_wallet() {
+            let mut self_with_cleanup = self_;
+            self_with_cleanup.cleanup().await?;
             return Err(SettleMatchInternalTaskError::WalletLocked(
```
joeykraut, line 278: I'd do two things to make this error handling cleaner and more extensible:
1. We can declare `self_` as mutable when it is first created. I.e. `let mut self = Self { ... }` on line 246
2. Let's just add a method that does all fallible setup for the task and error handling at a higher granularity, maybe something like:

```rust

fn new(...) -> Result<Self, SettleMatchInternalTaskError> {
    let mut self_ = Self { ... };
    if let Err(e) = self.setup_task(...) {
        self.cleanup().await?;
        return Err(e);
    }

    Ok(self_)
}


fn setup_task(&mut self, ...) -> Result<(), SettleMatchInternalTaskError> {
    // Wallet setup and locking
}

```

That way we get more generic cleanup logic as we add more (possibly fallible) setup logic

joeykraut (approved):

PR #231, opened 2023-11-02
workers: Enable CORS on error responses

sehyunc:
This PR enables CORS on error responses, fixing the issue of API responses not containing error data in the response when received by the client.

### DIFF
--- a/workers/api-server/src/router.rs
+++ b/workers/api-server/src/router.rs
@@ -130,19 +130,26 @@ impl<
 
         // Forward to the typed handler
         let res = self.handle_typed(headers, req_body, url_params).await;
-        if let Ok(resp) = res {
-            // Serialize the response into a body. We explicitly allow for all cross-origin
-            // requests, as users connecting to a locally-run node have a different origin
-            // port.
+        let builder = Response::builder().header("Access-Control-Allow-Origin", "*");
+        match res {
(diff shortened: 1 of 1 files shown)


workers/api-server/src/router.rs
```diff
@@ -130,19 +130,29 @@ impl<
...
         let res = self.handle_typed(headers, req_body, url_params).await;
-        if let Ok(resp) = res {
-            // Serialize the response into a body. We explicitly allow for all cross-origin
-            // requests, as users connecting to a locally-run node have a different origin
-            // port.
-
-            // TODO: Either remove this in the future, or ensure that no sensitive
-            // information can leak from cross-origin requests.
-            Response::builder()
+        match res {
```
joeykraut, line 133: I'd propose simplify this code by setting up all common fields on the response outside of this match, and then inside the match we apply a body and status if necessary:
```rust
let mut builder = Response::builder().header("Access-Control-Allow-Origin", "*");
match res {
    Ok(resp) => builder.body(...),
    Err(ApiServerError::HttpStatusCode(status, msg)) => builder.status(status).body(Body::from(msg)),
    Err(_) => builder.status(...),
}.unwrap()
```

joeykraut (approved):
One simplification suggestion, o/w lgtm

PR #232, opened 2023-11-07
renegade-crypto: hash: poseidon2: Define Poseidon2 interface

joeykraut:
### Purpose
This PR implements the Poseidon 2 hash ([paper](https://eprint.iacr.org/2023/323.pdf)) and tests it against the reference implementation written by the authors.

Our implementation tests the permutation against that defined by the reference implementation, and attaches an Arkworks style sponge construction on top of the permutation to give an interface analogous to the existing poseidon implementation. The round constants are taken from the reference implementation [link](https://github.com/HorizenLabs/poseidon2/blob/main/plain_implementations/src/poseidon2/poseidon2_instance_bn256.rs#L32).

Our implementation differs from the reference in that it is designed to be efficient for binary sizes and smart contract use (i.e. in our Stylus contracts). This is done in the following way:
- The implementation is non-generic, saving a significant amount of code over the reference
- Our implementation defines round constants as constants evaluated by the compiler. The reference implementation defines them via `lazy_static`s that parse the constants from hex strings. The issue with this approach when deployed in a smart contract is the `lazy_static`s will be re-evaluated in every transaction -- as a new runtime is created.
- Our implementation allows a consumer to optionally disable inlining; reducing code bloat in a size-constrained smart contract.

### Testing
- All tests pass
- Tested against reference implementation as discussed above


renegade-crypto/src/hash/poseidon2.rs
```diff
@@ -0,0 +1,297 @@
...
+    #[test]
+    fn test_hash() {
+        let mut rng = thread_rng();
+        let values: [ScalarField; RATE] = rng.gen();
+
+        let expected_hasher = Poseidon2::new(&POSEIDON2_BN256_PARAMS);
+        let mut hasher = Poseidon2Sponge::new();
+
+        // Run the expected implementation
+        let perm_input = [
```
akirillo, line 279: nit: I'm assuming the Horizen implementation expects the first input element to be zero, or something to that effect? would be nice to call this out explicitly in the comment above

akirillo (approved):

renegade-crypto/src/hash/poseidon2.rs
joeykraut: Yeah good call. It's actually to mimic the capacity element of a sponge, their impl only defines a permutation so we have to get a little hacky

PR #235, opened 2023-11-09
circuit-types: traits: Refactor traits over plonk proof system

joeykraut:
### Purpose
This PR refactors the `traits` module in the `circuit-types` crates to reflect the new plonkish arithmetization and multiprover paradigm. Of notable difference:
- The plonk arithmetization requires one gate per addition, so we do not have the `Variable` and `LinearCombination` distinction that Dalek gave. This simplifies the associated types on Variable types. As well, it allow us to reuse variable types between single and multi prover
- We no longer have a need for proof linkable types, the proof linking scheme will come later
- Commitment types are removed, Plonk commits to its witness using KZG, which is unstructured wrt runtime types

### Testing
- CI passes for the crate

### DIFF
--- a/circuit-types/src/traits.rs
+++ b/circuit-types/src/traits.rs
@@ -13,53 +13,45 @@
 //!       values
 //!     - Single-prover variable types: base types allocated in a single-prover
 //!       constraint system
-//!     - Single-prover commitment types: commitments to base types in a
-//!       single-prover system
 //!     - MPC types: base types that have been allocated in an MPC fabric
-//!     - Multi-prover variable types: base types allocated in a multi-prover
-//!       constraint system
-//!     - Multi-prover commitment types: commitments to base types in a
-//!       multi-prover system
-//!     - Linkable types: Types that may be commitment linked across proofs
 //!     - Secret share types: Additive sharings of a base type
 
+use ark_mpc::{algebra::AuthenticatedScalarResult, network::PartyId};
 use async_trait::async_trait;
+use constants::{AuthenticatedScalar, Scalar, ScalarField, SystemCurve, SystemCurveGroup};
 use futures::future::join_all;
 use itertools::Itertools;
-use mpc_bulletproof::{
-    r1cs::{
-        LinearCombination, Prover, R1CSProof, RandomizableConstraintSystem, Variable, Verifier,
+use mpc_plonk::{
+    errors::PlonkError,
+    multiprover::proof_system::{
+        CollaborativeProof, MpcCircuit, MpcPlonkCircuit as GenericMpcPlonkCircuit,
+        MultiproverPlonkKzgSnark,
     },
-    r1cs_mpc::{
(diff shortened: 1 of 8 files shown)


circuit-types/src/traits.rs
```diff
@@ -427,30 +248,28 @@ pub trait SecretShareType: Sized + BaseType {
 }
 
 /// Implementing types represent a secret share allocated in a constraint system
-pub trait SecretShareVarType<L: LinearCombinationLike>: Sized + CircuitVarType<L> {
+pub trait SecretShareVarType: Sized + CircuitVarType {
     /// The base type that this secret share is a representation of
-    type Base: CircuitVarType<LinearCombination>;
-    /// The type that results from blinding or unblinding the share
-    type BlindType: CircuitVarType<LinearCombination>;
```
akirillo, line 434: Why do we no longer represent the blinded/unblinded type as a separate type? Was this just not really necessary in the first place?

circuit-types/src/traits.rs
```diff
@@ -1078,87 +554,75 @@ pub trait SingleProverCircuit {
...
     fn verify(
-        witness_commitment: <Self::Witness as CircuitBaseType>::CommitmentType,
         statement: Self::Statement,
-        proof: R1CSProof,
-        bp_gens: &BulletproofGens,
-        mut verifier: Verifier,
+        proof: Proof<SystemCurve>,
+        vk: &VerifyingKey<SystemCurve>,
     ) -> Result<(), VerifierError> {
         // Commit to the witness and statement
```
akirillo, line 597: Nit: we shouldn't mention the witness here, and we shouldn't say that we're "committing" to the statement, perhaps just "Serialize the statement into scalars"

akirillo (approved):
couple nits & qs, nothing blocking tho - very nice

circuit-types/src/traits.rs
```diff
@@ -13,53 +13,45 @@
...
+use mpc_relation::{constraint_system::Circuit, PlonkCircuit as GenericPlonkCircuit, Variable};
 use num_bigint::BigUint;
-use rand::{thread_rng, CryptoRng, RngCore};
+use rand::thread_rng;
 use renegade_crypto::fields::{biguint_to_scalar, scalar_to_biguint, scalar_to_u64};
 
 use crate::{
     errors::{MpcError, ProverError, VerifierError},
-    AuthenticatedLinkableCommitment, LinkableCommitment,
+    Fabric,
```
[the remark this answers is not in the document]
joeykraut: Yeah this type is defined in `lib.rs` which isn't building yet so I didn't commit it

circuit-types/src/traits.rs
joeykraut: They were different only for dalek-specific reasons; i.e. adding two `Variable`s created a `LinearCombination` which is a different runtime type, these separate associated types realize this at higher layers. Because `Variable` is "closed" under addition we no longer have to do this

PR #238, opened 2023-11-09
circuit-macros: Refactor crate wrt new trait definitions

joeykraut:
### Purpose
This PR refactors the macros defined in `circuit-macros` over the new trait definitions. These macros automatically derive proof system, mpc, and secret share types for a base type.

### Tests
- Unit tests pass, including the macro tests


circuit-macros/src/circuit_type/secret_share_types.rs
```diff
@@ -141,58 +127,15 @@ fn build_addition_impl(base_type: &ItemStruct) -> TokenStream2 {
...
             fn add(self, rhs: #secret_share_type_name) -> Self::Output {
-                self.add_shares(rhs)
-            }
-        }
-    };
-    impl_block.to_token_stream()
-}
-
-/// Build an addition implementation on secret share variable types
-fn build_var_type_addition_impl(secret_share_type: &ItemStruct) -> TokenStream2 {
```
akirillo, line 160: Out of curiosity, why do we remove this `Add` impl? We still have variable types for secret-shared objects allocated in the proof system, so this could still be used, right?

akirillo (approved):
lgtm, super nice

just one small q about deriving `Add` impls

circuit-macros/src/circuit_type/secret_share_types.rs
joeykraut: We need access to the circuit to add variables now (dalek just generated a linear combination). We can't pass the circuit into the Add interface, and it's a bit too complicated to make all variables hold references to the circuit, so we just remove the Add impl and use add_shares instead

PR #240, opened 2023-11-11
circuit-types: Refactor application level types over new proof system

joeykraut:
### Purpose
This PR finishes refactoring all types in the `circuit-types` crate over the new proof system. This largely involves deleting code and fixing imports where traits have changed.

### Testing
- Unit tests pass

### DIFF
--- a/circuit-types/Cargo.toml
+++ b/circuit-types/Cargo.toml
@@ -10,6 +10,7 @@ edition = "2021"
 ark-ff = "0.4"
 ark-ec = "0.4"
 ark-mpc = { workspace = true }
+ed25519-dalek = "1.0.1"
 mpc-plonk = { workspace = true }
 mpc-relation = { workspace = true }
 
--- a/circuit-types/src/lib.rs
+++ b/circuit-types/src/lib.rs
@@ -19,16 +19,17 @@ pub mod traits;
 pub mod transfers;
 pub mod wallet;
 
+use ark_ff::BigInt;
+use ark_mpc::MpcFabric;
 use bigdecimal::Num;
-use constants::{MAX_BALANCES, MAX_FEES, MAX_ORDERS, MERKLE_HEIGHT};
+use constants::{
+    Scalar, ScalarField, SystemCurveGroup, MAX_BALANCES, MAX_FEES, MAX_ORDERS, MERKLE_HEIGHT,
+};
 use fixed_point::DEFAULT_FP_PRECISION;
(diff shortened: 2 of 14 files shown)


circuit-types/src/lib.rs
```diff
@@ -37,13 +38,25 @@ use wallet::{Wallet, WalletShare};
 // | Constants |
 // -------------
 
+/// The zero value in the Scalar field we work over
+pub const SCALAR_ZERO: ScalarField = ScalarField::new(BigInt::new([0, 0, 0, 0]));
+/// The one value in the Scalar field we work over
+pub const SCALAR_ONE: ScalarField = ScalarField::new(BigInt::new([1, 0, 0, 0]));
```
akirillo, line 44: Confirming this is correct even in Montgomery form?

circuit-types/src/lib.rs
```diff
@@ -318,8 +259,11 @@ pub mod native_helpers {
     {
         // Sample new wallet blinders from the `blinder` CSPRNG
         // See the comments in `valid_reblind.rs` for an explanation of the two CSPRNGs
-        let mut blinder_samples =
-            evaluate_hash_chain(private_secret_shares.blinder, 2 /* length */);
+        let mut blinder_samples = evaluate_hash_chain(
+            private_secret_shares.blinder,
+            2, /* length
+                */
```
akirillo, line 265: was this `cargo fmt`?

circuit-types/Cargo.toml
```diff
@@ -10,6 +10,7 @@ edition = "2021"
 ark-ff = "0.4"
 ark-ec = "0.4"
 ark-mpc = { workspace = true }
+ed25519-dalek = "1.0.1"
```
akirillo, line 13: I think we can remove this?

akirillo (approved):
couple small questions

circuit-types/src/lib.rs
joeykraut: The `BigInt` constructor is not concerned with montgomery form, only the `Fp` (which is a `MontgomeryBackend`) will make such changes to the underlying bigint

circuit-types/Cargo.toml
joeykraut: No, we use it for keys for the moment

circuit-types/src/lib.rs
joeykraut: it was, i fixed it

PR #242, opened 2023-11-13
circuits: mpc-gadgets: bits, comparators, modulo: Refactor + add tests

joeykraut:
### Purpose
This PR refactors the `bits`, `comparators`, and `modulo` mpc gadgets and adds tests for previously untested (or only tested in integration tests) gadgets. Some gadgets are tested implicitly through others.

### Testing
- Unit tests pass
- Added tests for refactored gadgets

### DIFF
--- a/circuits/src/mpc_gadgets/comparators.rs
+++ b/circuits/src/mpc_gadgets/comparators.rs
@@ -2,10 +2,10 @@
 
 use std::iter;
 
-use mpc_stark::{
-    algebra::{authenticated_scalar::AuthenticatedScalarResult, scalar::Scalar},
-    MpcFabric,
-};
+use circuit_types::Fabric;
+use constants::{AuthenticatedScalar, Scalar};
+
+use crate::{SCALAR_BITS_MINUS_TWO, SCALAR_MAX_BITS};
 
 use super::{
     arithmetic::product,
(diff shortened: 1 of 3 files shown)


circuits/src/mpc_gadgets/comparators.rs
```diff
@@ -15,28 +15,33 @@ use super::{
...
+///
+/// So we enforce that the values are less than 2^252
 pub fn less_than_zero<const D: usize>(
-    a: &AuthenticatedScalarResult,
-    fabric: &MpcFabric,
-) -> AuthenticatedScalarResult {
+    a: &AuthenticatedScalar,
+    fabric: &Fabric,
+) -> AuthenticatedScalar {
+    assert!(D <= SCALAR_MAX_BITS - 2, "D mus");
```
akirillo, line 31: Could we finish this assert statement? "D must have <= 252 bits"?

circuits/src/mpc_gadgets/comparators.rs
```diff
@@ -15,28 +15,33 @@ use super::{
...
     // Truncate the first 250 bits of the input
```
akirillo, line 32:
```suggestion
    // Truncate the first 252 bits of the input
```

circuits/src/mpc_gadgets/comparators.rs
akirillo: Also, it looks like `SCALAR_MAX_BITS = 253`, so this enforces that the input bitlength <= 251, not 252

circuits/src/mpc_gadgets/comparators.rs
```diff
@@ -15,28 +15,33 @@ use super::{
...
-    let truncated = truncate::<250>(a, fabric);
+    let truncated = truncate::<252>(a, fabric);
```
akirillo, line 33: nit: we should constantize this, and use the same constant in the assertion above

circuits/src/mpc_gadgets/comparators.rs
```diff
@@ -15,28 +15,33 @@ use super::{
...
 
-    // Because the Starknet scalar field is a prime field of order slightly greater
-    // than 2^251 values are negative if either their 251st bit or 252nd bit are
-    // set. Therefore, we truncate all bits below this and compare the value to
-    // zero.
+    // Because the our scalar field (Bn254) is a prime field of order greater than
```
akirillo, line 35:
```suggestion
    // Because our scalar field (Bn254) is a prime field of order greater than
```

akirillo (approved):
looks great, just some pesky nits from me

PR #245, opened 2023-11-13
circuits: zk-gadgets: poseidon: Refactor and add unit tests

joeykraut:
### Purpose
This PR refactors the Poseidon gadget to work over the new `mpc-plonk` system. This gadget also works for both single and multiprovers. I added a test for coherence with the sponge defined in `renegade-crypto`.

### Testing
- Unit tests pass

### DIFF
--- a/circuits/src/zk_gadgets/poseidon.rs
+++ b/circuits/src/zk_gadgets/poseidon.rs
@@ -1,19 +1,14 @@
 //! Groups logic for adding Poseidon hash function constraints to a Bulletproof
 //! constraint system
 
-use circuit_types::{
-    errors::ProverError,
-    traits::{LinearCombinationLike, MpcLinearCombinationLike},
-};
+use ark_ff::One;
+use constants::ScalarField;
 use itertools::Itertools;
-use mpc_bulletproof::{
-    r1cs::{LinearCombination, RandomizableConstraintSystem},
-    r1cs_mpc::{MpcLinearCombination, MpcRandomizableConstraintSystem, MpcVariable, R1CSError},
+use mpc_plonk::errors::PlonkError;
+use mpc_relation::{constants::GATE_WIDTH, ConstraintSystem, Variable};
+use renegade_crypto::hash::{
+    CAPACITY, FULL_ROUND_CONSTANTS, PARTIAL_ROUND_CONSTANTS, RATE, R_F, R_P, WIDTH as SPONGE_WIDTH,
 };
-use mpc_stark::{algebra::scalar::Scalar, MpcFabric};
-use renegade_crypto::hash::PoseidonParams;
-
-use super::arithmetic::{ExpGadget, MultiproverExpGadget};
 
 // -----------------------
 // | Singleprover Gadget |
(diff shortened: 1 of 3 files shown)


circuits/src/zk_gadgets/poseidon.rs
```diff
@@ -142,451 +127,217 @@ impl PoseidonHashGadget {
...
 
-    /// Permute the digest by applying the Poseidon round function
-    fn permute<CS: RandomizableConstraintSystem>(&mut self, cs: &mut CS) -> Result<(), R1CSError> {
-        // Compute full_rounds / 2 rounds in which the sbox is applied to all elements
-        for round in 0..self.params.full_rounds / 2 {
-            self.add_round_constants(round);
-            self.apply_sbox(true /* full_round */, cs)?;
-            self.apply_mds()?;
+    /// Permute the state using the Poseidon 2 permutation
+    #[allow(clippy::missing_docs_in_private_items)]
```
akirillo, line 153: nit: do we still need this lint allowance? there's a doc comment here now

circuits/src/zk_gadgets/poseidon.rs
```diff
@@ -142,451 +127,217 @@ impl PoseidonHashGadget {
...
-        if self.next_index == self.params.rate {
-            self.permute(cs)?;
-            self.next_index = 0;
+        cs: &mut C,
+    ) -> Result<(), PlonkError> {
+        let coeffs = [ScalarField::one(); GATE_WIDTH];
+        let in_wires = self.state.clone();
+        for state_elem in self.state.iter_mut() {
+            let lc_wires = [*state_elem, in_wires[0], in_wires[1], in_wires[2]];
+            *state_elem = cs.lc(&lc_wires, &coeffs)?;
```
akirillo, line 231: would it be more appropriate to use `cs.sum` here?

circuits/src/zk_gadgets/poseidon.rs
```diff
@@ -142,451 +127,217 @@ impl PoseidonHashGadget {
...
-            )?
+        cs: &mut C,
+    ) -> Result<(), PlonkError> {
+        let mut coeffs = [ScalarField::one(); GATE_WIDTH];
+        let in_wires = self.state.clone();
+
+        // The first two elements of the state
+        for state_elem in self.state[..SPONGE_WIDTH - 1].iter_mut() {
+            let lc_wires = [*state_elem, in_wires[0], in_wires[1], in_wires[2]];
+            *state_elem = cs.lc(&lc_wires, &coeffs)?;
```
akirillo, line 289: here as well, would it make more sense to use `cs.sum`?

akirillo (approved):
couple nits, lgtm

circuits/src/zk_gadgets/poseidon.rs
joeykraut: Yes, its for the constant in the function

circuits/src/zk_gadgets/poseidon.rs
joeykraut: No, `cs.sum` will break up the input into groups of 3 inputs at a time, so this would require two gates instead of one.

PR #247, opened 2023-11-14
circuit-types: Add boolean type and `BoolVar`, `AuthenticatedBool` analogs

joeykraut:
### Purpose
This PR introduces a boolean type to MPC and ZK circuits. This necessitates that variable types have a reference to the underlying constraint system when they are deserialized, so that they may apply constraints on their underlying variables. For the boolean type this adds a constraint that the value is boolean (i.e. $x(1 -x) \stackrel{?}{=} 0$)

### Testing
- Unit tests pass

### DIFF
--- a/circuit-types/src/traits.rs
+++ b/circuit-types/src/traits.rs
@@ -18,7 +18,7 @@
 
 use ark_mpc::{algebra::AuthenticatedScalarResult, network::PartyId};
 use async_trait::async_trait;
-use constants::{AuthenticatedScalar, Scalar, SystemCurve};
+use constants::{AuthenticatedScalar, Scalar, ScalarField, SystemCurve};
 use futures::future::join_all;
 use itertools::Itertools;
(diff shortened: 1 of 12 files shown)


circuit-types/src/traits.rs
```diff
@@ -394,15 +444,40 @@ impl CircuitVarType for Variable {
...
+    }
+
+    fn from_vars<I: Iterator<Item = Variable>, C: ConstraintSystem<ScalarField>>(
+        i: &mut I,
+        cs: &mut C,
+    ) -> Self {
+        let var = i.next().unwrap();
+        cs.enforce_bool(var).unwrap();
+
+        BoolVar::new_unchecked(var)
```
joeykraut, line 469: the warning that `new_unchecked` is private is fixed by an analogous PR in `mpc-jellyfish`

akirillo (approved):

PR #248, opened 2023-11-15
circuits: zk-gadgets: select: Refactor and add unit tests

joeykraut:
### Purpose
This PR refactors the conditional select gadgets over the new `mpc-jellyfish` system. This makes use of the `mux` method introduced to the `ConstraintSystem` trait. As well, this PR adds unit tests for the gadgets.

### Testing
- Unit tests pass


circuits/src/zk_gadgets/select.rs
```diff
@@ -96,58 +36,26 @@ impl MultiproverCondSelectGadget {
...
-                cs,
-            ));
-        }
-
-        selected
-    }
-}
-
-/// A multiprover variant of the CondSelectVectorGadget
-pub struct MultiproverCondSelectVectorGadget;
```
akirillo, line 124: just making sure my understanding is correct: we don't need a `MultiProverCondSelectVectorGadget` anymore because the implementation differences b/t single/multiprover for the mux gate are handled by the constraint system, over which the gadget is generic, right?

if so, is the general high-level goal to have all gadgets be implemented in a way that is agnostic of a single/multiprover setting, and have this abstracted entirely by the choice of constraint system?

akirillo (approved):
lgtm, really like the abstractions

circuits/src/zk_gadgets/select.rs
joeykraut: That is correct, the `mux` implementation defines the differences. I'd say this is _generally_ the goal, but it won't be achieved for all gadgets. Some are really hard to do in an MPC so it's not worth implementing just for generic functionality. 

But generally yes, note that those already done (poseidon, bits, etc) are all generic over the proof system

PR #249, opened 2023-11-16
circuit-types, circuit-macros: Remove `ConstraintSystem` use

joeykraut:
### Purpose
This PR removes the use of `ConstraintSystem` in the `circuit-types` and `circuit-macros` crates following changes to the upstream to merge `ConstraintSystem` and `Circuit`.

### Testing
- All tests pass

### DIFF
--- a/circuit-types/src/balance.rs
+++ b/circuit-types/src/balance.rs
@@ -5,7 +5,7 @@ use std::ops::Add;
 
 use circuit_macros::circuit_type;
 use constants::{AuthenticatedScalar, Scalar, ScalarField};
-use mpc_relation::{ConstraintSystem, Variable};
+use mpc_relation::{traits::Circuit, Variable};
 use num_bigint::BigUint;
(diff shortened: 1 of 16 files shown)


circuit-types/src/balance.rs
```diff
@@ -5,7 +5,7 @@ use std::ops::Add;
 
 use circuit_macros::circuit_type;
 use constants::{AuthenticatedScalar, Scalar, ScalarField};
-use mpc_relation::{ConstraintSystem, Variable};
+use mpc_relation::{traits::Circuit, Variable};
```
akirillo, line 8: getting unresolved import errors for `traits`, is it a public module?

akirillo (approved):
lgtm other than import errors

circuit-types/src/balance.rs
joeykraut: yes, it's just dependent on the other PR

PR #250, opened 2023-11-16
arbitrum-client: Initialize Arbitrum client

akirillo:
This PR introduces a new crate, `arbitrum-client`, and establishes some scaffolding for a configurable client used to interact with our contracts on Arbitrum.

The client is broadly based on the `starknet-client`, but makes many simplifications, both to avoid premature optimization, and because a lot of functionality is handled by `ethers-rs`.

The highest-level changes are the following:
- Client only manages a single Arbitrum account
- Client keeps a handle on an `ethers` `ContractInstance`, which provides bindings for contract methods, events, etc., with an embedded RPC client


arbitrum-client/src/client/config.rs
```diff
@@ -0,0 +1,87 @@
...
+
+pub struct ArbitrumClientConfig {
+    /// The address of the darkpool proxy contract.
+    ///
+    /// This is the main entrypoint to interaction with the darkpool.
+    pub proxy_address: String,
+    /// The address of the darkpool implementation contract.
+    ///
+    /// This is used to filter for events emitted by the darkpool.
+    pub implementation_address: String,
```
joeykraut, line 25: Again let's hide the fact that this is proxied, perhaps `event_source` or similar if it's only used for events.

arbitrum-client/src/client/config.rs
```diff
@@ -0,0 +1,87 @@
...
+    /// The HTTP-addressable URL of the Arbitrum RPC endpoint
+    pub rpc_url: String,
+    /// The private key of the account to use for signing transactions
+    pub arb_priv_key: String,
+}
+
+#[derive(Clone, Debug)]
+pub enum ArbitrumClientConfigError {
+    RpcClientInitialization(String),
+    AddressParsing(String),
```
joeykraut, line 35: These need doc comments, the lint I asked you to add will suggest this

arbitrum-client/src/client/config.rs
```diff
@@ -0,0 +1,87 @@
...
+}
+
+impl Display for ArbitrumClientConfigError {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "{self:?}")
+    }
+}
+impl Error for ArbitrumClientConfigError {}
```
joeykraut, line 43: Let's put this in `error.rs` to follow convention elsewhere

arbitrum-client/src/client/config.rs
```diff
@@ -0,0 +1,87 @@
...
+
+    fn get_proxy_address(&self) -> Result<Address, ArbitrumClientConfigError> {
+        Address::from_str(&self.proxy_address)
+            .map_err(|e| ArbitrumClientConfigError::AddressParsing(e.to_string()))
+    }
+
+    pub fn get_implementation_address(&self) -> Result<Address, ArbitrumClientConfigError> {
+        Address::from_str(&self.implementation_address)
+            .map_err(|e| ArbitrumClientConfigError::AddressParsing(e.to_string()))
+    }
```
joeykraut, line 77: Do we ever need these fields as strings? Probably best to store them as `Address` to avoid expensive string parsing when needed

joeykraut (approved):

arbitrum-client/src/client/config.rs
akirillo: I'm just thinking ahead to how this config object will be constructed - at the end of the day, it will come from CLI args which come in as strings. So at some point we will need to do string parsing, and it makes sense to me to keep this config struct lightweight to instantiate, and derive rich types for the full client struct.

Fwiw this is also the precedent set in the `StarknetClientConfig`

joeykraut: Oh gotcha, if it's only parsed once then it doesn't matter

PR #253, opened 2023-11-17
arbitrum-client: add contract interaction methods

akirillo:
This PR adds bindings for the darkpool contract's external methods for the Arbitrum client, including utilities for converting between relayer-defined and smart-contract-defined types (down the line, would be nice to unify these types, but a bit of a distraction at the moment). One thing to note is that the interfaces/type definitions are currently a bit muddled between relayer-defined and smart-contract-defined types, in that we use the smart-contract-defined statement types directly since the relayer-defined statement types are not yet adapted to Plonk (which also felt out of scope for this PR).

This should achieve parity with the `contract_interaction` module in `starknet-client` save for the `get_public_blinder_tx` method, which in our case is no longer an external method on the smart contract, and now has to be implemented via event indexing, which I will defer to the next PR.


Cargo.toml
```diff
@@ -72,3 +72,7 @@ itertools = "0.10"
 serde = { version = "1.0.139" }
 serde_json = "1.0.64"
 tracing = "0.1"
+
+[patch.crates-io]
```
akirillo, line 76: I'm not thrilled about this patch section, but we need it to be able to import types from `renegade-contracts::common`. I'm hoping to be able to remove the need for this when we unify types later, but if it's concerning, I can try thinking about this now

joeykraut: Why not just mock them out? This has a pretty large impact on the relayer as a whole and is not worth what is a relatively minor benefit in "having the types be the same".

arbitrum-client/src/client/contract_interaction.rs
```diff
@@ -0,0 +1,156 @@
...
+    /// Check whether the given Merkle root is a valid historical root
+    pub async fn check_merkle_root_valid(
+        &self,
+        root: MerkleRoot,
+    ) -> Result<bool, ArbitrumClientError> {
+        self.darkpool_contract
+            .root_in_history(serialize_calldata(&SerdeScalarField(root.inner()))?)
+            .call()
+            .await
+            .map_err(|e| ArbitrumClientError::ContractInteraction(e.to_string()))
```
joeykraut, line 48: nit: but i think this is something that should be done more generally; these one liners with lots of nested expressions are hard to read and reason about, I suggest:
```rust
let calldata = serialize_calldata(...);
self.darkpool_contract.root_in_history(calldata)...
```
Same with below; it is in general hard to read to put everything in a chain of method calls.

arbitrum-client/src/client/contract_interaction.rs
```diff
@@ -0,0 +1,156 @@
...
+                serialize_calldata(&proof.to_contract_type()?)?,
+                serialize_calldata(&valid_wallet_create_statement)?,
+            )
+            .send()
+            .await
+            .map_err(|e| ArbitrumClientError::ContractInteraction(e.to_string()))?
+            .await
+            .map_err(|e| ArbitrumClientError::ContractInteraction(e.to_string()))?;
+
+        Ok(())
```
joeykraut, line 91: Same here, it's way more readable to introduce intermediate variable names so that the reader can understand the result of intermediate computations! It certainly takes longer to read this and understand what's going on than if we separately assigned the arguments.

joeykraut, line 91: Just return, no need to separately return `Ok(()`. If the method has a return type you can `.map(|_| ())` same below

constants/src/lib.rs
```diff
@@ -62,6 +64,12 @@ pub type CurvePoint = GenericCurvePoint<SystemCurveGroup>;
 #[cfg(feature = "mpc-types")]
 pub type AuthenticatedScalar = AuthenticatedScalarResult<SystemCurveGroup>;
 
+/// The (unbatched) proof type that the proof system generates
+pub type SystemProof = Proof<SystemCurve>;
```
joeykraut, line 68: This probably belongs better elsewhere, I think it's okay to put it in the Arbitrum client for now. My reasoning: this does not parameterize the system in any way, it is _parameterized by_ other system parameters. `constants` should be kept really lean and have few dependencies -- essentially only to specify system parameters.

constants/src/lib.rs
```diff
@@ -62,6 +64,12 @@ pub type CurvePoint = GenericCurvePoint<SystemCurveGroup>;
...
+
+/// The polynomial commitment type that the proof system utilizes
+pub type SystemCommitment = Commitment<SystemCurve>;
```
joeykraut, line 71: Same here, does not belong in constants IMO, `circuit-types` may be a better spot. Also would not name these `SystemCommitment` as they do not parameterize the system. Maybe `PlonkProof` (analog of `PlonkCircuit` already defined in `circuits`) and `PolynomialCommitment` to give specificity

arbitrum-client/src/conversion.rs
```diff
@@ -0,0 +1,61 @@
...
+        .collect::<Vec<_>>()
+        .try_into()
+        .map_err(|_| ConversionError::InvalidLength)
+}
+
+/// Trait for converting from relayer types to their smart contract analogues
+pub trait ToContractType<T> {
+    /// Convert the type to its smart contract analogue
+    fn to_contract_type(self) -> Result<T, ConversionError>;
+}
```
joeykraut, line 28: Why does this need a trait? Just use `Into`/`From`

Cargo.toml
akirillo: it's more about not having to redefine serialization for the types than just being able to import them from elsewhere. the serialization derivation in `renegade-contracts` is quite hairy, redefining it here increases the surface area of potential inconsistencies... but I agree that I don't want this patch to stick around, so I'll redefine serde stuff + the types in the relayer for now, and this may be a good base / source of truth for the unification later

constants/src/lib.rs
akirillo: sounds good, will move it into `circuit-types` along with the commitment type below if that works - just feels like a more appropriate home than the Arbitrum client

arbitrum-client/src/conversion.rs
akirillo: `Into`/`From` is a bit unwieldy here because we can't implement it directly between `mpc-jellyfish` and `renegade-contracts` types, so I would need to introduce wrapper types, which muddies the interface / readability more than I think is worth it to just use canonical traits (since we only do this conversion in one place, explicitly, anyway)

But if I'm mocking out the `renegade-contracts` types here anyway then that should no longer be an issue

arbitrum-client/src/serde_def_types.rs
akirillo, line 1: This file is directly copied over from https://github.com/renegade-fi/renegade-contracts/blob/main/common/src/serde_def_types.rs, with unused types removed and additional comments added

arbitrum-client/src/types.rs
akirillo, line 1: This file is directly copied over from https://github.com/renegade-fi/renegade-contracts/blob/main/common/src/types.rs, with unused types removed and additional comments added

arbitrum-client/src/client/contract_interaction.rs
```diff
@@ -0,0 +1,199 @@
...
+    /// Get the current Merkle root in the contract
+    pub async fn get_merkle_root(&self) -> Result<ScalarField, ArbitrumClientError> {
+        let merkle_root_bytes = self
+            .darkpool_contract
+            .get_root()
+            .call()
+            .await
+            .map_err(|e| ArbitrumClientError::ContractInteraction(e.to_string()))?;
+
+        Ok(deserialize_retdata::<SerdeScalarField>(&merkle_root_bytes)?.0)
```
joeykraut, line 36: nit: Add a comment here (maybe inline) for the `.0` just like we do when we pass magic numbers as arguments, it's unclear to the reader what the zeroth field is

joeykraut (approved):

PR #254, opened 2023-11-18
circuits: zk-gadgets: arithmetic, fixed-point, wallet-operations: Refactor and add unit tests

joeykraut:
### Purpose
This PR refactors the `arithmetic`, `fixed-point`, and `wallet-operations` modules of the `zk-gadgets` to work over the new `mpc-jellyfish` system. I also added unit tests for all refactored gadgets, some are tested implicitly.

### Testing
- Unit tests pass

### DIFF
--- a/circuits/src/zk_gadgets/arithmetic.rs
+++ b/circuits/src/zk_gadgets/arithmetic.rs
@@ -1,21 +1,19 @@
 //! Groups ZK gadgets used as arithmetic primitives in more complicated
 //! computations
 
-use ark_ff::Zero;
-use circuit_types::errors::ProverError;
+use ark_ff::{One, Zero};
+use circuit_types::traits::{CircuitBaseType, CircuitVarType};
+use circuit_types::PlonkCircuit;
+use constants::ScalarField;
+use mpc_relation::errors::CircuitError;
+use mpc_relation::traits::Circuit;
+use mpc_relation::Variable;
 use renegade_crypto::fields::{biguint_to_scalar, scalar_to_biguint};
 
-use circuit_types::traits::{LinearCombinationLike, MpcLinearCombinationLike};
-use mpc_bulletproof::r1cs::{LinearCombination, RandomizableConstraintSystem, Variable};
-use mpc_bulletproof::r1cs_mpc::{MpcLinearCombination, MpcRandomizableConstraintSystem, R1CSError};
-use mpc_stark::algebra::scalar::Scalar;
-use mpc_stark::MpcFabric;
 use num_bigint::BigUint;
 use num_integer::Integer;
 
-use super::bits::ToBitsGadget;
-use super::comparators::{EqZeroGadget, LessThanGadget};
-use super::select::CondSelectGadget;
+use super::comparators::LessThanGadget;
 
 // -------------------------
 // | Single Prover Gadgets |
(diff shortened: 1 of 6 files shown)


circuits/src/zk_gadgets/fixed_point.rs
```diff
@@ -101,46 +101,147 @@ pub struct MultiproverFixedPointGadget;
...
-            diff,
-            fabric,
-            cs,
+        let diff = cs.lc(
+            &[one_var, lhs_minus_rhs, zero_var, zero_var],
+            &[*TWO_TO_M_SCALAR, -one, one, one],
+        )?;
+
+        MultiproverGreaterThanEqZeroGadget::<DEFAULT_FP_PRECISION>::constrain_greater_than_zero(
+            diff, fabric, cs,
```
akirillo, line 131: this seems to be a different constraint than what was there previously - seems like it used to be `2^m - 1 > (lhs - rhs*2^m)`, and is now `2^m - (lhs - rhs*2^m) > 0`. am i reading this wrong?

akirillo (approved):
some small concerns but nothing that requires further discussion so approving to unblock

circuits/src/zk_gadgets/arithmetic.rs
```diff
@@ -29,347 +27,122 @@ use super::select::CondSelectGadget;
...
-            let (_, _, out_var1) = cs.multiply(recursive_result.clone(), recursive_result);
-            let (_, _, out_var2) = cs.multiply(out_var1.into(), x_lc);
-            out_var2.into()
-        }
-    }
-}
-
-/// An exponentiation gadget on a private exponent; compute x^\alpha
-#[derive(Clone, Debug)]
-pub struct PrivateExpGadget<const ALPHA_BITS: usize> {}
```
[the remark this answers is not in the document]
joeykraut: correct, it was for elgamal which is gone

circuits/src/zk_gadgets/arithmetic.rs
```diff
@@ -29,347 +27,122 @@ use super::select::CondSelectGadget;
...
+    ) -> Result<(Variable, Variable), CircuitError> {
+        let a_bigint = scalar_to_biguint(&a.eval(cs));
+        let b_bigint = scalar_to_biguint(&b.eval(cs));
 
         // Compute the divrem outside of the circuit
-        // Verifier evals all wires to zero -- it only knows commitments
-        // Handle this case explicitly
         let (q, r) = if b_bigint == BigUint::zero() {
-            (BigUint::zero(), b_bigint)
+            (BigUint::zero(), a_bigint)
```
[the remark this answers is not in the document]
joeykraut: The operation is perfectly well defined when `b = 0`, this was a bug before. To be clear, setting `r = 0` does not _make_ the constraints fail. Remember that the prover may assign these wires however they like, it is simply that we assign them honestly in our codebase. The constraints will be able to be satisfied when `b = 0` and that is okay from our perspective

circuits/src/zk_gadgets/fixed_point.rs
joeykraut: It was `2^m - 1 >= (lhs - rhs << m)`, rearranging the terms gives:
`2^m - (lhs - rhs*2^m) - 1 >= 0`. Ours is a discrete domain so this is equivalent to the constraint:
`2^m - (lhs - rhs * 2^m) > 0` which is what we have here

akirillo: ah, i missed the `>=`, thank you

PR #256, opened 2023-11-18
circuits: zk-circuits: valid-wallet-create: Refactor over `mpc-plonk`

joeykraut:
### Purpose
This PR makes the changes necessary to support full ZK prove/verify flows and uses this to refactor the `VALID WALLET CREATE` circuit. This includes structuring how the proving and verifying keys fit into the `SingleProverCircuit` and `MultiProverCircuit` traits.

### Testing
- Unit tests pass

### DIFF
--- a/circuit-types/src/traits.rs
+++ b/circuit-types/src/traits.rs
@@ -34,6 +34,7 @@ use mpc_relation::{traits::Circuit, BoolVar, Variable};
 use num_bigint::BigUint;
 use rand::thread_rng;
 use renegade_crypto::fields::{biguint_to_scalar, scalar_to_biguint, scalar_to_u64};
+use std::sync::Arc;
 
 use crate::{
     errors::{MpcError, ProverError, VerifierError},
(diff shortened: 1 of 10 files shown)


circuit-types/src/traits.rs
```diff
@@ -669,14 +696,44 @@ impl<const N: usize, T: SecretShareVarType> SecretShareVarType for [T; N] {
...
     /// The witness type, given only to the prover, which generates a blinding
     /// commitment that can be given to the verifier
     type Witness: CircuitBaseType;
     /// The statement type, given to both the prover and verifier, parameterizes
     /// the underlying NP statement being proven
     type Statement: CircuitBaseType;
 
+    /// Returns a reference to the proving key for the circuit
+    #[cfg(not(feature = "test-helpers"))]
+    fn proving_key() -> Rc<ProvingKey<SystemCurve>> {
```
akirillo, line 709: Should this (and the other references to proving/verifying keys) be `Arc`s? I forget if we have proofs of the same circuit being generated on multiple threads

akirillo (approved):
so clean, nice

PR #258, opened 2023-11-19
circuits: zk-circuits: valid-commitments: Refactor over `mpc-plonk`

joeykraut:
### Purpose
This PR refactors the `VALID COMMITMENTS` circuit over `mpc-plonk`. This circuit is proved once per order and allows a relayer to augment a wallet with balances for received mints in a trade.

### Testing
- Unit tests pass

### DIFF
--- a/circuits/src/zk_circuits/valid_commitments.rs
+++ b/circuits/src/zk_circuits/valid_commitments.rs
@@ -10,27 +10,21 @@
 //! VALID COMMITMENTS is proven once per order in the wallet
 
 use crate::{
-    zk_gadgets::{
-        comparators::EqGadget,
-        gates::{AndGate, ConstrainBinaryGadget, OrGate},
-        select::CondSelectVectorGadget,
-    },
+    zk_gadgets::{comparators::EqGadget, select::CondSelectVectorGadget},
     SingleProverCircuit,
 };
 use circuit_macros::circuit_type;
 use circuit_types::{
-    balance::{BalanceVar, LinkableBalance},
-    fee::{FeeVar, LinkableFee},
-    order::{LinkableOrder, OrderVar},
-    traits::{
-        BaseType, CircuitBaseType, CircuitCommitmentType, CircuitVarType, LinearCombinationLike,
-    },
-    wallet::{LinkableWalletShare, WalletVar},
+    balance::{Balance, BalanceVar},
+    fee::{Fee, FeeVar},
+    order::{Order, OrderVar},
+    traits::{BaseType, CircuitBaseType, CircuitVarType, SecretShareVarType},
+    wallet::{WalletShare, WalletVar},
+    PlonkCircuit,
 };
-use constants::{MAX_BALANCES, MAX_FEES, MAX_ORDERS};
-use mpc_bulletproof::r1cs::{LinearCombination, R1CSError, RandomizableConstraintSystem, Variable};
-use mpc_stark::algebra::{scalar::Scalar, stark_curve::StarkPoint};
-use rand::{CryptoRng, RngCore};
+use constants::{Scalar, ScalarField, MAX_BALANCES, MAX_FEES, MAX_ORDERS};
+use mpc_plonk::errors::PlonkError;
+use mpc_relation::{errors::CircuitError, traits::Circuit, BoolVar, Variable};
 use serde::{Deserialize, Serialize};
 
 // ----------------------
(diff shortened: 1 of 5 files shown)


circuits/src/zk_circuits/valid_commitments.rs
```diff
@@ -86,110 +90,103 @@ where
...
 
         // Verify that the fee is contained in the wallet
-        Self::contains_fee(witness.fee, &augmented_wallet, cs);
-    }
-
-    /// Constrains the order input to the matching engine is valid
-    fn constrain_valid_order<CS: RandomizableConstraintSystem>(
-        order: &OrderVar<Variable>,
-        cs: &mut CS,
-    ) {
```
joeykraut, line 130: moved this into the calling method above

circuits/src/zk_circuits/valid_commitments.rs
```diff
@@ -198,98 +195,97 @@ where
...
-            let balances_eq = EqGadget::eq(balance.clone(), target_balance.clone(), cs);
+            let index_mask = EqGadget::eq(&curr_index, &index, cs)?;
+            let balances_eq = EqGadget::eq(balance, target_balance, cs)?;
+
+            let found = cs.logic_and(index_mask, balances_eq)?;
 
-            let found = AndGate::and(index_mask, balances_eq, cs);
-            balance_found += found;
-            curr_index += Variable::One();
+            balance_found = cs.logic_or(balance_found, found)?;
```
akirillo, line 229: Using `logic_or` here feels more readable / semantically appropriate, but doesn't it make it such that we no longer constrain that a *single* balance is found? I.e. previously, if somehow two balances were matched, `balance_found = 2 =\= 1` would not satisfy the constraint, but now, `true || true == true` does satisfy

circuits/src/zk_circuits/valid_commitments.rs
```diff
@@ -198,98 +195,97 @@ where
...
-                EqGadget::eq::<LinearCombination, Variable, _, _, _>(curr_index.clone(), index, cs);
-            let orders_eq = EqGadget::eq(order.clone(), target_order.clone(), cs);
+            let index_mask = EqGadget::eq(&curr_index, &index, cs)?;
+            let orders_eq = EqGadget::eq(order, target_order, cs)?;
 
-            let found = AndGate::and(index_mask, orders_eq, cs);
-            order_found += found;
-            curr_index += Variable::One();
+            let found = cs.logic_and(index_mask, orders_eq)?;
+            order_found = cs.logic_or(order_found, found)?;
```
akirillo, line 251: Same concern as above for balances - is it ok if we potentially find multiple orders? I suppose the index mask should effectively prevent this though

circuits/src/zk_circuits/valid_commitments.rs
```diff
@@ -198,98 +195,97 @@ where
...
 
     /// Verify that the wallet contains the given balance at an unspecified
     /// index
-    fn contains_balance<CS: RandomizableConstraintSystem>(
-        target_balance: BalanceVar<Variable>,
-        wallet: &WalletVar<LinearCombination, MAX_BALANCES, MAX_ORDERS, MAX_FEES>,
-        cs: &mut CS,
-    ) {
-        let mut balance_found: LinearCombination = Variable::Zero().into();
+    fn contains_balance(
```
akirillo, line 260: nit: ik this was previously the fn name, but I think something like `constrain_balance_contained` would do better to indicate in caller fns that this adds a constraint, rather than returning a Boolean variable (also applies to other similarly named fns)

circuits/src/zk_circuits/valid_commitments.rs
```diff
@@ -198,98 +195,97 @@ where
...
+        wallet: &WalletVar<MAX_BALANCES, MAX_ORDERS, MAX_FEES>,
+        cs: &mut PlonkCircuit,
+    ) -> Result<(), CircuitError> {
+        let mut balance_found = cs.false_var();
         for balance in wallet.balances.iter() {
-            let balances_eq = EqGadget::eq(balance.clone(), target_balance.clone(), cs);
-            balance_found += balances_eq;
+            let balances_eq = EqGadget::eq(balance, target_balance, cs)?;
+
+            balance_found = cs.logic_or(balance_found, balances_eq)?;
```
akirillo, line 269: Considering this function in isolation, using `logic_or` definitely allows for multiple of the same balance. Ig it doesn't impact the stated purpose of this fn, as long as that's constrained elsewhere (or explicitly allowed) this is fine

akirillo (approved):
Looks good, just have some concerns about using logic or instead of addition

circuits/src/zk_circuits/valid_commitments.rs
joeykraut: the index mask makes these equivalent, the weird one is fees which i'll note in the comments.

PR #260, opened 2023-11-20
Andrew/event indexing

akirillo:
This PR adds the `get_public_blinder_tx`, `find_commitment_in_state`, and `find_merkle_authentication_path` event indexing methods to the Arbitrum client.

These follow the [Starknet client implementation](https://github.com/renegade-fi/renegade/blob/main/starknet-client/src/client/chain_state.rs#L141).

Implementation of the remaining indexing functionality, namely parsing wallet shares from calldata, is left for the next PR.


arbitrum-client/src/constants.rs
```diff
@@ -0,0 +1,64 @@
...
+    /// Testnet chain
+    Testnet,
+    /// Devnet chain
+    Devnet,
+}
+
+/// The RPC url for a devnet chain
+pub const DEVNET_RPC_URL: &str = "http://localhost:8547";
+/// The RPC url for a testnet chain
+pub const TESTNET_RPC_URL: &str = "https://stylus-testnet.arbitrum.io/rpc";
```
joeykraut, line 25: We probably want these configurable, hardcoding them is pretty brittle

arbitrum-client/src/constants.rs
```diff
@@ -0,0 +1,64 @@
...
+
+/// The block number at which the darkpool was deployed on devnet
+pub const DEVNET_DEPLOY_BLOCK: u64 = 0;
+/// The block number at which the darkpool was deployed on testnet
+pub const TESTNET_DEPLOY_BLOCK: u64 = 604069;
```
joeykraut, line 30: This may belong better in `constants` since it's so central to the system, I know this deviates from the `starknet-client`, but i feel it makes more sense that way

arbitrum-client/src/constants.rs
```diff
@@ -0,0 +1,64 @@
...
+
+/// The value of an empty leaf in the Merkle tree
```
joeykraut, line 32: Let's add an explanation of this constant, a nothing-up-my-sleeve number is shady without a good doc comment

arbitrum-client/src/client/event_indexing.rs
```diff
@@ -0,0 +1,105 @@
...
+                .map_err(|e| ArbitrumClientError::EventQuerying(e.to_string()))?;
+
+            let value = events
+                .last()
+                .map(|event| {
+                    postcard::from_bytes::<SerdeScalarField>(&event.new_value)
+                        .map_err(|e| ArbitrumClientError::Serde(e.to_string()))?
+                        .0 // Scalar
+                })
+                .ok_or(ArbitrumClientError::NodeChangeNotFound)?;
```
joeykraut, line 77: This shouldn't error, it should just do nothing, i.e. leave the default value in the authentication path.

arbitrum-client/src/errors.rs
```diff
@@ -12,7 +12,14 @@ pub enum ArbitrumClientError {
...
     /// Error thrown when converting between relayer & smart contract types
-    ConversionError(ConversionError),
+    Conversion(ConversionError),
+    /// Error thrown when querying events
+    EventQuerying(String),
+    /// Error thrown when a commitment can't be found in the Merkle tree
+    CommitmentNotFound,
+    /// Error thrown when an update can't be found for a specific node
+    /// in the Merkle tree
+    NodeChangeNotFound,
```
joeykraut, line 22: From my comment above, I don't think semantically this error should ever be thrown. If we can't find a node change event, it is implicitly the default value at that height. So unless I'm missing something we can remove this

arbitrum-client/src/helpers.rs
```diff
@@ -18,3 +22,10 @@ pub fn deserialize_retdata<'de, T: Deserialize<'de>>(
 ) -> Result<T, ArbitrumClientError> {
     postcard::from_bytes(retdata).map_err(|e| ArbitrumClientError::Serde(e.to_string()))
 }
+
+/// Computes the Keccak-256 hash of the serialization of a scalar,
+/// used for filtering events that have indexed scalar topics
+pub fn hash_scalar(scalar: Scalar) -> Result<H256, ArbitrumClientError> {
```
joeykraut, line 28: I'd propose renaming this to avoid name collisions + confusion elsewhere in the relayer codebase. Maybe something that indicates it's a keccak hash, like `keccak_hash_scalar`?

joeykraut (approved):
beautifully written

PR #261, opened 2023-11-20
arbitrum-client: fetch_public_shares_from_tx

akirillo:
This PR adds the `fetch_public_shares_from_tx` method to the Arbitrum client, allowing it to read a wallet's public secret shares from calldata.

Currently, this method only expects to parse shares from calldata where the {`new_wallet`, `update_wallet`, `process_match_settle`} call was a top-level call issued by an EOA, for simplicity. I've added a todo for parsing shares from calldata where the Darkpool function was called by another contract, I felt it was not worth trying to implement before having an integration testing stack.

This marks the end of the core functionality required by the client, the next PRs will focus on integration testing of the client, after which I'll focus on deprecating the Starknet client throughout the relayer in favor of the Arbitrum client.

### DIFF
--- a/arbitrum-client/src/client/event_indexing.rs
+++ b/arbitrum-client/src/client/event_indexing.rs
@@ -1,17 +1,29 @@
 //! Defines `ArbitrumClient` helpers that allow for indexing events
 //! emitted by the darkpool contract
 
+use alloy_sol_types::SolCall;
+use circuit_types::{traits::BaseType, SizedWalletShare};
 use common::types::merkle::{MerkleAuthenticationPath, MerkleTreeCoords};
 use constants::{Scalar, MERKLE_HEIGHT};
 use ethers::types::TxHash;
 use num_bigint::BigUint;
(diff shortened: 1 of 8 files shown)


arbitrum-client/src/client/event_indexing.rs
```diff
@@ -101,4 +110,81 @@ impl ArbitrumClient {
...
     }
+
+    /// Fetch and parse the public secret shares from the calldata of the given
+    /// transaction
+    ///
+    /// In the case that the referenced transaction is a `process_match_settle`,
+    /// we disambiguate between the two parties by adding the public blinder of
+    /// the party's shares the caller intends to fetch
+    // TODO: Add support for nested calls
+    pub async fn fetch_public_shares_from_tx(
```
joeykraut, line 121: nit: might be more readable to factor this into helper methods for each tx type

joeykraut (approved):
very nice

PR #262, opened 2023-11-21
circuits: zk-circuits: valid-match-mpc: Refactor over `mpc-plonk`

joeykraut:
### Purpose
This PR refactors the `VALID MATCH MPC` circuit over the `mpc-plonk` system. This method proves validity of a match that results from an MPC or the internal matching engine.

Note that we intend to combine this circuit with `VALID SETTLE`, this will come in a follow up PR.

Also deferred to a follow up are more extensive unit tests -- this circuit is tested heavily in integration tests, but given the new mocking interface for MPCs, it makes sense to test them in unit as well.

### Testing
- Unit tests pass
- Tested prove/verify

### DIFF
--- a/circuits/src/zk_circuits/valid_match_mpc.rs
+++ b/circuits/src/zk_circuits/valid_match_mpc.rs
@@ -4,46 +4,31 @@
 //! See the whitepaper (https://renegade.fi/whitepaper.pdf) appendix A.5
 //! for a formal specification
 
+use ark_ff::{One, Zero};
 use circuit_types::{
-    balance::{AuthenticatedBalanceVar, BalanceVar, LinkableBalance},
-    errors::ProverError,
-    fixed_point::{AuthenticatedFixedPointVar, FixedPoint, FixedPointVar, DEFAULT_FP_PRECISION},
-    order::{AuthenticatedOrderVar, LinkableOrder, OrderVar},
-    r#match::{AuthenticatedMatchResultVar, LinkableMatchResult, MatchResultVar},
+    balance::{Balance, BalanceVar},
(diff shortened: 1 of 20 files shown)


circuits/src/zk_circuits/valid_wallet_update.rs
```diff
@@ -312,11 +312,6 @@ where
         new_timestamp: Variable,
         cs: &mut PlonkCircuit,
     ) -> Result<(), CircuitError> {
-        // Ensure all orders have boolean side
-        for order in new_wallet.orders.iter() {
-            cs.enforce_bool(order.side)?;
-        }
```
joeykraut, line 318: Changing the type of `Order::side` to be boolean means this constraint is automatically generated when the type is allocated

circuits/src/zk_circuits/valid_match_mpc.rs
```diff
@@ -54,109 +39,100 @@ use rand::{CryptoRng, RngCore};
...
-        // Constrain the min_amount_order_index to be binary
-        // i.e. 0 === min_amount_order_index * (1 - min_amount_order_index)
-        let (_, _, mul_out) = cs
-            .multiply(
-                &witness.match_res.min_amount_order_index.clone().into(),
-                &(MpcLinearCombination::from_scalar(Scalar::one(), fabric.clone())
-                    - &witness.match_res.min_amount_order_index),
-            )
-            .map_err(ProverError::Collaborative)?;
-        cs.constrain(mul_out.into());
```
joeykraut, line 141: same here, changing the type to `bool` automatically generates this constraint

akirillo (approved):

PR #264, opened 2023-11-22
circuits: zk-circuits: valid-match-settle: Merge match and settle circuits

joeykraut:
### Purpose
This PR merges the match and settle circuits into one. This is necessary to enable proof linking across all proofs. This PR also refactors the circuits over the `mpc-plonk` system.

This circuit now validates the execution of the matching engine as well as the execution of the settlement logic into the two parties' wallets.

Along the way it was necessary to build a multiprover `==` gadget, this exposes an interface to test equality on only public values -- i.e. those that have been opened. We use this only on statement types (indices) so this is safe.

### Testing
- Unit tests pass

### DIFF
--- a/circuits/src/zk_circuits/valid_match_settle/multi_prover.rs
+++ b/circuits/src/zk_circuits/valid_match_settle/multi_prover.rs
@@ -6,16 +6,23 @@ use circuit_types::{
     fixed_point::{FixedPointVar, DEFAULT_FP_PRECISION},
     order::OrderVar,
     r#match::MatchResultVar,
+    wallet::WalletShareVar,
     Fabric, MpcPlonkCircuit, AMOUNT_BITS, PRICE_BITS,
 };
 use constants::ScalarField;
-use mpc_relation::{errors::CircuitError, traits::Circuit};
+use mpc_relation::{errors::CircuitError, traits::Circuit, Variable};
 
-use super::{ValidMatchSettle, ValidMatchSettleWitnessVar};
-use crate::zk_gadgets::{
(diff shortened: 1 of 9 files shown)


circuits/src/zk_circuits/valid_match_settle/multi_prover.rs
```diff
@@ -230,16 +241,185 @@ impl ValidMatchSettle {
...
+where
+    [(); MAX_BALANCES + MAX_ORDERS + MAX_FEES]: Sized,
+{
     /// Validate settlement of a match result into the wallets of the two
     /// parties
-    #[allow(clippy::needless_pass_by_ref_mut)]
     pub(crate) fn validate_settlement(
-        witness: &ValidMatchSettleWitnessVar,
-        statement: (),
+        witness: &ValidMatchSettleWitnessVar<MAX_BALANCES, MAX_ORDERS, MAX_FEES>,
```
joeykraut, line 252: this is copied from valid settle directly and refactored for use in `mpc-plonk`, same with the single prover impl

circuits/src/zk_gadgets/comparators.rs
```diff
@@ -288,6 +286,114 @@ impl<const D: usize> LessThanGadget<D> {
 // | Multiprover Gadgets |
 // -----------------------
 
+/// A multiprover version of the equal gadget
+pub struct MultiproverEqGadget;
+impl MultiproverEqGadget {
+    /// Computes whether the given input is equal to zero
+    ///
+    /// Warning: This requires opening the value to both parties so that they
```
akirillo, line 294: in order to make this more clear to the caller, should we require that the value be opened outside of the gadget?

akirillo (approved):

PR #265, opened 2023-11-23
circuits: zk-circuits: valid-match-settle: Add settlement tests

joeykraut:
### Purpose
This PR refactors the `VALID SETTLE` unit tests into the `VALID MATCH SETTLE` tests. This PR also removes `valid-settle` from the repo as this circuit is now part of `VALID MATCH SETTLE`.

### Testing
- All unit tests pass


circuits/src/zk_circuits/valid_match_settle/mod.rs
akirillo, line 1: it seems some of the tests check corrupted inputs for both parties, and some only check it for one party. if we're permuting which party gets which order/balance in the test setup, then i think we can only test one party's inputs in each test. otherwise, we should check both partyies' inputs in each test

akirillo (approved):

PR #268, opened 2023-11-24
circuits: mpc-ciruits: settle: Add settlement ciruit

joeykraut:
### Purpose
This PR adds the `settle` circuit to the `mpc-circuits` module. This circuit settles a shared match object into two wallets after the `match` circuit runs. This completes the process of constructing a shared witness and statement for `VALID MATCH SETTLE`.

Note that the settlement indices here are public to provide efficiency. These indices are already made public in the statement type of `VALID COMMITMENTS` so they will simply be pulled from that proof's statement at match time.

### Testing

### DIFF
--- a/circuits/src/mpc_circuits/settle.rs
+++ b/circuits/src/mpc_circuits/settle.rs
@@ -1,0 +1,214 @@
+//! Settles a match into secret shared wallets
+
+use circuit_types::{r#match::AuthenticatedMatchResult, wallet::AuthenticatedWalletShare};
+
+use crate::{
+    mpc_gadgets::comparators::cond_select_vec,
+    zk_circuits::valid_commitments::OrderSettlementIndices,
+};
+
+/// Settles a match into two wallets and returns the updated wallet shares
+///
+/// We settle directly into the public shares both for efficiency and to avoid
+/// the need to share private shares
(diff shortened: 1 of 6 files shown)


circuits/src/mpc_circuits/settle.rs
```diff
@@ -0,0 +1,214 @@
...
+
+    /// Get a dummy set of inputs for a settlement circuit
+    fn generate_test_params() -> SettlementTest {
+        let mut rng = thread_rng();
+        let quote_mint = scalar_to_biguint(&Scalar::random(&mut rng));
+        let base_mint = scalar_to_biguint(&Scalar::random(&mut rng));
+
+        let match_res = MatchResult {
+            quote_mint: quote_mint.clone(),
+            base_mint: base_mint.clone(),
```
akirillo, line 123: are these two also unused? if not, then how does using `random_shares` and `random_indices` work for the test (i would think the indices would need to point to shares that line up with the correct quote/base mints)

akirillo (approved):

circuits/src/mpc_circuits/settle.rs
joeykraut: Nothing about the act of settling guarantees that the mints are correct. This is done elsewhere: concretely in the `VALID COMMITMENTS` circuit. So such indices are guaranteed to be correct for a commitment linked order before they are used. The way the tests use them is misleading, you're right, I'll mark that

PR #270, opened 2023-11-27
arbitrum-client: integration: setup pre-allocated state

akirillo:
This PR introduces the setup of pre-allocated state in the Arbitrum client integration tests, including the necessary helpers to implement this.

It's worth noting that the client in the integration tests will assume that the contracts have had the appropriate verification keys set prior to the test beginning. This will be implemented as another script in the contracts repo which will be invoked within the docker compose stack for the devnet.

At the moment, we use the actual circuitry (on the smaller `(2, 2, 1)` parameter set) in the integration tests for ease of use with existing relayer utilities. If this proves burdensome in terms of proving time in integration tests, or failing proofs (which is out of scope for the client integration tests), I will write dummy circuits that expect the same witnesses but are trivially constrained (and small).

### DIFF
--- a/arbitrum-client/integration/main.rs
+++ b/arbitrum-client/integration/main.rs
@@ -122,7 +122,7 @@ impl From<CliArgs> for IntegrationTestArgs {
 
 /// Sets up pre-allocated state used by the integration tests
 fn setup_pre_allocated_state(client: &ArbitrumClient) -> Result<PreAllocatedState> {
-    // Insert two new wallets into the contract
+    // Insert three new wallets into the contract
     let (index0_commitment, index0_shares) = block_on_result(deploy_new_wallet(client))?;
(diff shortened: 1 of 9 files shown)


arbitrum-client/integration/main.rs
```diff
@@ -82,10 +111,32 @@ impl From<CliArgs> for IntegrationTestArgs {
...
+        Self {
+            client,
+            pre_allocated_state,
+        }
     }
 }
 
+/// Sets up pre-allocated state used by the integration tests
+fn setup_pre_allocated_state(client: &ArbitrumClient) -> Result<PreAllocatedState> {
+    // Insert two new wallets into the contract
```
joeykraut, line 125: nit: 
```suggestion
    // Insert three new wallets into the contract
```

joeykraut (approved):

PR #279, opened 2023-11-29
arbitrum-client: use proof bundle types in contract interaction methods

akirillo:
This PR changes the interfaces of the Arbitrum client's contract interaction methods to accept the proof bundle types created by the relayer. It also adds the necessary conversion logic between relayer and contract statement types to support this, which notably changes the structure of ECDSA public signing keys to store both the `x` and `y` affine coordinates of the (in our case, `secp256k1`) curve point representing the public key.

Existing integration tests pass.

### DIFF
--- a/arbitrum-client/src/client/contract_interaction.rs
+++ b/arbitrum-client/src/client/contract_interaction.rs
@@ -1,7 +1,12 @@
 //! Defines `ArbitrumClient` helpers that allow for interacting with the
 //! darkpool contract
 
-use circuit_types::{merkle::MerkleRoot, wallet::Nullifier, PlonkProof};
+use circuit_types::{merkle::MerkleRoot, wallet::Nullifier};
+use common::types::proof_bundles::{
+    GenericMatchSettleBundle, GenericValidCommitmentsBundle, GenericValidReblindBundle,
+    GenericValidWalletCreateBundle, GenericValidWalletUpdateBundle, ValidCommitmentsBundle,
+    ValidMatchSettleBundle, ValidReblindBundle, ValidWalletCreateBundle, ValidWalletUpdateBundle,
+};
 use constants::Scalar;
 
 use crate::{
@@ -9,8 +14,8 @@ use crate::{
     helpers::{deserialize_calldata, serialize_calldata},
     serde_def_types::SerdeScalarField,
     types::{
-        ContractProof, ContractValidMatchSettleStatement, ContractValidWalletCreateStatement,
-        ContractValidWalletUpdateStatement, MatchPayload,
+        ContractProof, ContractValidWalletCreateStatement, ContractValidWalletUpdateStatement,
+        MatchPayload,
     },
 };
 
@@ -66,8 +71,6 @@ impl ArbitrumClient {
             .map_err(|e| ArbitrumClientError::ContractInteraction(e.to_string()))
     }
 
-    // TODO: Implement analogue of `get_public_blinder_tx`
-
     // -----------
     // | SETTERS |
     // -----------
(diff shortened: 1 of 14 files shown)


arbitrum-client/src/client/contract_interaction.rs
```diff
@@ -76,18 +80,27 @@ impl ArbitrumClient {
...
+        const MAX_ORDERS: usize,
+        const MAX_FEES: usize,
+    >(
         &self,
-        wallet_blinder_share: Scalar,
-        valid_wallet_create_statement: ContractValidWalletCreateStatement,
-        proof: PlonkProof,
-    ) -> Result<(), ArbitrumClientError> {
+        valid_wallet_create: GenericValidWalletCreateBundle<MAX_BALANCES, MAX_ORDERS, MAX_FEES>,
+    ) -> Result<(), ArbitrumClientError>
```
joeykraut, line 90: I'd suggest _not_ making this generic sized. We don't have much need to shrink the size for one (the only reason we do in the `circuits` crate is to make unit tests run faster, and even this is a somewhat old reason. As well, we want to take in a boxed version of the bundle to avoid obliterating the stack.

For that reason I'd use `ValidWalletCreateBundle` [here](https://github.com/renegade-fi/renegade/blob/joey/circuits-refactor/common/src/types/proof_bundles.rs#L35)

arbitrum-client/src/client/contract_interaction.rs
```diff
@@ -107,19 +120,34 @@ impl ArbitrumClient {
     /// `VALID WALLET UPDATE` statement
     ///
     /// Awaits until the transaction is confirmed on-chain
-    pub async fn update_wallet(
+    pub async fn update_wallet<
+        const MAX_BALANCES: usize,
+        const MAX_ORDERS: usize,
+        const MAX_FEES: usize,
+        const MERKLE_HEIGHT: usize,
```
joeykraut, line 127: Same here with the generics and on all interfaces, each bundle type has an associated sized, boxed type

arbitrum-client/src/types.rs
```diff
@@ -165,6 +204,39 @@ pub struct ContractValidWalletUpdateStatement {
...
+impl<const MAX_BALANCES: usize, const MAX_ORDERS: usize, const MAX_FEES: usize>
+    TryFrom<ValidWalletUpdateStatement<MAX_BALANCES, MAX_ORDERS, MAX_FEES>>
+    for ContractValidWalletUpdateStatement
+where
+    [(); MAX_BALANCES + MAX_ORDERS + MAX_FEES]: Sized,
+{
+    type Error = ConversionError;
+
+    fn try_from(
+        value: ValidWalletUpdateStatement<MAX_BALANCES, MAX_ORDERS, MAX_FEES>,
```
joeykraut, line 216: Use the `SizedValidWalletUpdateStatement` here as well

circuit-types/src/keychain.rs
```diff
@@ -26,8 +26,8 @@ use super::{biguint_from_hex_string, biguint_to_hex_string};
 pub const NUM_KEYS: usize = 4;
 /// The number of bytes used in a single scalar to represent a key
 pub const SCALAR_MAX_BYTES: usize = 31;
-/// The number of words needed to represent a non-native root key
-pub const ROOT_KEY_WORDS: usize = 2;
+/// The number of words needed to represent a non-native root key coordinate
```
joeykraut, line 29: Can we expand this doc comment to indicate that we are representing a curve point under the hood? "coordinate" alone may not make sense without context

circuit-types/src/keychain.rs
```diff
@@ -185,14 +185,14 @@ impl<'de, const KEY_WORDS: usize> Deserialize<'de> for NonNativeKey<KEY_WORDS> {
...
-    fn from(val: &NonNativeKey<KEY_WORDS>) -> Self {
+impl<const SCALAR_WORDS: usize> From<&NonNativeScalar<SCALAR_WORDS>> for DalekKey {
+    fn from(val: &NonNativeScalar<SCALAR_WORDS>) -> Self {
         let key_bytes = BigUint::from(val).to_bytes_le();
         DalekKey::from_bytes(&key_bytes).unwrap()
     }
 }
 
-impl<const KEY_WORDS: usize> From<DalekKey> for NonNativeKey<KEY_WORDS> {
+impl<const SCALAR_WORDS: usize> From<DalekKey> for NonNativeScalar<SCALAR_WORDS> {
```
joeykraut, line 195: This can be addressed later, but I believe we will want to remove this `DalekKey` use in favor of an secp256 key, right?

joeykraut (approved):
Just a few small suggestions: I think broadly `arbitrum-client` should not expose the sizing constraints of the circuit types, nor should really any of the relayer for that matter

circuit-types/src/keychain.rs
akirillo: ack, yes, but also think it should be addressed later - this was just me find-and-replacing

PR #282, opened 2023-11-29
workers: Persist timestamps in global orderbook

sehyunc:
This PR persists timestamps of NetworkOrder objects so that when the global orderbook is fetched, timestamps accurately represent when the order was received.

### DIFF
--- a/common/src/types/network_order.rs
+++ b/common/src/types/network_order.rs
@@ -5,6 +5,7 @@ use std::fmt::{Display, Formatter, Result as FmtResult};
 
 use circuit_types::wallet::Nullifier;
 use serde::{Deserialize, Serialize};
+use util::get_current_time_seconds;
 
 use super::{
     gossip::ClusterId,
@@ -66,6 +67,8 @@ pub struct NetworkOrder {
     /// have `None` in place
     #[serde(skip)]
     pub validity_proof_witnesses: Option<OrderValidityWitnessBundle>,
+    /// The timestamp this order was received at
+    pub timestamp: u64,
 }
 
 impl NetworkOrder {
@@ -84,6 +87,7 @@ impl NetworkOrder {
             state: NetworkOrderState::Received,
             validity_proofs: None,
             validity_proof_witnesses: None,
+            timestamp: get_current_time_seconds(),
         }
     }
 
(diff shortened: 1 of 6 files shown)


joeykraut (commented):
I'd suggest a slightly different approach: specifically we don't intervene at the API layer, but just store the data correctly.

The changes would be something like:
- Adding a `timestamp` field to the [`NetworkOrder`](https://github.com/renegade-fi/renegade/blob/main/common/src/types/network_order.rs#L44) as defined in the state.
- Modifying its constructor to set this timestamp to `util::get_current_time_seconds`
- We should probably also overwrite this field in order book's [`add_order`](https://github.com/renegade-fi/renegade/blob/main/state/src/orderbook.rs#L289) method to be the current time.
- In the `Into` method you modified in this PR, we can just actually use the timestamp from the original `NetworkOrder`

sehyunc:
> I'd suggest a slightly different approach: specifically we don't intervene at the API layer, but just store the data correctly.
> 
> The changes would be something like:
> 
> * Adding a `timestamp` field to the [`NetworkOrder`](https://github.com/renegade-fi/renegade/blob/main/common/src/types/network_order.rs#L44) as defined in the state.
> * Modifying its constructor to set this timestamp to `util::get_current_time_seconds`
> * We should probably also overwrite this field in order book's [`add_order`](https://github.com/renegade-fi/renegade/blob/main/state/src/orderbook.rs#L289) method to be the current time.
> * In the `Into` method you modified in this PR, we can just actually use the timestamp from the original `NetworkOrder`

Much cleaner solution, thanks for the breakdown. I've tested that this branch works well with the frontend, timestamps fetched from the global orders route are correctly persisted.

common/src/types/network_order.rs
```diff
@@ -66,6 +67,8 @@ pub struct NetworkOrder {
     /// have `None` in place
     #[serde(skip)]
     pub validity_proof_witnesses: Option<OrderValidityWitnessBundle>,
+    /// The timestamp this order was placed at
```
joeykraut, line 70: nit: It'd be more accurate to say:
```suggestion
    /// The timestamp this order was received at
```

joeykraut (approved):

PR #289, opened 2023-11-30
task-driver: settle-match-internal: Refactor task over `mpc-plonk`

joeykraut:
### Purpose
This PR refactors the `settle-match-internal` task over the `mpc-plonk` system. This task handles settling a match generated by the internal engine: i.e. on two orders managed by the local node. 

The changes here do some work to simplify the task implementation to use helpers in common with the `settle-match` task.

### Testing
- Unit tests pass


task-driver/src/settle_match_internal.rs
```diff
@@ -336,193 +266,53 @@ impl SettleMatchInternalTask {
...
-        self.global_state.update_wallet(buy_side_wallet).await;
-        self.global_state.update_wallet(sell_side_wallet).await;
+        // Reblind both wallets and update their merkle openings
+        wallet1.reblind_wallet();
+        wallet2.reblind_wallet();
 
-        Ok(())
-    }
+        self.find_opening(&mut wallet1).await?;
+        self.find_opening(&mut wallet2).await?;
```
akirillo, line 310: should we be doing this before reblinding?

task-driver/src/settle_match_internal.rs
```diff
@@ -336,193 +266,53 @@ impl SettleMatchInternalTask {
...
-        let tx_submit_res = self
-            .starknet_client
-            .submit_match(
-                party0_reblind_proof.original_shares_nullifier,
-                party1_reblind_proof.original_shares_nullifier,
-                party0_reblind_proof.reblinded_private_share_commitment,
-                party1_reblind_proof.reblinded_private_share_commitment,
-                valid_settle_proof.statement.party0_modified_shares.clone(),
-                valid_settle_proof.statement.party1_modified_shares.clone(),
+        let match_settle_proof = self.proof_bundle.clone().unwrap();
```
akirillo, line 277: can we `take` here to avoid a `clone`, or will we need the proof bundle in subsequent steps?

akirillo (approved):

PR #292, opened 2023-12-01
workers: api-server: Refactor over new `mpc-plonk` system

joeykraut:
### Purpose
This PR refactors the `api-server` over the new changes introduced with the `mpc-plonk` refactor. Among other things this includes changing the API authentication scheme from eddsa over curve 25519 to ecdsa over secp256k. Included in this are some conversion methods for wallet key types to the `k256` key types.

### Testing
- Unit tests pass, tested signature verification


Cargo.toml
```diff
@@ -29,6 +29,9 @@ members = [
 ]
 exclude = ["mpc-stark", "mpc-bulletproof"]
 
+[profile.test]
+debug-assertions = false
```
joeykraut, line 33: This is necessary to prevent `hyper` from validating our headers in the `test` profile. We need to attach raw signatures in unit tests

akirillo (approved):

PR #293, opened 2023-12-01
external-api: Add update_locked field to Wallet type to expose field to API

sehyunc:
This PR adds the `update_locked` field to the Wallet type so that the field is present when the type conversion happens within `GetWalletHandler`. For use in the frontend to disable functionality when the wallet is undergoing any wallet update operations.

### DIFF
--- a/external-api/src/types.rs
+++ b/external-api/src/types.rs
@@ -3,7 +3,10 @@
 use std::{
     collections::HashMap,
     convert::TryInto,
-    sync::{atomic::AtomicBool, Arc},
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        Arc,
+    },
 };
 
 use circuit_types::{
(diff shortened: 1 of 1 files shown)


external-api/src/types.rs
```diff
@@ -105,6 +111,7 @@ impl From<IndexedWallet> for Wallet {
             blinded_public_shares,
             private_shares,
             blinder: scalar_to_biguint(&wallet.blinder),
+            update_locked: wallet.update_locked.load(Ordering::Relaxed),
```
sehyunc, line 114: @joeykraut Let me know if I should do `update_locked: wallet.is_locked()` here. Purposefully didn't because of build errors saying `wallet.fees` partially moved due to this method call on line 89.

joeykraut: Yeah probably best to use `wallet.is_locked`. Just do `let update_locked = wallet.is_locked();` above the line where the move happens

joeykraut (approved):

PR #297, opened 2023-12-02
arbitrum-client: clear & re-initialize preallocated state

akirillo:
This PR changes the Arbitrum client integration tests to clear and re-allocate Merkle state at the beginning/end of all relevant tests. This allows the tests to run in any order, without relying on e.g. the Merkle paths remaining unchanged.

### DIFF
--- a/arbitrum-client/src/abi.rs
+++ b/arbitrum-client/src/abi.rs
@@ -5,6 +5,7 @@
 use alloy_sol_types::sol;
 use ethers::contract::abigen;
 
+#[cfg(not(feature = "integration"))]
 abigen!(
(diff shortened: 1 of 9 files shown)


arbitrum-client/src/abi.rs
```diff
@@ -22,6 +23,26 @@ abigen!(
     ]"#
 );
 
+#[cfg(feature = "integration")]
```
joeykraut, line 26: nit: Can we add a comment here delineating the difference in the generated ABI, it's simpler than trying to compare them manually if I'm reading this for the first time

joeykraut (approved):

PR #299, opened 2023-12-02
core, config: Refactor over new `arbitrum` + `mpc-plonk` system

joeykraut:
### Purpose
This PR refactors the `core` and `config` crates over the new `mpc-plonk` based system.

### Testing


core/src/main.rs
```diff
@@ -122,33 +119,21 @@ async fn main() -> Result<(), CoordinatorError> {
...
+    .map_err(|e| CoordinatorError::Arbitrum(e.to_string()))?;
 
     // Build a task driver that may be used to spawn long-lived asynchronous tasks
     // that are common among workers
     let task_driver_config = TaskDriverConfig::default_with_bus(system_bus.clone());
     let task_driver = TaskDriver::new(task_driver_config);
 
-    // Spawn a thread to sync the relayer-global state with on-chain state and
-    // network state
-    let task = InitializeStateTask::new(
```
akirillo, line 143: do we no longer need this? and if not, should we delete the struct definition as well?

akirillo (approved):
unsure about removal of `InitializeStateTask`, otherwise lgtm

core/src/main.rs
joeykraut: correct this was used for passing in wallets directly via the config, we now just use the API. I removed the task definition in #285

PR #306, opened 2023-12-09
external-api: (De)serialize all keys from/to hex strings

joeykraut:
### Purpose
This PR changes all key types to be serialized and deserialized to/from hex strings at the API layer. 

### Testing
- Tested with `POST /v0/wallet/`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2834,12 +2834,14 @@ version = "0.1.0"
 dependencies = [
  "circuit-types",
  "common",
+ "constants",
+ "hex 0.4.3",
  "itertools 0.10.5",
  "num-bigint",
- "num-traits",
  "renegade-crypto",
  "serde",
  "serde_json",
+ "util",
  "uuid 1.6.1",
 ]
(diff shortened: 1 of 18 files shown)


sehyunc (approved):
LGTM, small point on consistency, there's [another definition](https://github.com/renegade-fi/renegade/blob/4b93caff1ec28ae694ad0a73a4695b33c549226e/task-driver/integration/helpers.rs#L36) of helper functions for conversion between biguint <> hex string, should these all just use this new `hex.rs` file?

sehyunc:
Also I'll go ahead and test this locally with the SDK

joeykraut:
> LGTM, small point on consistency, there's [another definition](https://github.com/renegade-fi/renegade/blob/4b93caff1ec28ae694ad0a73a4695b33c549226e/task-driver/integration/helpers.rs#L36) of helper functions for conversion between biguint <> hex string, should these all just use this new `hex.rs` file?

Great catch, will do

PR #307, opened 2023-12-11
arbitrum-client, docker: adapt to new contract interface

akirillo:
This PR updates the Arbitrum client to adapt to the new contract interface, in which, namely, there are no longer any setter methods, and events/external methods expect/return `uint256` types where `Scalar`s are used, as opposed to `bytes`. This also makes event parsing more straightforward, in that we no longer need to generate a hash of the serialized `Scalar` to listen for events containing it.

`arbitrum-client` integration tests pass.

### DIFF
--- a/docker/contracts/deploy_contracts.sh
+++ b/docker/contracts/deploy_contracts.sh
@@ -30,97 +30,50 @@ no_verify() {
 
 # Deploy verifier contract
 cargo run \
-    -p scripts -- \
-    -p $DEVNET_PKEY \
-    -r $DEVNET_RPC_URL \
-    -d $DEPLOYMENTS_PATH \
+    --package scripts -- \
+    --priv-key $DEVNET_PKEY \
+    --rpc-url $DEVNET_RPC_URL \
+    --deployments-path $DEPLOYMENTS_PATH \
     deploy-stylus \
     --contract verifier
 
 # Deploy Merkle contract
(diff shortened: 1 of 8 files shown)


docker/contracts/Dockerfile
```diff
@@ -25,6 +25,7 @@ RUN cargo install wasm-opt --locked
 # Clone the contracts repo
 WORKDIR /sources
 RUN git clone \
+    -b andrew/srs-file \
```
akirillo, line 28: Note: this will be removed when https://github.com/renegade-fi/renegade-contracts/pull/153 is merged

joeykraut: approve that PR, bumping this so it doesn't get overlooked before merge

docker/contracts/deploy_contracts.sh
```diff
@@ -64,7 +64,8 @@ cargo run \
     -r $DEVNET_RPC_URL \
     -d $DEPLOYMENTS_PATH \
     deploy-proxy \
-    -o $DEVNET_ACCOUNT_ADDRESS
+    -o $DEVNET_ACCOUNT_ADDRESS \
+    -t
```
joeykraut, line 68: is this `--test`? Flags should use the full name for readability

joeykraut (approved):

PR #311, opened 2023-12-21
task-driver: integration: Fix integration tests & bugs

joeykraut:
### Purpose
This PR gets the `task-driver` integration tests into a passing state. This involves fixing the tests and a few bugs in the task implementations directly.

### Testing
- Unit tests pass
- `task-driver` integration tests pass


task-driver/integration/helpers.rs
```diff
@@ -94,21 +99,38 @@ pub async fn new_wallet_in_darkpool(client: &ArbitrumClient) -> Result<(Wallet,
...
+    client: &ArbitrumClient,
+) -> Result<()> {
     let share_comm = wallet.get_private_share_commitment();
 
     let mut proof = dummy_valid_wallet_create_bundle();
     proof.statement.public_wallet_shares = wallet.blinded_public_shares.clone();
     proof.statement.private_shares_commitment = share_comm;
 
-    client.new_wallet(proof).await.map_err(Into::into)
+    client.new_wallet(proof).await.map_err(Into::<Report>::into)?;
```
akirillo, line 120: nit: I think you can just do `client.new_wallet(proof).await?` and any type implementing `Error` will automatically be turned `Into` a `Report`

but if you tried this already and it didn't work then nvm, must be misremembering

task-driver/integration/helpers.rs
```diff
@@ -125,36 +147,110 @@ pub async fn mock_wallet_update(wallet: &mut Wallet, client: &ArbitrumClient) ->
...
+    if funds_needed > 0 {
+        fund_wallet_with_erc20(funds_needed, test_args).await?;
+    }
+
+    Ok(())
+}
+
+/// Wrap a given amount of native ETH in WETH
+pub(crate) async fn wrap_eth(amount: u64, test_args: &IntegrationTestArgs) -> Result<()> {
+    let weth_client =
```
akirillo, line 192: nit: It seems we only use this method in a context where a `weth_client` already exists, we could just accept a reference to it as an argument

same for `fund_wallet_with_erc20` below

task-driver/integration/helpers.rs
```diff
@@ -125,36 +147,110 @@ pub async fn mock_wallet_update(wallet: &mut Wallet, client: &ArbitrumClient) ->
...
+        .balance_of(addr)
+        .await
+        .map_err(|e| eyre!("error getting erc20 balance: {e}"))?;
+
+    let funds_needed = FUNDING_AMOUNT.saturating_sub(erc20_balance.as_u64());
+    if funds_needed > 0 {
+        fund_wallet_with_erc20(funds_needed, test_args).await?;
+    }
+
+    Ok(())
```
akirillo, line 187: thoughts on also approving the darkpool to spend the wallet's funds here before returning? seems like a sensible place to do it

task-driver/integration/tests/settle_match.rs
```diff
@@ -171,9 +174,18 @@ async fn setup_match_result(
         min_amount_order_index: false,
     };
 
-    // Simulate a reblind that would happen before a match
-    wallet1.reblind_wallet();
-    wallet2.reblind_wallet();
```
akirillo, line 176: confirming we can remove this b/c reblinding occurs in `setup_wallet_with_order_balance`, right?

task-driver/integration/tests/update_wallet.rs
```diff
@@ -52,11 +53,17 @@ lazy_static! {
...
 pub(crate) async fn execute_wallet_update(
-    old_wallet: Wallet,
+    mut old_wallet: Wallet,
     new_wallet: Wallet,
     transfer: Option<ExternalTransfer>,
     test_args: IntegrationTestArgs,
-) -> Result<()> {
+) -> Result<Wallet> {
+    // Make sure the Merkle proof is present
+    if old_wallet.merkle_proof.is_none() {
```
akirillo, line 62: it feels like `old_wallet.merkle_proof` should never be `None`, this would just mean the wallet wasn't properly allocated in the darkpool, right?

akirillo (approved):
couple qs/nits, lgtm tho

task-driver/integration/tests/update_wallet.rs
joeykraut: Not really, this is just a convenient safety check on the tests. The tests are more brittle the more state we expect them to compose.

task-driver/integration/helpers.rs
joeykraut: I'd prefer a simpler interface that matches the rest of the helpers (doesn't require any caller setup). Much nicer to work with in tests

PR #314, opened 2024-01-05
proof-manager: mock: Check constraint satisfaction in mock

joeykraut:
### Purpose
This PR changes the behavior of the mock proof manager to check constraint satisfaction of the witness and statement. This allows us to check the validity of witness, statement pairs generated by the tasks.

As a result, this PR fixes a few bugs in the task driver integration tests.

### Testing
- Unit tests pass
- Integration tests mostly pass, some error for nitro specific reasons which will be fixed when time permits.

### DIFF
--- a/workers/proof-manager/src/mock.rs
+++ b/workers/proof-manager/src/mock.rs
@@ -1,12 +1,22 @@
 //! Defines a mock for the proof manager that doesn't prove statements, but
 //! instead immediately returns dummy proofs that will not verify
 
+use circuit_types::traits::SingleProverCircuit;
 use circuits::zk_circuits::{
-    valid_commitments::{SizedValidCommitmentsWitness, ValidCommitmentsStatement},
-    valid_match_settle::{SizedValidMatchSettleStatement, SizedValidMatchSettleWitness},
-    valid_reblind::{SizedValidReblindWitness, ValidReblindStatement},
-    valid_wallet_create::{SizedValidWalletCreateStatement, SizedValidWalletCreateWitness},
-    valid_wallet_update::{SizedValidWalletUpdateStatement, SizedValidWalletUpdateWitness},
+    check_constraint_satisfaction,
+    valid_commitments::{
+        SizedValidCommitments, SizedValidCommitmentsWitness, ValidCommitmentsStatement,
+    },
+    valid_match_settle::{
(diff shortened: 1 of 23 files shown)


workers/proof-manager/src/mock.rs
```diff
@@ -31,134 +43,121 @@ pub struct MockProofManager;
...
+        Self::check_constraints::<SizedValidMatchSettle>(&witness, &statement)?;
+
         let proof = dummy_proof();
-        Box::new(GenericMatchSettleBundle { statement, proof })
+        Ok(Box::new(GenericMatchSettleBundle { statement, proof }))
     }
-}
-
-#[cfg(test)]
-mod test {
```
joeykraut, line 120: we don't need to test the mock, and it's out of scope now that it requires a valid witness and statement

workers/proof-manager/src/mock.rs
```diff
@@ -31,134 +43,121 @@ pub struct MockProofManager;
...
+        });
     }
 
     /// The execution loop for the mock
-    fn execution_loop(job_queue: &Receiver<ProofManagerJob>) {
+    fn execution_loop(job_queue: &Receiver<ProofManagerJob>) -> Result<(), ProofManagerError> {
         loop {
             match job_queue.recv() {
                 Err(e) => {
                     log::error!("error receiving job: {e}... shutting down");
```
akirillo, line 58: should we still log this? seems like the error will necessarily get logged above in `start`. and if we do wanna log here, should we still say "shutting down"?

workers/proof-manager/src/mock.rs
```diff
@@ -31,134 +43,121 @@ pub struct MockProofManager;
...
         };
 
         response_channel.send(bundle).expect(ERR_RESPONSE_CHANNEL_CLOSED);
+        Ok(())
     }
 
     /// Generate a dummy proof of `VALID WALLET CREATE`
     fn valid_wallet_create(
-        _witness: &SizedValidWalletCreateWitness,
+        witness: SizedValidWalletCreateWitness,
```
akirillo, line 95: clippy complaining about pass by value here, and for the other mock proof generation methods

akirillo (approved):
just a couple nits, lgtm

PR #315, opened 2024-01-07
arbitrum-client: import types directly from contracts repo

akirillo:
This PR has the `arbitrum-client` crate import the smart contract analogues to the proof and statement types directly from the `contracts-common` crate of the contracts repo, instead of redefining them. This is the actual source of truth for the types expected by the contracts, and makes integration testing in the contracts repo significantly more straightforward.

**Testing:**
`arbitrum-client` integration tests pass.


arbitrum-client/src/conversion.rs
```diff
@@ -0,0 +1,203 @@
...
+}
+
+/// Convert a [`SizedValidWalletUpdateStatement`] to its corresponding smart
+/// contract type
+pub fn to_contract_valid_wallet_update_statement(
+    statement: SizedValidWalletUpdateStatement,
+) -> Result<ContractValidWalletUpdateStatement, ConversionError> {
+    let new_public_shares = wallet_shares_to_scalar_vec(&statement.new_public_shares);
+    let external_transfer: Option<ContractExternalTransfer> =
+        if statement.external_transfer == ExternalTransfer::default() {
```
joeykraut, line 107: I'd suggest adding an `is_null` or otherwise to the `ExternalTransfer` type, like what we do for orders, [balances](https://github.com/renegade-fi/renegade/blob/dfc4d1c6ff1f193da2baa9992747081cf690e3b5/circuit-types/src/balance.rs#L38-L40), and fees. It is somewhat nicer to work with, I'd imagine this functionality pops up elsewhere

joeykraut (commented):
Nice, glad we could remove all those extra types

PR #319, opened 2024-01-09
circuit-macros: proof-linking: Implement linking witness allocation

joeykraut:
### Purpose
This PR implements the single-prover variant of the macro derived link group witness allocation. This works by overriding the `create_witness` method on the `CircuitBaseType` impl for types that need proof linking. The macro constructs a list of linking groups needed for each witness element and constructs a `create_witness` implementation that calls `cs.create_variable_with_link_groups(var, &groups)` for each witness element.

The syntax for specifying a witness element with link groups is as follows:
```rust
#[circuit-type(singleprover)]
struct TestWitness {
    pub a: u64, // no link needed
    #[link_groups = "group1, group2, ..."]
    pub b: bool
}
```
For which the macro will expand to `cs.create_variable_with_link_groups(b, &["group1".to_string(), "group2".to_string()])`

### Todo
- Implementation for `MultiproverCircuitBaseType`

### Testing
- All unit tests pass
- Added link groups to the test structure in `macro_tests.rs` and verified that the circuit layout contained these values correctly.

### DIFF
--- a/circuit-macros/src/circuit_type/proof_linking.rs
+++ b/circuit-macros/src/circuit_type/proof_linking.rs
@@ -1,0 +1,163 @@
+//! Helpers for generating proof linking helper methods, i.e. implementations of
+//! variable allocation that use proof linking groups
+//!
+//! Linking groups are specified by annotating fields in a struct with the
+//! `link_groups` attribute. For example, the following struct will
+//! allocate `field1` into groups `group1` and `group2` and `field2` into group
+//! `group3`:
+//!
+//! ```
+//! #[circuit-type(singleprover)]
+//! struct MyCircuit {
+//!    #[link_groups = "group1, group2")]
+//!    field1: Scalar,
+//!    #[link_groups = "group3"]
+//!    field2: Scalar,
+//! }
+
+use proc_macro2::{Literal, TokenStream, TokenTree};
+use quote::quote;
+use syn::{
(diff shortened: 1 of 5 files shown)


akirillo (approved):

circuit-macros/src/circuit_type/proof_linking.rs
```diff
@@ -0,0 +1,163 @@
...
+    match next {
+        TokenTree::Punct(punct) => {
+            assert_eq!(punct.as_char(), '=', "expected '='");
+        },
+        _ => panic!("expected '='"),
+    }
+
+    // Pull a string literal and parse groups from this directly
+    let groups = match token_iter.next().expect("expected string literal") {
+        TokenTree::Literal(lit) => parse_linking_groups_from_str(&lit),
```
akirillo, line 146: If we expect the attribute to contain a string literal, this won't exactly play well with using constantized link group names as in the last PR, will it?

Probably not worth the complexity to support constants, but just calling it out

akirillo (approved):

circuit-macros/src/circuit_type/proof_linking.rs
joeykraut: Correct, that's not the intent, you register a link group with the circuit and refer to it in the witness definition

PR #322, opened 2024-01-10
circuits: zk-circuits: proof-linking: Implement linking helpers

joeykraut:
### Purpose
This PR adds linking helpers for computing proof links between our application level circuits. Testing is deferred to a follow up PR

### Todo
- Add unit tests

### Testing
- Unit tests pass

### DIFF
--- a/circuits/src/zk_circuits/proof_linking.rs
+++ b/circuits/src/zk_circuits/proof_linking.rs
@@ -1,0 +1,258 @@
+//! Helpers for linking proofs between circuits
+//!
+//! TODO: Figure out which proving and verifying keys to use once we generate
+//! them for the circuits
+
+use core::panic;
+
+use ark_mpc::{network::PartyId, PARTY0, PARTY1};
+use circuit_types::{
+    errors::ProverError, traits::SingleProverCircuit, Fabric, MpcPlonkLinkProof,
+    MpcProofLinkingHint, PlonkLinkProof, PlonkProof, ProofLinkingHint,
+};
+use constants::{MAX_BALANCES, MAX_FEES, MAX_ORDERS, MERKLE_HEIGHT};
+use mpc_plonk::{
+    multiprover::proof_system::MultiproverPlonkKzgSnark, proof_system::PlonkKzgSnark,
+    transcript::SolidityTranscript,
+};
+use mpc_relation::proof_linking::GroupLayout;
+
+use crate::zk_circuits::{valid_reblind::ValidReblind, VALID_REBLIND_COMMITMENTS_LINK};
+
+use super::{
+    valid_match_settle::ValidMatchSettle, VALID_COMMITMENTS_MATCH_SETTLE_LINK0,
+    VALID_COMMITMENTS_MATCH_SETTLE_LINK1,
+};
+
+// ---------------------------------------
+// | Valid Reblind <-> Valid Commitments |
+// ---------------------------------------
+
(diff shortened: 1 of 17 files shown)


circuits/src/zk_circuits/valid_commitments.rs
```diff
@@ -411,10 +413,26 @@ pub mod test_helpers {
...
-
         // Split the wallet into secret shares
-        let (private_secret_shares, public_secret_shares) = create_wallet_shares(wallet);
+        let (private_share, public_share) = create_wallet_shares(wallet);
+        create_witness_and_statement_with_shares(wallet, &public_share, &private_share)
+    }
+
+    /// Create a witness and statement with wallet shares specified alongside
+    /// the wallet
+    pub fn create_witness_and_statement_with_shares<
```
joeykraut, line 423: this was done for a testing PR yet to come

circuits/src/zk_circuits/valid_commitments.rs
```diff
@@ -341,7 +341,7 @@ where
     type Witness = ValidCommitmentsWitness<MAX_BALANCES, MAX_ORDERS, MAX_FEES>;
 
     fn name() -> String {
-        "Valid Commitments".to_string()
+        format!("Valid Commitments ({MAX_BALANCES}, {MAX_ORDERS}, {MAX_FEES})")
```
joeykraut, line 344: i changed the names to include sizing constants so that the layout cache has different entries for different sizes -- this is necessary during tests

circuits/src/zk_circuits/proof_linking.rs
```diff
@@ -0,0 +1,218 @@
...
+    link_commitments_match_settle::<MAX_BALANCES, MAX_ORDERS, MAX_FEES>(
+        party_id,
+        commitments_link_hint,
+        match_settle_link_hint,
+        fabric,
+    )
+}
+
+/// Link a proof of MATCH SETTLE with a proof of VALID COMMITMENTS
+pub fn link_commitments_match_settle<
```
akirillo, line 139: Should we make an analogue of this method for the singleprover (internal match) setting? We'd just use `ProofLinkingHint`s and no `Fabric` necessary right

circuits/src/zk_circuits/proof_linking.rs
```diff
@@ -0,0 +1,218 @@
+//! Helpers for linking proofs between circuits
+//!
+//! TODO: Figure out which proving and verifying keys to use once we generate
+//! them for the circuits
```
akirillo, line 4: FWIW, I think it's fair and safe to assume / assert that all circuits use the same SRS, of sufficient degree for each. We could cache this SRS (similar to `TESTING_SRS`) and extract the KZG commit / open keys directly from it.

Would make the review of these helpers simpler, wouldn't have to think about "is this the right commit / open key"

In fact, this assumption is already implicit, because we use wiring polynomial commitments from different circuits to construct the linking identity polynomial commitment

akirillo (approved):
lgtm, just a couple qs

circuits/src/zk_circuits/proof_linking.rs
joeykraut: it is a matter of length, the proving key for one circuit is a trimmed version of the common SRS. I'll leave the todo for now, but agree it is likely trivially obviated.

circuits/src/zk_circuits/proof_linking.rs
joeykraut: good call out, will do

PR #323, opened 2024-01-10
circuit-types: Impl for `usize` and use in `OrderSettlementIndices`

joeykraut:
### Purpose
This PR changes the type from `u64` -> `usize` for settlement indices in circuit types. This makes out-of-circuit indexing for e.g. witness generation much easier. Concretely this involves:
- Define a base type impl for `usize`
- Replace all instances of u64 with usize

The one exception is in `arbitrum-client`. I keep the types `u64` here to ensure correct serialization independent of machine architecture. @akirillo Please let me know if this is unnecessary.

### Testing
- Unit tests pass workspace-wide

### DIFF
--- a/circuit-types/src/traits.rs
+++ b/circuit-types/src/traits.rs
@@ -359,6 +359,18 @@ impl BaseType for u64 {
     }
 }
 
+impl BaseType for usize {
+    const NUM_SCALARS: usize = 1;
+
+    fn to_scalars(&self) -> Vec<Scalar> {
+        vec![Scalar::from(*self as u64)]
+    }
+
+    fn from_scalars<I: Iterator<Item = Scalar>>(i: &mut I) -> Self {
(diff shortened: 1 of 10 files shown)


circuit-types/src/traits.rs
```diff
@@ -359,6 +359,18 @@ impl BaseType for u64 {
     }
 }
 
+impl BaseType for usize {
```
akirillo, line 362: do we still need to `impl BaseType for u64`?

akirillo (approved):
Yeah, let's leave as `u64` in case this is an issue for serialization, o/w lgtm - but there is a clippy lint failure for an unnecessary `usize -> usize` conversion

circuit-types/src/traits.rs
joeykraut: yeah a few other fields in witness types make more sense as a u64 (numeric) vs usize (index)

PR #324, opened 2024-01-11
circuits: zk-circuit: proof-linking: Test linking helpers

joeykraut:
### Purpose
This PR adds testing for the proof linking helpers defined in #322. These test a variety of correct and incorrect linking cases.

### Testing
- All unit tests pass

### DIFF
--- a/circuits/src/zk_circuits/proof_linking.rs
+++ b/circuits/src/zk_circuits/proof_linking.rs
@@ -259,3 +259,476 @@ where
         .map_err(ProverError::Plonk)?;
     Ok(layout.get_group_layout(group_id))
 }
+
+#[cfg(all(test, feature = "large_tests"))]
+mod test {
+    use ark_mpc::{network::PartyId, test_helpers::execute_mock_mpc, PARTY0, PARTY1};
+    use circuit_types::{
+        balance::Balance,
+        errors::ProverError,
+        order::OrderSide,
+        traits::{BaseType, MpcBaseType, SingleProverCircuit},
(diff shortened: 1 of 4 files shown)


circuits/src/zk_circuits/proof_linking.rs
```diff
@@ -259,3 +259,476 @@ where
...
+        let mut result_shares = match_settle_statement.party0_modified_shares.to_scalars();
+        let modification_idx = (0..modified_shares.len()).sample_single(&mut rng);
+
+        let modification = Scalar::random(&mut rng);
+        modified_shares[modification_idx] += modification;
+        result_shares[modification_idx] += modification;
+        match_settle_witness.party0_public_shares =
+            WalletShare::from_scalars(&mut modified_shares.into_iter());
+        match_settle_statement.party0_modified_shares =
+            WalletShare::from_scalars(&mut result_shares.into_iter());
```
akirillo, line 666: nit: wanna break this out into a helper, seeing as though we use it across multiple tests?

akirillo (approved):
just one small nit, looking very clean

circuits/src/zk_circuits/proof_linking.rs
joeykraut: It's only done twice, and is used differently in both cases (add to both in one, add/subtract in the other), so I don't think a helper is going to do much here and would need an odd behavioral switch. I'll leave it for now, but will add in the future if this becomes more repetitive

PR #330, opened 2024-01-12
task-driver: helpers: Prove reblind <-> commitments link

joeykraut:
### Purpose
This PR takes the steps to prove the `VALID REBLIND` <-> `VALID COMMITMENTS` link in the task driver's helper to update a wallet's validity proofs.

### Testing
- Unit tests pass crate wide, next step is to validate this in gossip

### DIFF
--- a/task-driver/src/helpers.rs
+++ b/task-driver/src/helpers.rs
@@ -12,9 +12,10 @@ use circuit_types::{
     },
     order::{Order, OrderSide},
     r#match::OrderSettlementIndices,
-    SizedWallet,
+    ProofLinkingHint, SizedWallet,
 };
 use circuits::zk_circuits::{
+    proof_linking::link_sized_commitments_reblind,
     valid_commitments::{
         SizedValidCommitmentsWitness, ValidCommitmentsStatement, ValidCommitmentsWitness,
     },
@@ -25,10 +26,7 @@ use common::types::{
     wallet::{Wallet, WalletAuthenticationPath},
 };
 use common::types::{
-    proof_bundles::{
-        OrderValidityProofBundle, OrderValidityWitnessBundle, ValidCommitmentsBundle,
-        ValidReblindBundle,
-    },
+    proof_bundles::{OrderValidityProofBundle, OrderValidityWitnessBundle},
     wallet::OrderIdentifier,
 };
 use crossbeam::channel::Sender as CrossbeamSender;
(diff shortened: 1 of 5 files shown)


task-driver/src/helpers.rs
```diff
@@ -337,44 +335,39 @@ pub(crate) async fn update_wallet_validity_proofs(
...
-        .read_order_book()
-        .await
-        .attach_validity_proof_witness(order_id, witness_bundle)
-        .await;
-}
-
-/// Attach a validity proof to the locally managed state
+/// Attach a validity proof and witness to the locally managed state
 ///
 /// This is gossipped to the network so that peers may verify the validity
```
akirillo, line 340: nit: can we be explicit that the proof (not the witness) is gossiped to the network

task-driver/src/helpers.rs
```diff
@@ -337,44 +335,39 @@ pub(crate) async fn update_wallet_validity_proofs(
...
+    reblind_witness: &SizedValidReblindWitness,
+    commitments_bundle: &ProofBundle,
+    reblind_bundle: &ProofBundle,
     global_state: &RelayerState,
     network_sender: &TokioSender<GossipOutbound>,
 ) -> Result<(), String> {
+    // Prove the link between the reblind and commitments proofs
+    let reblind_link_hint = &reblind_bundle.link_hint;
+    let comms_link_hint = &commitments_bundle.link_hint;
+    let linking_proof = link_sized_commitments_reblind(reblind_link_hint, comms_link_hint)
```
akirillo, line 354: perhaps obvious, but want to ask explicitly: computing link proofs is definitely lightweight enough that we don't need it to be separately handled in the proof manager?

task-driver/src/helpers.rs
```diff
@@ -337,44 +335,39 @@ pub(crate) async fn update_wallet_validity_proofs(
...
-        .await;
-}
-
-/// Attach a validity proof to the locally managed state
+/// Attach a validity proof and witness to the locally managed state
 ///
 /// This is gossipped to the network so that peers may verify the validity
 /// bundle and schedule matches on the proven order
-pub(crate) async fn record_validity_proof(
+async fn record_validity_proof_and_witness(
```
akirillo, line 342: nit: can we change the name (and certainly doc comment) of this method to indicate that it also computes a linking proof? for the name, maybe something like `record_linked_validity_proof_and_witness`, or `link_and_record_validity_bundle`

akirillo (approved):
couple nits but lgtm

task-driver/src/helpers.rs
joeykraut: Correct

PR #332, opened 2024-01-16
arbitrum-client: integration: use new sequencer service

akirillo:
This PR updates the `arbitrum-client` integration testing stack to use the new `sequencer` image with predeployed contracts.

**Testing:**
All `arbitrum-client` integration tests other than `test_process_match_settle` pass. This is expected, since the Arbitrum client hasn't yet been updated to match the new contract interface.


test-helpers/src/arbitrum.rs
```diff
@@ -3,7 +3,7 @@
 /// The deployed address of the WETH ERC20 contract on a devnet node
 pub const PREDEPLOYED_WETH_ADDR: &str = "0x408Da76E87511429485C32E4Ad647DD14823Fdc4";
 /// The default hostport that the Nitro devnet L2 node runs on
-pub const DEFAULT_DEVNET_HOSTPORT: &str = "http://localhost:8547";
+pub const DEFAULT_DEVNET_HOSTPORT: &str = "http://sequencer:8547";
```
joeykraut, line 6: can we make this localhost instead? It's easier to specify a non-default argument in a docker-compose than it is if we're running these tests locally. So I'd prefer all the defaults be ergonomic for out-of-compose use

joeykraut (approved):

PR #337, opened 2024-01-17
arbitrum-client, task-driver: make use of predeployed erc20s in sequencer

akirillo:
This PR updates the `arbitrum-client` and `task-driver` integration testing stacks to:
1. Use the new Arbitrum sequencer image
2. Make use of the pre-deployed, pre-approved dummy ERC20s in the sequencer image

In the case of the `task-driver` integration testing stack, this meant being able to remove a lot of WETH contract interaction code.

**Testing:**
All of the `arbitrum-client` integration tests other than `process_match_settle` pass, as expected.
All of the `task-driver` integration tests other than `test_update_wallet__place_order`, `test_settle_mpc_match`, `test_settle_internal_match`, and `test_lookup_wallet__invalid_wallet` pass (when using the `NO_VERIFY` flag). However these tests fail for reasons seemingly unrelated to the usage of predeployed ERC20s (as opposed to WETH), those failure messages being `wallet not found in wallet_last_updated map`, and `error enqueuing job with proof manager`.
Regardless, fixing these test cases is out of scope for this PR.

PR #342, opened 2024-01-19
task-driver: integration: Fix integration tests

joeykraut:
### Purpose
This PR fixes the `task-driver` integration tests after the new integration testing stack was introduced in #337. The fixes are small and mostly relate to the setup code in how we allocate initial states into the darkpool.

### Testing
- All `task-driver` integration tests pass

### DIFF
--- a/task-driver/integration/helpers.rs
+++ b/task-driver/integration/helpers.rs
@@ -23,7 +23,6 @@ use task_driver::{
     lookup_wallet::LookupWalletTask,
 };
 use test_helpers::{assert_eq_result, assert_true_result};
-use uuid::Uuid;
 
 use crate::IntegrationTestArgs;
 
@@ -45,10 +44,10 @@ pub(crate) async fn lookup_wallet_and_check_result(
     test_args: IntegrationTestArgs,
 ) -> Result<()> {
     // Start a lookup task for the new wallet
-    let new_wallet_id = Uuid::new_v4();
+    let wallet_id = expected_wallet.wallet_id;
     let state = test_args.global_state;
     let task = LookupWalletTask::new(
-        new_wallet_id,
+        wallet_id,
         blinder_seed,
         share_seed,
         expected_wallet.key_chain.clone(),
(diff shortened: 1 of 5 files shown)


task-driver/integration/helpers.rs
```diff
@@ -96,6 +95,22 @@ pub async fn new_wallet_in_darkpool(client: &ArbitrumClient) -> Result<(Wallet,
     Ok((wallet, blinder_seed, share_seed))
 }
 
+/// Sets up a new wallet in the system by:
+///     1. Generating secret shares for the wallet
+///     2. Allocating it in the darkpool directly
+///     3. Looking up the wallet in the contract state so that the wallet
+///        appears in the global state
+pub(crate) async fn setup_initial_wallet(
```
akirillo, line 103: should this method also re-read the wallet out of global state and return it so that order IDs are correct (as we now do in `setup_wallet_with_order_balance`)? that, or set the order IDs on the `&mut wallet`

task-driver/integration/tests/settle_match.rs
```diff
@@ -82,41 +81,25 @@ async fn setup_wallet_with_order_balance(
...
-    wallet.reblind_wallet();
-
-    wallet = execute_wallet_update(
-        old_wallet,
-        wallet.clone(),
-        None, // transfer
-        test_args.clone(),
-    )
-    .await?;
+    setup_initial_wallet(blinder_seed, share_seed, &mut wallet, &test_args).await?;
```
akirillo, line 93: just so we're on the same page: this will only work while verification is disabled (since the proof of `VALID WALLET CREATE` should not verify). this means once we re-enable verification we'll have to reimplement this method to reblind & update the wallet on-chain as it used to do

akirillo (approved):
couple suggestions / qs, but lgtm

task-driver/integration/helpers.rs
joeykraut: It's not generally needed, more just for the settlement tests

task-driver/integration/tests/settle_match.rs
joeykraut: It'd be preferable to expose a contract mock method that allowed direct allocation. Setting up complex wallets without that will be very slow

akirillo: ah, this is what we had discussed internally, right

PR #343, opened 2024-01-19
workers: api-server: wallet: Validate and use timestamp from request

joeykraut:
### Purpose
This PR changes the API server to use the timestamp specified by the client if it is valid (not stale) in the `ValidWalletUpdateStatement`. This keeps the shares of the new wallet consistent with what the client has signed.

### Testing
- Unit tests pass


sehyunc (approved):
LGTM this should do the trick

akirillo (approved):

PR #345, opened 2024-01-21
circuit-types, util: update public key serialization to use uncompressed bytes

sehyunc:
Adds a method to convert the PublicSigningKey to bytes in uncompressed form, for correct serialization to a hex string.

### DIFF
--- a/circuit-types/src/keychain.rs
+++ b/circuit-types/src/keychain.rs
@@ -253,9 +253,9 @@ impl PublicSigningKey {
     }
 
     /// Convert the key to bytes
-    pub fn to_compressed_bytes(&self) -> Vec<u8> {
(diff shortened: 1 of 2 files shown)


circuit-types/src/keychain.rs
```diff
@@ -257,6 +257,12 @@ impl PublicSigningKey {
         let verifying_key = K256VerifyingKey::from(self);
         verifying_key.to_sec1_bytes().to_vec()
     }
+
+    /// Convert the key to bytes in uncompressed form
```
joeykraut, line 261: We can probably delete the compressed bytes method now that we've switched serialization

joeykraut (approved):

PR #348, opened 2024-01-23
statev2: interface: Define state handle

joeykraut:
### Purpose
This PR adds a new `State` type that will be passed around to workers for state access and modification. As a PoC, I added a `new_wallet` method and associated getter and a test for it.

### Testing
- All unit tests pass


akirillo (approved):
Really like this abstraction, nice

PR #349, opened 2024-01-23
statev2: interface: notifications: Add method to wait for proposal

joeykraut:
### Purpose
This PR adds a primitive for a caller to await a proposal's completion, either with an error or successfully. This is done via a oneshot channel that the consensus layer maintains. Once a proposal is either accepted or rejected, the consensus layer responds on this channel.

This will be useful for example in the tasks, which cannot be considered complete until a raft entry is successfully appended representing the task's update.

### Testing
- Modified the tests to await response via this primitive
- All unit tests pass

### DIFF
--- a/util/src/errors.rs
+++ b/util/src/errors.rs
@@ -1,0 +1,11 @@
+//! Helpers for error handling
+
+/// Expands a given error type to wrap a stringified version of a given error
+///
+/// To be used in a map_err() call
+#[macro_export]
+macro_rules! err_str {
+    ($x:expr) => {
+        |e| $x(e.to_string())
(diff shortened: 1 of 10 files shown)


util/src/errors.rs
```diff
@@ -0,0 +1,11 @@
+//! Helpers for error handling
+
+/// Expands a given error type to wrap a stringified version of a given error
+///
+/// To be used in a map_err() call
+#[macro_export]
+macro_rules! err_str {
```
akirillo, line 7: this is tight

akirillo (approved):

PR #351, opened 2024-01-23
statev2: storage: tx: Add raft log access methods

joeykraut:
### Purpose
This PR introduces high-level transaction methods for accessing and modifying the raft log, and replaces the manual implementations at callsites.

### Testing
- All unit tests pass

### DIFF
--- a/statev2/src/storage/tx/raft_log.rs
+++ b/statev2/src/storage/tx/raft_log.rs
@@ -1,0 +1,167 @@
+//! High level transaction interface for accessing the raft log
+
+use std::cmp;
+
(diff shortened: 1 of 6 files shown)


statev2/src/storage/tx/raft_log.rs
```diff
@@ -0,0 +1,167 @@
+//! High level transaction interface for accessing the raft log
```
joeykraut, line 1: mostly just copied over and manipulated to fit the interfaces

akirillo (approved):

PR #353, opened 2024-01-24
statev2: storage: tx: Add `order_book` access methods

joeykraut:
### Purpose
This PR adds access methods for the `order_book` state. This involves changing the applicator as well to use these methods.

### Testing
- Unit tests pass


statev2/src/storage/tx/order_book.rs
```diff
@@ -0,0 +1,308 @@
...
+        let mut order = self.get_order_info_or_err(order_id)?;
+        order.validity_proofs = Some(proof);
+        order.public_share_nullifier = new_nullifier;
+        order.state = NetworkOrderState::Verified;
+
+        self.write_order(&order)
+    }
+
+    /// Cancel an order in the order book
+    pub fn cancel_order(&self, order_id: &OrderIdentifier) -> Result<(), StorageError> {
```
akirillo, line 134: should cancelling an order remove it from the nullifier set?

joeykraut: yeah good call

PR #355, opened 2024-01-24
task-driver, statev2: Update task-driver helpers to use `statev2`

joeykraut:
### Purpose
This PR modifies the `helpers` module in the `task-driver` to use the new `statev2` raft-based primitive. This also involves awaiting consensus when gossiping the validity proofs to cluster peers. We can change this to not wait if we so choose down the line.

### Testing
- `statev2` unit tests pass
- All other tests will pass when integration is complete


task-driver/src/driver.rs
```diff
@@ -110,29 +104,30 @@ pub trait Task: Send {
...
+    // /// The state object for the new wallet task
+    // NewWallet(NewWalletTaskState),
+    // /// The state object for the settle match task
+    // SettleMatch(SettleMatchTaskState),
+    // /// The state object for the settle match internal task
+    // SettleMatchInternal(SettleMatchInternalTaskState),
+    // /// The state object for the update Merkle proof task
+    // UpdateMerkleProof(UpdateMerkleProofTaskState),
+    // /// The state object for the update wallet task
+    // UpdateWallet(UpdateWalletTaskState),
```
joeykraut, line 118: will uncomment this incrementally as I refactor individual tasks

config/src/lib.rs
```diff
@@ -180,7 +180,7 @@ pub struct RelayerConfig {
     /// Bootstrap servers that the peer should connect to
     pub bootstrap_servers: Vec<(WrappedPeerId, Multiaddr)>,
     /// The cluster keypair
-    pub cluster_keypair: Keypair,
+    pub cluster_keypair: DalekKeypair,
```
akirillo, line 183: nit: can we make sure there's a `TODO` somewhere (even internally) to switch to Bn254 ECDSA keys for the cluster?

statev2/src/applicator/mod.rs
```diff
@@ -66,8 +66,8 @@ impl StateApplicator {
             StateTransition::AddWallet { wallet } => self.add_wallet(&wallet),
             StateTransition::UpdateWallet { wallet } => self.update_wallet(&wallet),
             StateTransition::AddOrder { order } => self.new_order(order),
-            StateTransition::AddOrderValidityProof { order_id, proof } => {
-                self.add_order_validity_proof(order_id, proof)
+            StateTransition::AddOrderValidityProof { order_id, proof, witness } => {
+                self.add_order_validity_proof(order_id, proof, witness)
```
akirillo, line 70: out of curiosity, why do we need to share witness w/ peers?
at any rate, why is this an `Option` as opposed to being necessary?

akirillo (approved):
lgtm, just a couple nits/qs

statev2/src/applicator/mod.rs
joeykraut: we share the witness with cluster peers so that they may use it when performing matches. It holds important information like the reblinded shares of the wallet, or the augmented shares for a given order. It's optional because we _also_ have validity proofs from _other_ clusters. They of course will not share their witness

config/src/lib.rs
joeykraut: part of a broader TODO with fees, this keypair is more for gossip

PR #357, opened 2024-01-25
task-driver: update-wallet: Refactor task to use `statev2`

joeykraut:
### Purpose
This PR refactors the `update-wallet` task to use the `statev2` implementation. This includes the `update-wallet` integration tests and bugs fixed by getting these test to pass.

I also changed the way in which we add balances and order to a wallet, overriding old (zero'd) values in the case that a wallet is full.

### Testing
- All unit tests pass
- Integration tests for update-wallet pass


common/src/types/wallet.rs
akirillo, line 1: this is super neat

akirillo (approved):

PR #358, opened 2024-01-26
task-driver: settle-match: Update settlement tasks to use statev2

joeykraut:
### Purpose
This PR updates the `settle-match` and `settle-internal-match` tasks to use the new raft-based statev2 primitive. Tests do not yet pass, a follow up will fix issues with balance indexing.

### Testing
- Test fixes derferred to follow-up


task-driver/integration/tests/settle_match.rs
```diff
@@ -92,14 +92,6 @@ async fn setup_wallet_with_order_balance(
     wallet.orders.insert(Uuid::new_v4(), order);
     setup_initial_wallet(blinder_seed, share_seed, &mut wallet, &test_args).await?;
 
-    // Re-read the wallet out of the global state so that the order IDs match
```
joeykraut, line 95: the helper above was refactored previously to do this step

workers/handshake-manager/src/manager.rs
```diff
@@ -774,7 +774,8 @@ impl HandshakeExecutor {
             self.global_state.clone(),
```
akirillo, line 774: getting a clippy error here saying the `self.global_state` is still the old `RelayerState` type instead of the new `State` type

workers/handshake-manager/src/manager.rs
```diff
@@ -774,7 +774,8 @@ impl HandshakeExecutor {
...
             self.proof_manager_work_queue.clone(),
         )
-        .await;
+        .await
+        .map_err(|err| HandshakeManagerError::TaskError(err.to_string()))?;
         self.task_driver.start_task(task).await.0
```
akirillo, line 779: Error here as well, need to be returning a `Result`

akirillo (approved):
lgtm, just a couple errors

workers/handshake-manager/src/manager.rs
joeykraut: handshake manager isn't refactored yet, errors here are expected. I'm fixing the tasks first then the callers of tasks.

PR #359, opened 2024-01-26
common: types: wallet: Track wallet state in correct order

joeykraut:
### Purpose
This PR fixes the way in which we track orders and balances. Specifically we now use the `OrderedMap` type, which is a vector with a map like interface over it for convenience. The mapping methods are implemented via `find` which is efficient for the small vectors we use this data structure for.

### Testing
- All unit tests pass
- Integration tests that were previously failing with settlement now pass

### DIFF
--- a/common/src/keyed_list.rs
+++ b/common/src/keyed_list.rs
@@ -163,7 +163,7 @@ impl<K: Eq, V: Clone> KeyedList<K, V> {
     /// Removes a key from the map, returning the value at the key if the key
     /// was previously in the map.
     pub fn remove(&mut self, key: &K) -> Option<V> {
-        let index = self.elems.iter().position(|(k, _)| k == key)?;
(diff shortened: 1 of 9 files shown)


common/src/keyed_list.rs
```diff
@@ -0,0 +1,345 @@
...
+
+    /// Replace the key-value pair at the given index
+    pub fn replace_at_index(&mut self, index: usize, key: K, value: V) {
+        self.elems[index] = (key, value);
+    }
+
+    /// Removes a key from the map, returning the value at the key if the key
+    /// was previously in the map.
+    pub fn remove(&mut self, key: &K) -> Option<V> {
+        let index = self.elems.iter().position(|(k, _)| k == key)?;
```
akirillo, line 166: nit: can use `index_of` here

akirillo (approved):
dope

PR #362, opened 2024-01-27
workeres: chain-events: Refactor to use `statev2`

joeykraut:
### Purpose
This PR refactors the `chain-events` worker over the `statev2` crate

### Testing
- Unit tests pass in affected crates

### DIFF
--- a/statev2/src/interface/wallet_index.rs
+++ b/statev2/src/interface/wallet_index.rs
@@ -30,6 +30,15 @@ impl State {
         Ok(wallet_id)
     }
 
+    /// Get the ids of all wallets managed by the local relayer
+    pub fn get_all_wallets(&self) -> Result<Vec<Wallet>, StateError> {
+        let tx = self.db.new_read_tx()?;
+        let wallets = tx.get_all_wallets()?;
+        tx.commit()?;
+
+        Ok(wallets)
+    }
+
     // -----------
     // | Setters |
     // -----------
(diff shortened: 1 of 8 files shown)


statev2/src/interface/wallet_index.rs
```diff
@@ -30,6 +30,15 @@ impl State {
         Ok(wallet_id)
     }
 
+    /// Get the ids of all wallets managed by the local relayer
+    pub fn get_all_wallets(&self) -> Result<Vec<Wallet>, StateError> {
+        let tx = self.db.new_read_tx()?;
+        let wallets = tx.get_all_wallets()?;
+        tx.commit()?;
```
akirillo, line 37: Should we be committing universally across the interface?

akirillo (approved):
Lgtm, just a q about committing DB TXs in the interface

statev2/src/interface/wallet_index.rs
joeykraut: It doesn't matter since it's a read but yeah I'll add that to all these methods so that it's clear what's going on

akirillo: yeah, not necessary for safety, but it is nice to do this for semantic clarity, and fwiw it does seem to have an improvement on storage requirements: https://erthink.github.io/libmdbx/usage.html#:~:text=In%20addition%2C%20as,for%20an%20optimization).

PR #373, opened 2024-02-01
state: Add peers from same cluster to raft group automatically

joeykraut:
### Purpose
This PR implements cluster join semantics for the raft consensus layer. This is done automatically when a node learns of a cluster peer through peer discovery. The discovering node will propose a configuration change to add the newly discovered peer.

### Todo
- Promotions
- Node leave semantics

### Testing
- All unit tests pass
- Ran two peers in the same cluster, verified that they both joined the raft


state/src/interface/raft.rs
```diff
@@ -0,0 +1,26 @@
...
+    // | Setters |
+    // -----------
+
+    /// Add a raft peer as a learner to the local cluster
+    pub fn add_raft_learner(&self, peer_id: WrappedPeerId) -> Result<ProposalWaiter, StateError> {
+        // Add the peer ID to the translation map
+        let mut locked_map = self.translation_map.write().expect("translation map poisoned");
+        locked_map.insert(peer_id);
+        let raft_id = PeerIdTranslationMap::get_raft_id(&peer_id);
+        drop(locked_map); // Unlock the map
```
akirillo, line 21: nit: we can drop the lock before we calculate the raft id, while i'm sure hashing is quick it's not in the critical section here right

state/src/replication/raft_node.rs
```diff
@@ -353,33 +382,41 @@ impl<N: RaftNetwork> ReplicationNode<N> {
...
-                    // applied, either successfully or with an error
-                    self.notify_proposal_sender(&entry_id, res)?;
+                    self.applicator
+                        .handle_state_transition(transition)
+                        .map_err(ReplicationError::Applicator)
                 },
                 EntryType::EntryConfChangeV2 => {
                     // Apply a config change entry to the state machine
                     let mut config_change = ConfChangeV2::new();
+                    // let change_id = parse_conf_change_id(&config_change)?;
```
akirillo, line 393: nit: dead code

akirillo (approved):

PR #375, opened 2024-02-03
config, common: Setup token remaps from config file

joeykraut:
### Purpose
This PR adds chain-specific token remaps parsed in the config-parse step. These remaps are stored in a `static TOKEN_REMAPS` object. They are referenced in place of the addr remaps that previously existed. 

The remap config is parsed in two ways:
- If the `--token-remap-file` is provided, the remap config is parsed from the given file
- Otherwise the mapping file is pulled from the chain specific version in the [token map repo](https://github.com/renegade-fi/token-mappings)

### Testing
- Tested both parsing paths
- Unit tests pass


core/src/main.rs
```diff
@@ -73,7 +73,11 @@ async fn main() -> Result<(), CoordinatorError> {
     // ---------------------
 
     // Parse command line arguments
-    let args = config::parse_command_line_args().expect("error parsing command line args");
+    let args = tokio::task::spawn_blocking(config::parse_command_line_args)
```
akirillo, line 76: do we need to do this? it seems everything downstream of `parse_command_line_args` is synchronous

config/src/token_remaps.rs
```diff
@@ -0,0 +1,158 @@
...
+    // Read the file into a string
+    let file = std::fs::read_to_string(file_path)
+        .map_err(raw_err_str!("Failed to read remap file: {}"))?;
+    serde_json::from_str(&file).map_err(raw_err_str!("Failed to parse remap from file: {}"))
+}
+
+/// Pull the token remap from the repo
+fn fetch_remap_from_repo(chain: Chain) -> Result<TokenRemap, String> {
+    let url = format!("{}{}.json", REMAP_BASE_URL, chain);
+    let resp = reqwest::blocking::get(&url)
```
akirillo, line 73: nit: clippy saying to:
```suggestion
    let resp = reqwest::blocking::get(url)
```

config/src/token_remaps.rs
```diff
@@ -0,0 +1,158 @@
...
+        let (_dir, file, path) = get_temp_dir();
+        let remap = gen_dummy_remap(&file);
+
+        // Parse the remap from the file
+        let parsed = parse_remap_from_file(path).unwrap();
+        assert_eq!(remap, parsed);
+    }
+
+    /// Tests that the token remap is correctly shared across threads
+    #[test]
```
akirillo, line 136: should we also set up a test for fetching the remap from githb?

akirillo (approved):
couple nits/qs, but overall v clean

config/src/token_remaps.rs
joeykraut: i don't want to require the unit tests to have an internet connection, so it's okay to ad hoc

PR #377, opened 2024-02-03
common: token: Fix bug in querying uniswap supported tokens

joeykraut:
### Purpose
This PR fixes a small bug in fetching supported exchanges in that we previously checked the `EXCHANGE_TICKERS` map for `Exchange::UniswapV3` which is not present. Now we filter out uniswap


joeykraut:
@sehyunc will you own testing this change

akirillo (approved):

sehyunc (approved):
Tested fetching WETH prices and matching WETH<>USDC order successfully

PR #379, opened 2024-02-03
docker: release: get Docker image to build

akirillo:
This PR copies all of the requisite dependencies into the release Dockerfile such that the release image can properly build

### DIFF
--- a/docker/release/Dockerfile
+++ b/docker/release/Dockerfile
@@ -13,24 +13,10 @@ RUN cat rust-toolchain | xargs rustup toolchain install
 RUN apt-get update && \
     apt-get install -y pkg-config && \
     apt-get install -y protobuf-compiler && \
-    apt-get install -y libssl-dev
+    apt-get install -y libssl-dev && \
+    apt-get install -y libclang-dev
 
-COPY Cargo.toml Cargo.lock ./
-COPY ./core/Cargo.toml ./core/Cargo.toml
-COPY ./circuits ./circuits
-COPY ./crypto ./crypto
-COPY ./circuit-macros ./circuit-macros
-COPY ./integration-helpers ./integration-helpers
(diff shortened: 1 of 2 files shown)


docker/release/Dockerfile
```diff
@@ -13,14 +13,30 @@ RUN cat rust-toolchain | xargs rustup toolchain install
...
+COPY ./gossip-api ./gossip-api
+COPY ./renegade-crypto ./renegade-crypto
+COPY ./srs ./srs
+COPY ./state ./state
+COPY ./statev2 ./statev2
+COPY ./system-bus ./system-bus
+COPY ./task-driver ./task-driver
+COPY ./test-helpers ./test-helpers
+COPY ./util ./util
+COPY ./workers ./workers
```
joeykraut, line 39: we can just copy in the whole repo now, no need to do one by one. this was done before these were individual crates

workers/network-manager/src/manager.rs
```diff
@@ -267,7 +267,9 @@ impl NetworkManagerExecutor {
                             log::info!("Listening on {}/p2p/{}\n", address, self.local_peer_id);
                         },
                         // This catchall may be enabled for fine-grained libp2p introspection
-                        _ => {  }
+                        e => {
+                            log::info!("{:?}", e);
+                        }
```
joeykraut, line 272: seems this is likely not meant to be in this pr

joeykraut (approved):

workers/network-manager/src/manager.rs
akirillo: ah ack

docker/release/Dockerfile
akirillo: ack

PR #380, opened 2024-02-03
state: tx: task-queue: Add task queue access methods

joeykraut:
### Purpose
This PR implements the storage access primitives for the task queue. This includes `append`, `pop`, and `change_state`. Further layers will build on this to change the state of tasks in the queue as the raft logs controlling the queue are committed.

### Testing
- Unit tests pass

### DIFF
--- a/state/src/storage/tx/task_queue.rs
+++ b/state/src/storage/tx/task_queue.rs
@@ -1,0 +1,172 @@
+//! Defines storage primitives for the task queue
+//!
+//! Tasks are indexed in queues identified by the shared resource they occupy.
+//! For now this is only wallets (i.e. wallet_id).
+
+use common::types::{
+    task_descriptors::{QueuedTask, QueuedTaskState},
+    tasks::TaskIdentifier,
+    wallet::WalletIdentifier,
+};
+use libmdbx::{TransactionKind, RW};
+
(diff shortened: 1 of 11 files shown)


state/src/storage/tx/task_queue.rs
```diff
@@ -0,0 +1,172 @@
...
+        let tasks = tx.get_wallet_tasks(&wallet_id).unwrap();
+        tx.commit().unwrap();
+        assert_eq!(tasks.len(), 1);
+        assert_eq!(tasks[0].id, task.id);
+
+        // Read the task back by ID
+        let tx = db.new_read_tx().unwrap();
+        let task = tx.get_wallet_task_by_id(&wallet_id, &tasks[0].id).unwrap();
+        tx.commit().unwrap();
+        assert_eq!(task.unwrap().id, tasks[0].id);
```
akirillo, line 116: should we check equality of the whole `QueuedTask` struct here to be safe?

akirillo (approved):

state/src/storage/tx/task_queue.rs
joeykraut: Implementing equality is non trivial, some descriptors contain synchronous state and the blast radius is quite large. Tried this originally then abandoned for sake of it taking too much time. This should work for a unit test

PR #381, opened 2024-02-05
workers: task-driver: Refactor task driver into worker pattern

joeykraut:
### Purpose
This PR refactors the task driver into the `Worker` pattern defined throughout the repo. This is done to place a queue in between the state and `task-driver` so that they may reference one another. Additionally, the `task-driver` already holds a Tokio runtime, so it was in many ways already its own worker.

### Todo
- Refactor other workers to use the task queue

### Testing
- Deferred until task queue complete


workers/task-driver/src/driver.rs
```diff
@@ -0,0 +1,298 @@
+//! The task driver drives a task forwards and executes partial retries
```
joeykraut, line 1: mostly moved over, the implementation of `start_task` has changed and the `run` method is new

workers/task-driver/src/tasks/update_wallet.rs
```diff
@@ -263,45 +263,6 @@ impl Task for UpdateWalletTask {
 // -----------------------
 
 impl UpdateWalletTask {
-    /// Constructor
-    ///
-    /// Assumes that the state updates to the wallet have been applied in
-    /// the caller
-    #[allow(clippy::too_many_arguments)]
-    pub fn new(
```
akirillo, line 271: confirming we've removed the call sites for this throughout the codebase?

akirillo (approved):
one small q, lgtm

workers/task-driver/src/tasks/update_wallet.rs
joeykraut: we have not, that's part of the todo. It's more complicated than removing them -- they now need to go through the state module to queue tasks -- so more work to be done before the callsites change

PR #383, opened 2024-02-06
state: applicator: task-queue: Add `PopWalletTask` transition + impl

joeykraut:
### Purpose
This PR adds a `PopWalletTask` state transition that removes the top task from the wallet's task queue. If the queue is non-empty the next task will start running on its executor node.

### Testing
- Unit tests pass

PR #385, opened 2024-02-06
state: interface: Add task queue getters and setters

joeykraut:
### Purpose
This PR adds interface methods for the task queue. This includes appending and popping from the queue. These methods will be used by other workers to start tasks and by the task driver to mark a task as complete.

### Todo
- Update callsites to enqueue tasks
- Task preemption
- Task failover

### Testing
- Unit tests pass


akirillo (approved):
Looks like all of these changes were included in #383

PR #388, opened 2024-02-06
workers: api-server: Refactor to use new task-queue abstraction

joeykraut:
### Purpose
This PR adjusts the `api-server` to use the task queue when creating tasks. This involves going through the raft-backed state. For now, we await consensus on the queue update in the `api-server`, if this proves too burdensome (unlikely for now) we can return early or have an option to do so.'

### Todo
- Validity checks at the API layer

### Testing
- Unit tests pass, will test API more directly after codebase builds

### DIFF
--- a/state/src/interface/task_queue.rs
+++ b/state/src/interface/task_queue.rs
@@ -5,6 +5,7 @@ use common::types::{
     tasks::TaskIdentifier,
     wallet::WalletIdentifier,
 };
+use util::res_some;
 
 use crate::{error::StateError, notifications::ProposalWaiter, State, StateTransition};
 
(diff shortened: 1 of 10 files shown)


state/src/interface/task_queue.rs
```diff
@@ -13,6 +14,15 @@ impl State {
     // | Getters |
     // -----------
 
+    /// Whether or not the task queue contains a specific task
+    pub fn contains_task(&self, task_id: &TaskIdentifier) -> Result<bool, StateError> {
+        let tx = self.db.new_read_tx()?;
+        let wallet_id = self.get_task_wallet(task_id)?;
```
akirillo, line 20: do we expect to have any tasks that are not associated with a wallet? if so, let's call out either in the doc comment or in a code comment that this method assumes all tasks are associated with a wallet, and its implementation would have to change if that assumption is changed

akirillo (approved):

state/src/interface/task_queue.rs
joeykraut: Not right now no we don't

PR #393, opened 2024-02-07
workers: task-driver: Add `RunImmedate` job and initial preemption skeleton

joeykraut:
### Purpose
This PR adds a `RunImmediate` job type for the `task-driver`. This job will preempt any running tasks on the wallet it modifies and run in place. This will be used for match settlement which should be run as soon as the match completes, bypassing the queue.

However, running a preemption is not always possible: other tasks may have committed state on-chain, and cannot be preempted. In this case the match must fail and the task queue should resume.

### Todo
- Do not schedule matches on orders with in-flight updates

### Testing
- State tests pass
- Testing deferred until more components are refactored

### DIFF
--- a/workers/task-driver/src/driver.rs
+++ b/workers/task-driver/src/driver.rs
@@ -11,7 +11,7 @@ use common::{
     new_shared,
     types::{
         tasks::TaskIdentifier,
-        tasks::{QueuedTask, QueuedTaskState, TaskDescriptor},
+        tasks::{QueuedTaskState, TaskDescriptor},
         wallet::WalletIdentifier,
     },
     Shared,
@@ -23,7 +23,6 @@ use state::State;
 use system_bus::SystemBus;
 use tokio::runtime::{Builder as TokioRuntimeBuilder, Runtime as TokioRuntime};
 use tracing::log;
-use util::err_str;
 
 use crate::{
     error::TaskDriverError,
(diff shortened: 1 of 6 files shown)


workers/task-driver/src/driver.rs
```diff
@@ -201,49 +194,89 @@ impl TaskExecutor {
...
+
+        // Start the task optimistically assuming that the queue is paused
+        self.spawn_task_on_runtime(task_id, task);
+
+        // Unpause the queue for the wallet
+        todo!("implement task queue unpausing");
+    }
+
+    /// Spawn the given task on the runtime
+    fn spawn_task_on_runtime(&self, task_id: TaskIdentifier, descriptor: TaskDescriptor) {
```
akirillo, line 223: nit: we can have `spawn_task_on_runtime` return a `Result<(), TaskDriverError>` to match the signatures of `handle_run_immediate` and `handle_notification_request` and make use of the existing logging block after the match statement in `run`

workers/task-driver/src/driver.rs
```diff
@@ -201,49 +194,89 @@ impl TaskExecutor {
...
+        wallet_id: WalletIdentifier,
+        task_id: TaskIdentifier,
+        task: TaskDescriptor,
+    ) -> Result<(), TaskDriverError> {
+        // Check if any non-preemptable tasks conflict with this task
+        if let Some(conflicting_task) = self.state().current_committed_task(&wallet_id)? {
+            log::error!(
+                "run immediate conflicts with committed task {conflicting_task:?}, aborting..."
+            );
+            return Ok(());
```
akirillo, line 209: should we also pop the associated task from the queue, since it cannot be run?

akirillo (approved):
one nit and what i think is one consistency issue, o/w lgtm

workers/task-driver/src/driver.rs
joeykraut: there is no associated task, the idea is to bypass the queue entirely

workers/task-driver/src/driver.rs
joeykraut: it doesn't error

PR #394, opened 2024-02-07
workers: task-driver: pause and unpause state in `RunImmediate`

joeykraut:
### Purpose
This PR implements task queue pausing and unpausing and makes these transitions in the `task-driver` during a `RunImmediate` job. When a queue is paused it places any running tasks onto the queue in the `Queued` state. As well, if a task tries to transition its state when a queue is paused, the raft will fail to apply the log.

In the task driver we check for these failed applications when a task is _first_ committing. If failed, we assume the task has been preempted. 

### Testing
- State unit tests pass
- Testing deferred until more components implemented

### DIFF
--- a/workers/task-driver/src/driver.rs
+++ b/workers/task-driver/src/driver.rs
@@ -17,6 +17,7 @@ use common::{
     Shared,
 };
 use external_api::bus_message::{task_topic_name, SystemBusMessage};
+use futures::Future;
 use job_types::task_driver::{TaskDriverJob, TaskDriverReceiver, TaskNotificationSender};
 use serde::Serialize;
 use state::State;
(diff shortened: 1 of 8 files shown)


workers/task-driver/src/driver.rs
```diff
@@ -280,18 +296,14 @@ impl TaskExecutor {
...
-            if res { Ok(()) } else { Err(TaskDriverError::TaskError("task failed".to_string())) };
-
         // Notify any listeners that the task has completed
-        let str_err = return_val.clone().map_err(|e| e.to_string());
+        let str_err = res.clone().map_err(|e| e.to_string());
         for sender in task_notifications.write().unwrap().remove(&id).unwrap_or_default() {
             let _ = sender.send(str_err.clone());
         }
 
         // Pop the task from the queue so that the next task may run
```
akirillo, line 305: nit: remove this comment

akirillo (approved):

PR #395, opened 2024-02-07
ga: upgrade: check for rollback

akirillo:
This PR adds a final step to the upgrade action to check if the ECS deployment rolled back. The `amazon-ecs-deploy-task-definition` action [exposes no way to check this](https://github.com/aws-actions/amazon-ecs-deploy-task-definition/issues/191). It simply waits for the service to reach a steady state, which it will also do if it rolled back successfully.

Thus, this step checks if the task definition used in the most recent deployment (which may be a rollback) matches the newly created task definition. If not, a rollback must have occurred, and we error.

I have tested this on the staging branch with an image that fails trivially and asserted that a rollback is caught.


rust-toolchain
```diff
@@ -1 +1 @@
-nightly-2023-08-19
+nightly-2024-02-07
```
akirillo, line 1: I bumped to the most recent nightly b/c the most recent version of `cargo-chef` (used to build the image) requires `rustc >= 1.74`. Let me know if this is a concern, if so I can find a `cargo-chef` version to pin to in our release image.

joeykraut: tests are failing can you check that out. this often happens when bumping the version

.github/workflows/upgrade.yml
```diff
@@ -11,7 +11,7 @@ jobs:
   build-and-push-ecr:
```
joeykraut, line 11: feel like we should just call out to a script given how complex this is becoming. food for though, so feel free to disregard if you disagree

joeykraut (approved):

.github/workflows/upgrade.yml
akirillo: The only portion of this that isn't handled by the actions themselves is constructing the matrix and checking for rollbacks. I'm comfortable with this complexity being handled by us, and deferring as much as possible to widely-used actions that we don't have to maintain ourselves. Plus, getting parallelism out of the matrix strategy is neat. I'll keep it this way for now, and revisit if/when we try to optimize CI

rust-toolchain
akirillo: I pushed the version back to the nightly release the day after rust 1.74.1 was released on stable, and the repo builds on the x86_64 Github Actions runner. There's still a failing test `test_default_config`, but this fails on main as well, so out of scope for this PR

PR #404, opened 2024-02-10
workers: price-reporter: mock: Add mock price reporter for tests

joeykraut:
### Purpose
This PR adds a mock price reporter that will be used in integration tests. The mock reporter simply returns a dummy value for each queried pair.

### Testing
- I added tests for the mock

### DIFF
--- a/workers/price-reporter/src/mock.rs
+++ b/workers/price-reporter/src/mock.rs
@@ -1,0 +1,219 @@
+//! A mock price reporter used for testing
+
+use std::collections::HashMap;
+use std::thread;
+
+use bimap::BiMap;
+use common::types::exchange::{
+    Exchange, ExchangeConnectionState, PriceReport, PriceReporterState, ALL_EXCHANGES,
+};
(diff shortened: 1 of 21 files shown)


workers/price-reporter/src/mock.rs
```diff
@@ -0,0 +1,219 @@
...
+        base_token: Token,
+        quote_token: Token,
+        channel: OneshotSender<HashMap<Exchange, ExchangeConnectionState>>,
+    ) -> Result<(), PriceReporterError> {
+        // Construct a state and send it back on the queue
+        let timestamp = get_current_time_seconds();
+        let report = PriceReport {
+            base_token,
+            quote_token,
+            exchange: Some(Exchange::Binance),
```
akirillo, line 124: does it matter that this is the exchange for all of the `ExchangeConnectionState`s?

akirillo (approved):

workers/price-reporter/src/mock.rs
joeykraut: It doesn't matter but I will change it so it reads better, nice catch

PR #406, opened 2024-02-12
workers: api-server: Add `/v0/task_queue/:wallet_id` route + handler

joeykraut:
### Purpose
This PR adds a `/v0/task_queue/:wallet_id` route and handler that returns a set of queued tasks for a wallet. 

### Testing
- Unit tests pass
- Tested a simple `GET` against this endpoint

### DIFF
--- a/external-api/src/http/task.rs
+++ b/external-api/src/http/task.rs
@@ -1,5 +1,6 @@
 //! Defines API types for task status introspection
 
+use common::types::tasks::{QueuedTask, QueuedTaskState, TaskIdentifier};
 use serde::Serialize;
 
 /// The response type for a request to fetch task status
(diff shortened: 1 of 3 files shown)


sehyunc (approved):
LGTM, some clarifying questions:
- Where can I find the possible states of a task? (is it just "queued" + previously existing states?)
- What does it mean when a task has committed?

akirillo (approved):

joeykraut:
> LGTM, some clarifying questions:
> 
> * Where can I find the possible states of a task? (is it just "queued" + previously existing states?)
> * What does it mean when a task has committed?

"queued" is one state, any other state is "running" but the state field is a more descriptive version of the "running" state that is task-dependent. We can change the display logic here if you like

Committed means it won't be preempted -- i.e. if a match comes along then it will not stop this task. You probably don't need to worry about this FE side, but I added it for clarity.

PR #408, opened 2024-02-12
circuit-types: Update types as per fee design

joeykraut:
### Purpose
This PR makes adds the necessary fields and types for fee circuit implementation to the repo. This involves a few big changes:
1. Removed the `Fee` type in favor of the new fee balance
2. Added fee balances to the `Balance` type
3. Added a `Note` type that will be redeemable to the holder of the recipient key. This key is a `PublicIdentificationKey`, so redemption requires proof of preimage. 

### Testing
- Deferred until circuit changes are made, CI is red until then


circuit-types/src/note.rs
```diff
@@ -0,0 +1,43 @@
...
+    blinder: Scalar,
+}
+
+impl Note {
+    /// Compute a commitment to the note
+    pub fn commitment(&self) -> Scalar {
+        let mint = biguint_to_scalar(&self.mint);
+        let amount = Scalar::from(self.amount);
+        let receiver = Scalar::from(self.receiver);
+        let blinder = self.blinder;
```
akirillo, line 39: nit: thoughts on using `Note::to_scalars` here, just to ensure consistency with the scalar serialization?

circuit-types/src/wallet.rs
```diff
@@ -50,13 +52,23 @@ where
...
+    /// The public key of the cluster that this wallet has been delegated to for
+    /// matches
+    ///
+    /// Authorizes fees to be settled out of the wallet by the holder of the
+    /// corresponding private key
+    #[serde(
+        serialize_with = "biguint_to_hex_string",
+        deserialize_with = "biguint_from_hex_string"
+    )]
+    pub managing_cluster: BigUint,
```
akirillo, line 70: should this be a `PublicIdentificationKey`?

akirillo (approved):

PR #410, opened 2024-02-14
circuits: zk-circuits: valid-wallet-update: Add fees and bounds checks

joeykraut:
### Purpose
This PR updates the `VALID WALLET UPDATE` circuit with a number of changes ahead of audit
- We now require that balance, order, transfer amounts be less than $2^{128}$
- We require that any balance, order, or transfer with zero mint(s) be zero in every field
- We require that fees be paid before any withdrawal on a balance takes place
- We no longer require order mints to be unique (a user may place both a sell and buy order for example)
- We remove the timestamp field from the orders, it currently does nothing

### Testing
- Added unit tests for each case written in our circuit spec. For failure cases, I removed constraints that enforce the case, verified that the test failed (constraints were satisfied with an invalid witness) then added back the constraints to ablate test malformation.

### DIFF
--- a/circuits/src/zk_circuits/valid_wallet_update.rs
+++ b/circuits/src/zk_circuits/valid_wallet_update.rs
@@ -10,14 +10,12 @@
 use ark_ff::{One, Zero};
 use circuit_macros::circuit_type;
 use circuit_types::{
-    fixed_point::FixedPointVar,
     keychain::PublicSigningKey,
     merkle::{MerkleOpening, MerkleRoot},
-    order::OrderVar,
     traits::{BaseType, CircuitBaseType, CircuitVarType, SecretShareVarType},
     transfers::{ExternalTransfer, ExternalTransferVar},
     wallet::{Nullifier, WalletShare, WalletShareStateCommitment, WalletVar},
-    PlonkCircuit, AMOUNT_BITS,
+    PlonkCircuit,
 };
 use constants::{Scalar, ScalarField, MAX_BALANCES, MAX_ORDERS, MERKLE_HEIGHT};
 use mpc_plonk::errors::PlonkError;
@@ -26,11 +24,9 @@ use serde::{Deserialize, Serialize};
 
 use crate::{
     zk_gadgets::{
-        comparators::{
-            EqGadget, EqVecGadget, EqZeroGadget, GreaterThanEqZeroGadget, NotEqualGadget,
-        },
+        comparators::{EqGadget, EqVecGadget, EqZeroGadget, NotEqualGadget},
         merkle::PoseidonMerkleHashGadget,
-        wallet_operations::{NullifierGadget, WalletShareCommitGadget},
+        wallet_operations::{AmountGadget, NullifierGadget, WalletShareCommitGadget},
     },
     SingleProverCircuit,
 };
@@ -108,13 +104,7 @@ where
             statement.new_public_shares.unblind_shares(recovered_blinder, cs);
         let new_wallet = unblinded_public_shares.add_shares(&witness.new_wallet_private_shares, cs);
 
-        Self::verify_wallet_transition(
-            &old_wallet,
-            &new_wallet,
-            &statement.external_transfer,
-            statement.timestamp,
-            cs,
-        )
+        Self::verify_wallet_transition(&old_wallet, &new_wallet, &statement.external_transfer, cs)
     }
 
     /// Verify a state transition between two wallets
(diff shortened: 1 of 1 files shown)


circuits/src/zk_circuits/valid_wallet_update.rs
```diff
@@ -210,189 +338,64 @@ where
...
-            let prev_balance_not_zero = cs.logic_neg(prev_balance_zero)?;
-            let new_balance_not_zero = cs.logic_neg(new_balance_zero)?;
-            let valid_mint1 =
-                cs.logic_and_all(&[mints_equal, prev_balance_not_zero, new_balance_not_zero])?;
+            let valid_mint1 = mints_equal;
 
             // Condition 2 -- new mint added to wallet
-            let valid_mint2 =
-                cs.logic_and_all(&[equals_transfer_mint, prev_balance_zero, new_balance_not_zero])?;
+            let valid_mint2 = cs.logic_and(new_mint_equals_transfer_mint, old_mint_is_zero)?;
```
joeykraut, line 376: the invariant that zero mint has zero'd other fields ensures that we need not check the previous balance

circuits/src/zk_circuits/valid_wallet_update.rs
```diff
@@ -210,189 +338,64 @@ where
...
 
             // Condition 3 -- withdrawal of entire balance, mint is now zero
-            let valid_mint3 = cs.logic_and(mint_is_zero, new_balance_zero)?;
+            let valid_mint3 = cs.logic_and(old_mint_equals_transfer_mint, new_mint_is_zero)?;
```
joeykraut, line 379: similar thing here, the invariant that zero-mint balances have zeros in every other fields means we don't have to check the amount here

circuits/src/zk_circuits/valid_wallet_update.rs
```diff
@@ -640,31 +644,170 @@ mod test {
...
+            &mut cs,
+        )
+        .unwrap();
+        assert!(cs.check_circuit_satisfiability(&[]).is_err());
+    }
+
+    /// Tests the case in which the transfer amount is too large
+    #[test]
+    fn test_invalid_transfer__large_amount() {
+        setup_system_logger(LevelFilter::INFO);
```
akirillo, line 774: do we still need this

circuits/src/zk_circuits/valid_wallet_update.rs
```diff
@@ -640,31 +644,170 @@ mod test {
...
+        let mut cs = PlonkCircuit::new_turbo_plonk();
+        let transfer = ExternalTransfer {
+            account_addr: BigUint::from(0u8),
+            mint: BigUint::from(1u8),
+            amount: 0, // replaced
+            direction: ExternalTransferDirection::Deposit,
+        };
+
+        // Allocate the transfer then replace its amount with a large one
+        let large_amt = Scalar::from(u128::MAX) + Scalar::from(2u8);
```
akirillo, line 784: same here re: using `2^AMOUNT_BITS`

circuits/src/zk_circuits/valid_wallet_update.rs
```diff
@@ -673,20 +816,56 @@ mod test {
...
+            &new_wallet,
+            ExternalTransfer {
+                mint: old_wallet.balances[0].mint.clone(),
+                amount: transfer_amt,
+                direction: ExternalTransferDirection::Deposit,
+                account_addr: BigUint::from(0u8),
+            },
+        );
+
+        // Create a secret sharing of u128::MAX + 1 in the order amount field
```
akirillo, line 841: nit: remove

akirillo (approved):
just nits, lgtm

circuits/src/zk_circuits/valid_wallet_update.rs
```diff
@@ -108,31 +104,98 @@ where
...
     ) -> Result<(), CircuitError> {
-        // External transfer must have binary direction
-        cs.enforce_bool(external_transfer.direction)?;
+        // Validate the new wallet's orders
+        Self::validate_new_orders(new_wallet, cs)?;
 
-        // Validate updates to the orders within the wallet
-        Self::validate_order_updates(old_wallet, new_wallet, update_timestamp, cs)?;
+        // Validate the transfer
+        let transfer_is_zero = EqZeroGadget::eq_zero(&external_transfer.amount, cs)?;
```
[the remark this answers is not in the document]
joeykraut: yeah good catch, original version didn't match the spec

circuits/src/zk_circuits/valid_wallet_update.rs
```diff
@@ -640,31 +644,170 @@ mod test {
...
+        // Construct a statement and witness then modify the order amount of the first
+        // order to be too large
+        let (mut witness, mut statement) =
+            construct_witness_statement(&old_wallet, &new_wallet, ExternalTransfer::default());
+
+        // Create a secret sharing of u128::MAX + 1 in the order amount field
+        let blinder =
+            statement.new_public_shares.blinder + witness.new_wallet_private_shares.blinder;
+
+        statement.new_public_shares.orders[0].amount = Scalar::from(u128::MAX) + Scalar::from(1u8);
```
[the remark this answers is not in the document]
joeykraut: good call

PR #412, opened 2024-02-15
state: storage: cursor: Add key filters

joeykraut:
### Purpose
This PR adds key filters to the cursor type. This allows us to filter only pass keys which are relevant to a query rather than the entire table into the cursor consumer. Filtering is particularly useful when a table has multiple key types in it, e.g. the network order table.

This fixes a bug that occurs when constructing heartbeat messages -- the network order table is incorrectly filtered, and the cursor attempts to deserialize other entry type to `PeerInfo` types. 

I also modified the iterator on the cursor to implement `IntoIterator` so that the semantics are more clear, i.e. iterating consumes the cursor rather than allowing for mutations in between iterations.

### Testing
- Replicated the bug that @sehyunc observed and verified it was fixed after this PR
- Unit tests pass workspace-wide

### DIFF
--- a/state/src/storage/cursor.rs
+++ b/state/src/storage/cursor.rs
@@ -20,104 +20,178 @@ use super::{
 pub struct DbCursor<'txn, Tx: TransactionKind, K: Key, V: Value> {
     /// The underlying cursor
     inner: Cursor<'txn, Tx>,
-    /// The buffered kv pair at the current position
+    /// The filter applied to keys
     ///
-    /// This field is used to provide a `seek` implementation that does not
-    /// consume the value under the cursor after seeking. We prefer to seek
-    /// and then separately read the value out
-    buffered_value: Option<(K, V)>,
-    /// A phantom data field to hold the deserialized type of the table
+    /// Only keys for which this filter returns `true` will be returned by the
+    /// cursor methods
+    ///
+    /// Note however that the filter does not affect the position of the cursor,
+    /// so while methods like `next` and `prev` will skip over keys that do not
+    /// pass the filter, the cursor will still be positioned at those keys
+    key_filter: Option<fn(&K) -> bool>,
+    /// A phantom data field to hold the deserialized type of
+    /// the table
     _phantom: PhantomData<(K, V)>,
 }
 
-/// Query methods
+// -----------------
+// | Query Methods |
+// -----------------
+
 impl<'txn, Tx: TransactionKind, K: Key, V: Value> DbCursor<'txn, Tx, K, V> {
     /// Constructor
     pub fn new(cursor: Cursor<'txn, Tx>) -> Self {
-        Self { inner: cursor, buffered_value: None, _phantom: PhantomData }
+        Self { inner: cursor, key_filter: None, _phantom: PhantomData }
     }
 
-    /// Return an iterator over only the keys in the table
-    pub fn keys(self) -> impl Iterator<Item = Result<K, StorageError>> {
-        <Self as Iterator>::map(self, |res| res.map(|(k, _v)| k))
-    }
-
-    /// Return an iterator over only the values in the table
-    pub fn values(self) -> impl Iterator<Item = Result<V, StorageError>> {
-        <Self as Iterator>::map(self, |res| res.map(|(_k, v)| v))
+    /// Set a filter on the keys
+    ///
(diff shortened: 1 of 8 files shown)


state/src/storage/cursor.rs
```diff
@@ -396,4 +427,46 @@ mod test {
...
+    fn test_filtered_cursor() {
+        // Create a db and insert 0..N into the table randomly
+        const N: usize = 1000;
+        let db = mock_db();
+        db.create_table(TEST_TABLE).unwrap();
+
+        insert_n_random(&db, N);
+
+        // Create a filtered cursor that only returns even keys
+        let tx = db.new_raw_read_tx().unwrap();
```
joeykraut, line 442: the db auto-orders string keys, hence using a string key and not a numeric type.

akirillo (changes requested):
I feel that the filter should, actually affect the position of the cursor. `None` values returned from the cursor are now ambiguous: has the cursor processed all of its contents, or did the current element not match the filter?

For example, much of the functionality built on top of the `Iterator` trait assumes that an iterator is exhausted upon the first time it returns `None`. I ran your test, and not a single iteration of the for loop is executed. Other places in the codebase that make use of the `Iterator` impl (such as the `keys` and `values` functions) may similarly short-circuit early unless the caller knows to "push past" the first `None`, but at that point, when *should* the iterator stop?

The solution using the current approach would be to make sure the table's keys are sorted such that those matching the filter are 1) earliest in the sort and 2) consecutive

joeykraut:
> I feel that the filter should, actually affect the position of the cursor. `None` values returned from the cursor are now ambiguous: has the cursor processed all of its contents, or did the current element not match the filter?
> 
> 
> 
> For example, much of the functionality built on top of the `Iterator` trait assumes that an iterator is exhausted upon the first time it returns `None`. I ran your test, and not a single iteration of the for loop is executed. Other places in the codebase that make use of the `Iterator` impl (such as the `keys` and `values` functions) may similarly short-circuit early unless the caller knows to "push past" the first `None`, but at that point, when *should* the iterator stop?

That's a bug in the iterator, should definitely not move the cursor position, we can just continue searching if the filter isn't passed

state/src/storage/cursor.rs
```diff
@@ -20,104 +20,178 @@ use super::{
...
 
     /// Position at the first key greater than or equal to the given key
     pub fn seek_geq(&mut self, k: &K) -> Result<(), StorageError> {
+        // Position the cursor at the first key then seek forward to the next valid key
         let k_bytes = serialize_value(k)?;
-        let res =
-            self.inner.set_range::<CowBuffer, CowBuffer>(&k_bytes).map_err(StorageError::TxOp)?;
-
-        self.buffer_kv_pair(res)
+        self.inner.set_range::<CowBuffer, CowBuffer>(&k_bytes).map_err(StorageError::TxOp)?;
```
akirillo, line 132: should we check that the first key in the range matches the filter? If it doesn't, it gives us a chance to error early instead of potentially seeking to the end of the table and having the cursor appear empty

state/src/storage/cursor.rs
```diff
@@ -396,4 +509,60 @@ mod test {
...
+        let prev = cursor.get_current().unwrap().unwrap();
+        assert_eq!(prev, ("1".to_string(), 1));
+
+        // Seek to the last key, should be N - 1
+        cursor.seek_last().unwrap();
+        let last = cursor.get_current().unwrap().unwrap();
+        let idx = N - 1;
+        assert_eq!(last, (idx.to_string(), idx));
+
+        // Iterate over the cursor, should only return even keys
```
akirillo, line 555: nit typo
```suggestion
        // Iterate over the cursor, should only return odd keys
```

akirillo (approved):

state/src/storage/cursor.rs
joeykraut: I prefer it this way, that isn't really an erroneous case, it's just an empty cursor. This matches what would happen if _no_ filter is applied, i.e. if I have $1, \ldots, 9$ in the table and do `cursor.seek_geq("10")` I'll just get an empty cursor, not an error. It doesn't seem sensible to error only if a filter is there.

Plus semantically, empty range queries are completely valid and useful

PR #413, opened 2024-02-15
common: Add wallet balance to request amount when depositing

sehyunc:
### Purpose
This PR fixes a bug in the `add_balance` function where the `balance` variable was being overloaded with both the value to be added (deposited) and the balance that was currently in the wallet. This resulted in the post-deposit amount being incorrect (deposits would go from 1, 2, 4, 8, etc.)

### Testing
- Tested in the frontend


joeykraut (approved):
Wow nice catch

PR #417, opened 2024-02-17
workers: price-reporter: Do not fail on invalid pair

joeykraut:
### Purpose
This PR changes the price reporter to not fail on invalid asset pairs, i.e. if no exchange supports the pair. In this case the `Unsupported` message is returned to the caller. 

I tested this through the API to verify that this unsupported information makes its way back to the client.

### Testing
- All unit tests pass

### DIFF
--- a/workers/price-reporter/src/reporter.rs
+++ b/workers/price-reporter/src/reporter.rs
@@ -128,8 +128,13 @@ impl Reporter {
         quote_token: Token,
         config: PriceReporterConfig,
     ) -> Result<Self, ExchangeConnectionError> {
+        // Get the supported exchanges for the token pair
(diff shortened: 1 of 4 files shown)


sehyunc (approved):
Can confirm error makes its way to the client

workers/price-reporter/src/reporter.rs
```diff
@@ -128,8 +128,13 @@ impl Reporter {
         quote_token: Token,
         config: PriceReporterConfig,
     ) -> Result<Self, ExchangeConnectionError> {
+        // Get the
```
akirillo, line 131: nit: complete this

akirillo (approved):

PR #418, opened 2024-02-19
core, util: datadog tracing & logs

akirillo:
This PR makes the necessary changes to export traces to DataDog, and ensure that logs are appropriately linked to those traces. This entails the following:
1. The DataDog Agent must be running on the same host, and be configured to expose a gRPC OTLP collector endpoint. By default, the endpoint is expected to be `http://localhost:4317`, but it can be overridden by setting the [appropriate environment variables](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#endpoint-urls-for-otlphttp) in the relayer
2. We send traces constructed by the `tracing` crate to the OTLP collector, if the `trace-otlp` feature flag is set
3. We format logs as JSON objects including the trace/span IDs (in DataDog format) in which they were emitted, if the `datadog` feature flag is set. This required vendoring in the `DatadogFormatter` from https://github.com/will-bank/datadog-tracing/blob/main/src/formatter.rs, we do this so that we can configure the additional file/line number fields on the logs

I've tested a POC of this by running the relayer and DataDog agent locally, and can confirm that traces/logs are sent and linked:
Screenshot 2024-02-19 at 2 00 59 PM


util/src/logging/formatter.rs
```diff
@@ -0,0 +1,174 @@
...
+            serializer.serialize_entry("level", &meta.level().as_serde())?;
+            serializer.serialize_entry("target", meta.target())?;
+
+            if let Some(filename) = meta.file() {
+                serializer.serialize_entry("filename", filename)?;
+            }
+
+            if let Some(line_number) = meta.line() {
+                serializer.serialize_entry("line_number", &line_number)?;
+            }
```
akirillo, line 99: This file was copied form https://github.com/will-bank/datadog-tracing/blob/main/src/formatter.rs except for these lines, which were added (and themselves taken from https://github.com/tokio-rs/tracing/blob/master/tracing-subscriber/src/fmt/format/json.rs#L260-L270)

akirillo:
@joeykraut Am curious on your thoughts re: configuration via env vars as opposed to the config file here. On one hand, these env vars are standardized for OTLP, and we avoid cluttering the config file / CLI with more options. On the other, the config file is already an established pattern, and now we are fragmenting configuration across it and env vars

util/src/logging/formatter.rs
```diff
@@ -0,0 +1,174 @@
...
+    fn from(value: SpanId) -> Self {
+        Self(u64::from_be_bytes(value.to_bytes()))
+    }
+}
+
+fn lookup_trace_info<S>(span_ref: &SpanRef<S>) -> Option<TraceInfo>
+where
+    S: Subscriber + for<'a> LookupSpan<'a>,
+{
+    span_ref.extensions().get::<OtelData>().map(|o| {
```
joeykraut, line 54: this needs a comment to describe what is going on

util/src/logging/formatter.rs
```diff
@@ -0,0 +1,174 @@
...
+
+            serializer.end()
+        };
+
+        visit().map_err(|_| std::fmt::Error)?;
+        writeln!(writer)
+    }
+}
+
+struct WriteAdaptor<'a> {
```
joeykraut, line 121: nit: typo `Adapter`

joeykraut (approved):
CI is failing as well, likely clippy?

core/Cargo.toml
```diff
@@ -6,6 +6,9 @@ default-run = "renegade-relayer"
 
 [features]
 debug-tui = []
+trace-otlp = ["util/trace-otlp", "tracing-opentelemetry", "opentelemetry"]
+metrics = []
```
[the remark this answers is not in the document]
akirillo: going to address this and your comment re: `cfg!` in the next PR, want to lift configuration to the CLI / config file

PR #419, opened 2024-02-20
circuits: zk-circuits: valid-match-settle: singleprover circuit fees

joeykraut:
### Purpose
This PR adds fees, extra range constraints, and tests to the single-prover `VALID MATCH SETTLE` circuit. The MPC circuits and the multiprover circuit are deferred for a follow up.

### Testing
- All unit tests pass except multiprover tests

### DIFF
--- a/circuit-types/src/match.rs
+++ b/circuit-types/src/match.rs
@@ -17,9 +17,6 @@ use crate::{
 
 /// Represents the match result of a matching MPC in the cleartext
 /// in which two tokens are exchanged
-/// TODO: When we convert these values to fixed point rationals, we will need to
-/// sacrifice one bit of precision to ensure that the difference in prices is
-/// divisible by two
 #[circuit_type(serde, singleprover_circuit, mpc, multiprover_circuit)]
 #[derive(Debug, Default, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct MatchResult {
@@ -36,18 +33,11 @@ pub struct MatchResult {
     /// sells the quote
     pub direction: bool,
 
-    /// The following are supporting variables, derivable from the above, but
-    /// useful for shrinking the size of the zero knowledge circuit. As
-    /// well, they are computed during the course of the MPC, so it incurs
-    /// no extra cost to include them in the witness
-
-    /// The minimum amount of the two orders minus the maximum amount of the two
-    /// orders. We include it here to tame some of the non-linearity of the
-    /// zk circuit, i.e. we can shortcut some of the computation and
-    /// implicitly constrain the match result with this extra value
-    pub max_minus_min_amount: Amount,
     /// The index of the order (0 or 1) that has the minimum amount, i.e. the
     /// order that is completely filled by this match
+    ///
+    /// This is computed in the MPC and included in the witness as a hint to
+    /// accelerate the collaborative proof constraint generation
     ///
     /// We serialize this as a `bool` to automatically constrain it to be 0 or 1
     /// in a circuit. So `false` means 0 and `true` means 1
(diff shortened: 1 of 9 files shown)


circuits/src/zk_circuits/valid_match_settle/mod.rs
```diff
@@ -493,17 +585,6 @@ mod tests {
         assert!(!check_constraint_satisfaction::<SizedValidMatchSettle>(&witness, &statement));
     }
 
-    /// Test the case in which the `max_minus_min` field is incorrectly computed
-    #[test]
-    fn test_valid_match__incorrect_max_minus_min() {
```
joeykraut, line 498: we removed this field from the witness (it was redundant) so this test is no longer necessary

circuit-types/src/match.rs
```diff
@@ -36,19 +33,12 @@ pub struct MatchResult {
...
-    /// The following are supporting variables, derivable from the above, but
-    /// useful for shrinking the size of the zero knowledge circuit. As
-    /// well, they are computed during the course of the MPC, so it incurs
-    /// no extra cost to include them in the witness
-
-    /// The minimum amount of the two orders minus the maximum amount of the two
-    /// orders. We include it here to tame some of the non-linearity of the
-    /// zk circuit, i.e. we can shortcut some of the computation and
-    /// implicitly constrain the match result with this extra value
-    pub max_minus_min_amount: Amount,
```
joeykraut, line 48: this field was a redundant artifact of how we used to structure this circuit, no longer needed so i removed it

circuits/src/zk_circuits/valid_match_settle/single_prover.rs
```diff
@@ -333,6 +386,24 @@ where
         Ok(())
     }
 
+    /// Validate that the receive balance amounts after update do not overflow
+    fn validate_no_receive_overflow(
```
akirillo, line 390: If we are constraining the bit-length of the wallet's `match_fee` in `VALID WALLET CREATE`, do we still need these bound checks here? Same goes for the protocol fee balance, we sort of already assume it's set correctly by the contract (though that should be documented if not constrained)

akirillo (approved):

circuits/src/zk_circuits/valid_match_settle/mod.rs
```diff
@@ -95,6 +95,14 @@ where
...
     #[link_groups = "valid_commitments_match_settle0"]
     pub balance1: Balance,
+    /// The first party's receive balance
+    #[link_groups = "valid_commitments_match_settle0"]
+    pub balance_receive1: Balance,
+    /// The first party's managing relayer fee
+    #[link_groups = "valid_commitments_match_settle1"]
+    pub relayer_fee1: FixedPoint,
+    /// The first party's fee obligations as a result of the match
+    pub party0_fees: FeeTake,
```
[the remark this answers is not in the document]
joeykraut: Yes this is fair

circuits/src/zk_circuits/valid_match_settle/single_prover.rs
joeykraut: Yes we do, the balance fields can still overflow. Suppose you _started out_ with a very large balance and basically any match would put you over (there is a test for this), then regardless of the validation we've done to the fee, we still will violate the invariants. We really should strive to hold these as invariants (absolute) rather than rules of thumb, I'll give two reasons why:
1. The invariants around wallet value ranges are mostly to avoid overflowing/underflowing the field, but they also simplify the circuit designs of other circuits. If we maintain them absolutely we don't have to worry as much about interconnectedness.
2. Fewer total range checks need to be done in the system if we assume that each circuit does its part to maintain the invariants with no exeception

PR #423, opened 2024-02-24
circuits: zk-circuits: valid-relayer-fee-settlement: Add testing

joeykraut:
### Purpose
This PR adds tests for the `VALID RELAYER FEE SETTLEMENT` circuit. For each test I commented out expected constraints to make it fail, then re-added the constraints and verified that the test passed

### Testing
- All unit tests pass for this circuit


circuits/src/zk_circuits/valid_relayer_fee_settlement.rs
```diff
@@ -415,3 +415,589 @@ where
...
+        [(); MAX_BALANCES + MAX_ORDERS]: Sized,
+    {
+        let mut rng = thread_rng();
+        let mut sender_wallet = sender_wallet.clone();
+        let mut recipient_wallet = recipient_wallet.clone();
+
+        // Pick balances to send and receive the fees from
+        let send_idx = rng.gen_range(0..MAX_BALANCES);
+        let receive_idx = rng.gen_range(0..MAX_BALANCES);
+        recipient_wallet.balances[receive_idx].mint = sender_wallet.balances[send_idx].mint.clone();
```
akirillo, line 461: this assumes that `recipient_wallet` doesn't have a balance w/ the same mint as `sender_wallet.balances[send_idx]`, if it does then the random `receive_idx` may be a different index and we'd violate the invariant of unique balance mints. something the caller should be aware of

circuits/src/zk_circuits/valid_relayer_fee_settlement.rs
```diff
@@ -415,3 +415,589 @@ where
...
+    /// Test the case in which a fee is settled into a wallet with zero'd
+    /// balances
+    #[test]
+    fn test_valid_witness__zero_initial_receiver_balance() {
+        let (sender_wallet, mut recipient_wallet) = get_initial_wallets();
+        for bal in recipient_wallet.balances.iter_mut() {
+            *bal = Balance::default();
+        }
+
+        let (statement, witness) = create_witness_statement(&sender_wallet, &recipient_wallet);
```
akirillo, line 626: the zero mint in the default balance on the recipient wallet gets overwritten in `create_witness_statement`, so this won't test the settlement of a fee into a balance with a previously zero mint

circuits/src/zk_circuits/valid_relayer_fee_settlement.rs
```diff
@@ -415,3 +415,589 @@ where
...
+        let modification_idx = rng.gen_range(0..shares_len);
+        let modification = Scalar::random(&mut rng);
+
+        // Modify the private shares
+        let mut private_shares = witness.sender_updated_private_shares.to_scalars();
+        private_shares[modification_idx] += modification;
+        witness.sender_updated_private_shares =
+            WalletShare::from_scalars(&mut private_shares.into_iter());
+        statement.sender_wallet_commitment =
+            compute_wallet_private_share_commitment(&witness.sender_updated_private_shares);
```
akirillo, line 682: we should also update the nullifier appropriately, right, so that constraint isn't what fails

circuits/src/zk_circuits/valid_relayer_fee_settlement.rs
```diff
@@ -415,3 +415,589 @@ where
...
+        public_shares[modification_idx] -= modification;
+        statement.sender_updated_public_shares =
+            WalletShare::from_scalars(&mut public_shares.into_iter());
+
+        assert!(!check_constraints_satisfied(&statement, &witness));
+    }
+
+    /// Tests the case in which the public blinder share is incorrect
+    #[test]
+    fn test_invalid_public_blinder_share() {
```
akirillo, line 696: do we also want a test case for incorrect public secret shares, to test them being implicitly constrained by the state transition? that checks out to me mentally, but a test may be good to confirm

akirillo (approved):

circuits/src/zk_circuits/valid_relayer_fee_settlement.rs
joeykraut: correct, for testing sake this is okay though because we explicitly mask the index

circuits/src/zk_circuits/valid_relayer_fee_settlement.rs
joeykraut: that is what the test above this does

circuits/src/zk_circuits/valid_relayer_fee_settlement.rs
joeykraut: we didn't update the old wallet's shares

PR #424, opened 2024-02-24
circuits: zk-circuits: valid-offline-fee-settlement: Add initial impl

joeykraut:
### Purpose
This PR adds the initial implementation of the `VALID OFFLINE FEE SETTLEMENT` circuit. This circuit allows the managing party to settle the party's fee balances out of their wallet as an encrypted `Note`, which is inserted into the global state by the contract.

Testing is deferred to a follow-up.


circuits/src/zk_circuits/valid_offline_fee_settlement.rs
```diff
@@ -0,0 +1,356 @@
...
+    /// Verify coherence between the note and the send balance
+    fn verify_note_construction(
+        note: &NoteVar,
+        is_protocol_fee: BoolVar,
+        send_bal: &BalanceVar,
+        enc_key: &EncryptionKeyVar,
+        cs: &mut PlonkCircuit,
+    ) -> Result<(), CircuitError> {
+        // The note's mint should match the send balance
+        EqGadget::constrain_eq(&note.mint, &send_bal.mint, cs)?;
```
akirillo, line 135: do we also need the check that $\text{Note}.\text{mint} \neq 0$?

akirillo (approved):
very nice to see the mileage we're getting out of these helpers

circuits/src/zk_circuits/valid_offline_fee_settlement.rs
joeykraut: fixed in follow up

PR #425, opened 2024-02-24
circuits: valid-offline-fee-settlement: Add test setup and helpers

joeykraut:
### Purpose
This PR sets up some initial utilities for `VALID OFFLINE FEE SETTLEMENT` and refactors types for their use in other crates.  Mostly this is moving code around so that it may be referenced in more general crates (i.e. those that depend on `circuit-types`). 

Testing will come in a follow-up


circuits/src/zk_circuits/valid_offline_fee_settlement.rs
```diff
@@ -354,3 +351,156 @@ where
...
+            &original_wallet_private_shares,
+        );
+        let (root, opening) = create_multi_opening(&[commitment]);
+        let nullifier = compute_wallet_share_nullifier(commitment, wallet.blinder);
+
+        // Create the note
+        let note = Note {
+            mint: send_mint,
+            amount: send_amt,
+            receiver: wallet.managing_cluster,
```
akirillo, line 433: this should be `key` so that it depends on `is_protocol` appropriately

akirillo (approved):

circuits/src/zk_circuits/valid_offline_fee_settlement.rs
joeykraut: fixed in follow up

PR #427, opened 2024-02-25
circuits: valid-fee-redemption: Add circuit implementation + tests

joeykraut:
### Purpose
This PR implements the `VALID FEE REDEMPTION` circuit and associated tests. This circuit is used to settle a note into the recipient's wallet. Note settlement is authorized by the recipient proving knowledge of the decryption key associated with the note.

### Testing
- All tests pass

### DIFF
--- a/circuits/src/zk_circuits/valid_fee_redemption.rs
+++ b/circuits/src/zk_circuits/valid_fee_redemption.rs
@@ -1,0 +1,769 @@
+//! Settles a note into the recipient wallet, nullifying the note in the global
+//! state
+
+use circuit_macros::circuit_type;
+use circuit_types::elgamal::DecryptionKey;
+use circuit_types::keychain::PublicSigningKey;
+use circuit_types::PlonkCircuit;
+use circuit_types::{
+    merkle::{MerkleOpening, MerkleRoot},
+    note::{Note, NoteVar},
+    traits::{BaseType, CircuitBaseType, CircuitVarType, SingleProverCircuit},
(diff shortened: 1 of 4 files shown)


circuits/src/zk_circuits/valid_fee_redemption.rs
```diff
@@ -0,0 +1,746 @@
...
+            // zero'd balance
+            let was_zerod = EqVecGadget::eq_zero_vec(
+                &[old_bal.amount, old_bal.relayer_fee_balance, old_bal.protocol_fee_balance],
+                cs,
+            )?;
+            let old_mint_equals_note_mint = EqGadget::eq(&old_bal.mint, &note.mint, cs)?;
+            let valid_mint_replacement = cs.logic_or(was_zerod, old_mint_equals_note_mint)?;
+
+            let valid_old_mint = cs.logic_or(is_not_receive_idx, valid_mint_replacement)?;
+            cs.enforce_true(valid_old_mint)?;
```
akirillo, line 157: I don't believe these constraints are sufficient to ensure the right mint is being used across the old balance, new balance, and note. Take, for example:

`is_receive_idx = true`
`old_bal.mint = ETH`, o/w zeroed
`new_bal.mint = ETH`
`note.mint = SHITCOIN`

we will have `valid_new_mint == true && valid_old_mint == true`

akirillo (changes requested):
i think there's an issue with the mint constraints

akirillo (approved):

PR #435, opened 2024-02-28
circuits: mpc-circuits: settle: Compute & settle fees in MPC

joeykraut:
### Purpose
This PR adds fee computation and settlement to the `settle` MPC circuit. Single-sided fees are computed and then used to settle the match into the wallets. 

I also added a few fixes for clippy lints in `circuits` -- the repo is compiling enough locally for clippy lints to show up.

### Testing
- Updated settlement unit tests to account for fees
- Unit tests pass

### DIFF
--- a/circuits/src/mpc_circuits/settle.rs
+++ b/circuits/src/mpc_circuits/settle.rs
@@ -1,24 +1,33 @@
 //! Settles a match into secret shared wallets
 
 use circuit_types::{
-    r#match::{AuthenticatedMatchResult, OrderSettlementIndices},
+    fixed_point::{AuthenticatedFixedPoint, PROTOCOL_FEE_FP},
+    r#match::{AuthenticatedFeeTake, AuthenticatedMatchResult, OrderSettlementIndices},
     wallet::AuthenticatedWalletShare,
+    Fabric,
 };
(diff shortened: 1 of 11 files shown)


circuits/src/mpc_circuits/settle.rs
```diff
@@ -44,20 +53,61 @@ where
...
+    fabric: &Fabric,
+) -> (AuthenticatedFeeTake, AuthenticatedFeeTake) {
+    // Relayer fees
+    let party0_relayer_fee_fp = relayer_fee0 * party0_buy_amount;
+    let party0_relayer_fee = FixedPointMpcGadget::as_integer(&party0_relayer_fee_fp, fabric);
+    let party1_relayer_fee_fp = relayer_fee1 * party1_buy_amount;
+    let party1_relayer_fee = FixedPointMpcGadget::as_integer(&party1_relayer_fee_fp, fabric);
+
+    // Protocol fees
+    let protocol_fee = *PROTOCOL_FEE_FP;
```
akirillo, line 101: This is ok for now, but just calling out that down the line we should read the protocol fee from the contract

akirillo (approved):

PR #436, opened 2024-02-29
arbitrum-client, workers: task-driver, api-server: add `TransferAuxData` to wallet updates

akirillo:
This PR introduces a `TransferAuxData` type analogous to the one present in the contracts repo which encapsulates auxiliary data needed to validate an external transfer. The type is used in the `update_wallet` `ArbitrumClient` method and throughout the task driver and API server as appropriate (i.e., we now expect this struct to be submitted alongside deposit/withdrawal API requests).

When patching the rest of the repo to compile, everything builds, but updates to the existing Arbitrum client and task driver integration test are deferred for the next PR as there are a lot of helpers to be ported over from the contracts repo.

### DIFF
--- a/arbitrum-client/src/conversion.rs
+++ b/arbitrum-client/src/conversion.rs
@@ -150,39 +150,28 @@ pub fn to_contract_valid_wallet_update_statement(
 
 /// Convert a [`TransferAuxData`] to its corresponding smart contract type
 pub fn to_contract_transfer_aux_data(
-    data: &Option<TransferAuxData>,
+    data: TransferAuxData,
 ) -> Result<ContractTransferAuxData, ConversionError> {
-    let res = if let Some(data) = data {
-        match data {
-            TransferAuxData::Deposit(deposit) => ContractTransferAuxData {
-                permit_nonce: Some(
-                    AlloyU256::from_str(&biguint_to_hex_string(&deposit.permit_nonce))
-                        .map_err(|_| ConversionError::InvalidUint)?,
-                ),
-                permit_deadline: Some(
-                    AlloyU256::from_str(&biguint_to_hex_string(&deposit.permit_deadline))
-                        .map_err(|_| ConversionError::InvalidUint)?,
-                ),
-                permit_signature: Some(deposit.permit_signature.clone()),
-                transfer_signature: None,
-            },
-            TransferAuxData::Withdrawal(withdrawal) => ContractTransferAuxData {
-                permit_nonce: None,
-                permit_deadline: None,
-                permit_signature: None,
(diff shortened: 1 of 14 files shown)


arbitrum-client/src/conversion.rs
```diff
@@ -141,6 +148,43 @@ pub fn to_contract_valid_wallet_update_statement(
     })
 }
 
+/// Convert a [`TransferAuxData`] to its corresponding smart contract type
+pub fn to_contract_transfer_aux_data(
```
joeykraut, line 152: I'd suggest having this method take in `TransferAuxData`, implement `Default` on `ContractTransferAuxData` and in the caller we do a `.map(<this-method>).unwrap_or_default()`

This method's control flow would be cleaned up

common/src/types/tasks.rs
```diff
@@ -319,6 +320,8 @@ pub struct UpdateWalletTaskDescriptor {
     pub timestamp_received: u64,
     /// The external transfer, if one exists
     pub external_transfer: Option<ExternalTransfer>,
+    /// The auxiliary data for the external transfer, if one exists
```
joeykraut, line 323: might be cleaner to build `ExternalTransferWithAuthData` or similar. These two fields should always either both be `Some` or `None`. 

We can also add constructors for this like `ExternalTransferWithAuthData::deposit` (enforcing the direction to be a deposit) and `ExternalTransferWithAuthData::withdrawal`

arbitrum-client/src/conversion.rs
```diff
@@ -141,6 +148,43 @@ pub fn to_contract_valid_wallet_update_statement(
...
+                        .map_err(|_| ConversionError::InvalidUint)?,
+                ),
+                permit_deadline: Some(
+                    AlloyU256::from_str(&biguint_to_hex_string(&deposit.permit_deadline))
+                        .map_err(|_| ConversionError::InvalidUint)?,
+                ),
+                permit_signature: Some(deposit.permit_signature.clone()),
+                transfer_signature: None,
+            },
+            TransferAuxData::Withdrawal(withdrawal) => ContractTransferAuxData {
```
joeykraut, line 169: Why don't we just mimic the structure or even use the contracts type, don't see the need to separate out by withdrawals. Like I said elsewhere, we can group this with the external transfer struct and provide convenient constructors. These constructors could set fields to `None` as needed.

This would obviate the need for this logic which is sort of obscured by control flow. Plus the relayer never interprets this data anyways so we don't need much

joeykraut (approved):
Some suggestions that i think would make the implementation cleaner, but overall nice work!

PR #438, opened 2024-02-29
circuits: benches: `VALID RELAYER FEE SETTLEMENT`: Add benchmarks

joeykraut:
### Purpose
This PR adds benchmarks for the `VALID RELAYER FEE SETTLEMENT` circuit.

I also had to increase the size of the SRS now that we are proving larger circuits (fee settlement). This includes using git LFS to track the `srs.ptau` file as it is now too large to be tracked normally.

### Testing
- All tests pass
- Benchmarks run

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+srs/srs.ptau filter=lfs diff=lfs merge=lfs -text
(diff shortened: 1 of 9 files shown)


akirillo (approved):
just a couple indexing fixes, lgtm

PR #439, opened 2024-02-29
arbitrum-client, workers: task-driver: use aux transfer data in integration tests

akirillo:
This PR adds helpers for constructing auxiliary transfer data (ported over from the contracts repo) and updates the integration tests for the Arbitrum client and the task driver pertaining to wallet updates.

All existing Arbitrum client integration tests pass, but I have held off on running the task driver integration tests since the task driver still requires more intervention for it to compile, and it seems @joeykraut is already planning on handling that.


test-helpers/src/transfer_aux_data/mod.rs
```diff
@@ -0,0 +1,138 @@
...
+    let transfer = to_contract_external_transfer(transfer)?;
+
+    let res = if transfer.is_withdrawal {
+        let transfer_bytes = serialize_calldata(&transfer)?;
+        let transfer_hash = H256::from_slice(keccak256(&transfer_bytes).as_slice());
+        let transfer_signature = wallet.sign_hash(transfer_hash)?.to_vec();
+
+        TransferAuxData::Withdrawal(WithdrawalAuxData {
+            external_transfer_signature: transfer_signature,
+        })
```
joeykraut, line 44: still not sure why we have an enum for this? As per my last review it just adds control flow everywhere you want to interact with the struct. Why not just have `TransferAuxData` be a struct and set the deposit fields to `None` on a withdrawal and vice versa? Like how we do it in the contracts

joeykraut (approved):
Let's discuss these abstractions that didn't get discussed in the last PR before merging

test-helpers/src/transfer_aux_data/mod.rs
```diff
@@ -0,0 +1,138 @@
...
+    digest_input[1] = 0x01;
+    digest_input[2..34].copy_from_slice(&domain_separator[..]);
+    digest_input[34..66].copy_from_slice(&struct_hash[..]);
+    keccak256(digest_input)
+}
+
+/// Convert a `U256` to a `BigUint`
+fn u256_to_biguint(u: U256) -> BigUint {
+    BigUint::from_bytes_le(u.as_le_slice())
+}
```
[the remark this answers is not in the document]
akirillo: Yup I saw this, they are indeed different types

test-helpers/src/transfer_aux_data/mod.rs
akirillo: Ah, I thought this was ack'd on the last PR - I erred on the side of an enum so that the caller can be sure of exactly what aux data to find properly set, i.e. we don't have to scatter `unwrap`s or `Result`s throughout assuming that all the right fields on the aux data struct are set, this is now enforced by the type system.

I am not horribly opinionated, taking another look at it now, we could have `TransferAuxData` have two fields, an `Option<DepositAuxData>` and an `Option<WithdrawalAuxData>`, which trades off the enum control flow for one `unwrap` / `Result` when accessing the fields. How does that sound?

PR #443, opened 2024-02-29
arbitrum-client: `settle_online_relayer_fee`

akirillo:
This PR adds a method to the Arbitrum client for calling the `settle_online_relayer_fee` method on the Darkpool. This includes the necessary conversion & event parsing methods.

I've also added an integration test asserting that the method does not revert, and that the correct public shares can be recovered from calldata.

All `arbitrum-client` integration tests pass.

### DIFF
--- a/arbitrum-client/src/helpers.rs
+++ b/arbitrum-client/src/helpers.rs
@@ -5,6 +5,7 @@ use circuit_types::{traits::BaseType, SizedWalletShare};
 use constants::Scalar;
 use contracts_common::types::{
     ValidMatchSettleStatement as ContractValidMatchSettleStatement,
+    ValidRelayerFeeSettlementStatement as ContractValidRelayerFeeSettlementStatement,
     ValidWalletCreateStatement as ContractValidWalletCreateStatement,
     ValidWalletUpdateStatement as ContractValidWalletUpdateStatement,
 };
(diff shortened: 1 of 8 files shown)


arbitrum-client/src/helpers.rs
```diff
@@ -114,3 +115,44 @@ pub fn parse_shares_from_process_match_settle(
...
+        let mut shares = valid_relayer_fee_settlement_statement
+            .recipient_updated_public_shares
+            .into_iter()
+            .map(Scalar::new);
+
+        Ok(SizedWalletShare::from_scalars(&mut shares))
+    } else {
+        Err(ArbitrumClientError::BlinderNotFound)
+    }
+}
```
joeykraut, line 158: Could clean this up as:
```suggestion
    let sender_shares = &valid_relayer_fee_settlement_statement.sender_updated_public_shares;
    let sender_public_blinder_share = sender_shares.last().unwrap();

    // The blinder is expected to be the last public wallet share
    let recipient_shares = &valid_relayer_fee_settlement_statement.recipient_updated_public_shares
    let recipient_public_blinder_share = recipient_shares.last().unwrap();

    let target_share = public_blinder_share.inner();
    let selected_shares = if sender_public_blinder_share == &target_share {
        sender_shares
    } else if recipient_public_blinder_share == &target_share {
        recipient_shares
    } else {
        return Err(ArbitrumClientError::BlinderNotFound);
    }
    
    let mut share_iter = selected_shares.into_iter().map(Scalar::new);
    Ok(SizedWalletShare::from_scalars(&mut shares))
}
```

joeykraut (approved):

PR #450, opened 2024-03-01
task-driver: integration: Update integration tests with fees changes

joeykraut:
### Purpose
This PR fixes up the integration tests for the task driver given the fees changes.

I'm dealing with a serialization bug related to https://github.com/serde-rs/serde/issues/1682 so the tests do not currently pass. Will fix them up in the next follow up.

### Testing
- Unit tests pass

### DIFF
--- a/test-helpers/Cargo.toml
+++ b/test-helpers/Cargo.toml
@@ -29,7 +29,6 @@ ethers = "2"
 arbitrum-client = { path = "../arbitrum-client" }
 circuit-types = { path = "../circuit-types" }
 common = { path = "../common", features = ["mocks"] }
-contracts-common = { git = "https://github.com/renegade-fi/renegade-contracts.git" }
 constants = { path = "../constants" }
 renegade-crypto = { path = "../renegade-crypto" }
 
--- a/test-helpers/Cargo.toml
+++ b/test-helpers/Cargo.toml
@@ -29,6 +29,7 @@ ethers = "2"
 arbitrum-client = { path = "../arbitrum-client" }
 circuit-types = { path = "../circuit-types" }
 common = { path = "../common", features = ["mocks"] }
+contracts-common = { git = "https://github.com/renegade-fi/renegade-contracts.git" }
 constants = { path = "../constants" }
 renegade-crypto = { path = "../renegade-crypto" }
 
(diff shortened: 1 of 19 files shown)


workers/task-driver/integration/tests/update_wallet.rs
```diff
@@ -362,110 +369,29 @@ integration_test_async!(test_update_wallet__remove_fee);
 // | Balances |
 // ------------
 
-/// Tests updating a wallet by depositing into the pool
-#[allow(non_snake_case)]
-async fn test_update_wallet__deposit_and_withdraw(test_args: IntegrationTestArgs) -> Result<()> {
```
joeykraut, line 367: duplicate

test-helpers/src/contract_interaction/transfer_auth/mod.rs
```diff
@@ -33,56 +30,60 @@ pub fn gen_transfer_with_auth(
...
+        ExternalTransferDirection::Withdrawal => gen_withdrawal_with_auth(wallet, transfer),
+    }
+}
+
+/// Generate a withdrawal payload with proper auth data
+fn gen_withdrawal_with_auth(
+    wallet: &Wallet<SigningKey>,
+    transfer: ExternalTransfer,
+) -> Result<ExternalTransferWithAuth> {
+    let transfer_bytes = serialize_calldata(&transfer)?;
```
akirillo, line 48: this will not form the correct signature. notice how before, we would convert to a `ContractExternalTransfer` via `to_contract_external_transfer`. we need to do this in the withdrawal case as it is the serialization of that struct that the contracts expect a signature over.

test-helpers/Cargo.toml
```diff
@@ -29,6 +29,7 @@ ethers = "2"
 arbitrum-client = { path = "../arbitrum-client" }
 circuit-types = { path = "../circuit-types" }
 common = { path = "../common", features = ["mocks"] }
+contracts-common = { git = "https://github.com/renegade-fi/renegade-contracts.git" }
```
akirillo, line 32: do we actually need this dependency in the crate? doesn't seem like we use it.

akirillo (approved):
one necessary change, but doesn't require discussion so approving. you made the code i wrote much cleaner lol

PR #451, opened 2024-03-01
state: Replace flexbuffers serialization with bincode

joeykraut:
### Purpose
This PR changes the `state/storage` serialization format from `flexbuffer` -> `bincode`. Bincode supports `u128` and `flexbuffers` does not; and we now use `u128` to represent balance, transfer, order amounts etc.

Also fixed a few bugs with queues that were discovered along the way.

### Testing
- All unit tests pass in `state`

### DIFF
--- a/common/src/types/tasks.rs
+++ b/common/src/types/tasks.rs
@@ -44,7 +44,6 @@ pub struct QueuedTask {
 /// TODO: We can add completed and failed states if/when we implement a task
 /// history
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
-#[serde(tag = "status")]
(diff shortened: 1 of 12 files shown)


common/src/types/tasks.rs
```diff
@@ -44,7 +44,6 @@ pub struct QueuedTask {
 /// TODO: We can add completed and failed states if/when we implement a task
 /// history
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
-#[serde(tag = "status")]
```
joeykraut, line 47: `bincode` doesn't support tagged enums, the tag was for the FE, but we're planning to refactor FE task state information anyways.

PR #452, opened 2024-03-01
config: add tracing/logging config options

akirillo:
This PR removes the usage of compiler flags and environment variables for configuring telemetry with CLI / config file options. This centralizes relayer configuration and makes it more straightforward. I tested these changes ad-hoc by running the relayer locally and setting the values appropriately in a config file, confirming that traces / logs show up in Datadog as expected.

### DIFF
--- a/config/src/lib.rs
+++ b/config/src/lib.rs
@@ -153,6 +153,27 @@ struct Cli {
     /// should manage
     #[clap(short, long, value_parser)]
     pub wallet_file: Option<String>,
+
+    // -------------
+    // | Telemetry |
+    // -------------
+
+    /// Whether or not to enable OTLP tracing
+    #[clap(long = "otlp", value_parser)]
+    pub otlp_enabled: bool,
+
+    /// The deployment environment w/ which
+    /// to tag OTLP traces
+    #[clap(long, value_parser)]
+    pub otlp_env: Option<String>,
+
+    /// The OTLP collector endpoint to send traces to
+    #[clap(long, value_parser, default_value = "http://localhost:4317")]
+    pub otlp_collector_url: String,
+
+    /// Whether or not to enable Datadog-formatted logs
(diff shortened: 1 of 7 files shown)


config/src/lib.rs
```diff
@@ -153,6 +153,27 @@ struct Cli {
...
     #[clap(short, long, value_parser)]
     pub wallet_file: Option<String>,
+
+    // -------------
+    // | Telemetry |
+    // -------------
+
+    /// Whether or not to enable OTLP tracing
+    #[clap(long, value_parser)]
+    pub otlp_enabled: bool,
```
joeykraut, line 163: do we need this, can one of the others below be used to determine whether OTLP is enabled? The fewer flags the better

joeykraut: Also can we set `long="otlp"` here to give the switch `--otlp`

config/src/lib.rs
```diff
@@ -153,6 +153,27 @@ struct Cli {
...
+    #[clap(long, value_parser)]
+    pub otlp_env: Option<String>,
+
+    /// The OTLP collector endpoint to send traces to
+    #[clap(long, value_parser, default_value = "http://localhost:4317")]
+    pub otlp_collector_url: String,
+
+    /// Whether or not to enable Datadog-formatted logs
+    #[clap(long, value_parser)]
+    pub datadog_enabled: bool,
```
joeykraut, line 176: same here, can we set the switch `--datadog`?

joeykraut (approved):

config/src/lib.rs
akirillo: i was considering this, but i think it's a better experience to 1) have a flag that clearly defines its intent, and 2) follow conventions in the expected manner, in this case by using the same default collector endpoint as is expected in OTLP generally (which also means the flag can be omitted if you're using defaults)

But ack on changing the `long` name

PR #453, opened 2024-03-02
util: telemetry: expose telemetry builder & add metrics

akirillo:
This PR refactors the `telemetry` module to use a builder pattern so that configuration for the various telemetry services is more discoverable and easily configurable as we add functionality.

We use this pattern to introduce a metrics layer that exports metrics to a StatsD endpoint, and additionally leverages the [`metrics_tracing_context`](https://docs.rs/metrics-tracing-context/latest/metrics_tracing_context/) crate to inject span fields as metric labels.

I've tested this ad-hoc by running the relayer with the desired config options + a local Datadog agent and have confirmed that metrics are recorded.

### DIFF
--- a/config/src/lib.rs
+++ b/config/src/lib.rs
@@ -159,21 +159,28 @@ struct Cli {
     // -------------
 
     /// Whether or not to enable OTLP tracing
-    #[clap(long = "otlp", value_parser)]
+    #[clap(long = "enable-otlp", value_parser)]
     pub otlp_enabled: bool,
-
     /// The deployment environment w/ which
     /// to tag OTLP traces
     #[clap(long, value_parser)]
     pub otlp_env: Option<String>,
-
     /// The OTLP collector endpoint to send traces to
     #[clap(long, value_parser, default_value = "http://localhost:4317")]
     pub otlp_collector_url: String,
-
     /// Whether or not to enable Datadog-formatted logs
-    #[clap(long = "datadog", value_parser)]
+    #[clap(long = "enable-datadog", value_parser)]
     pub datadog_enabled: bool,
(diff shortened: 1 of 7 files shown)


config/src/lib.rs
```diff
@@ -268,6 +280,15 @@ pub struct RelayerConfig {
...
     pub datadog_enabled: bool,
+
+    /// Whether or not to enable metrics collection
+    pub metrics_enabled: bool,
+
+    /// The StatsD recorder host to send metrics to
+    pub statsd_host: String,
+
+    /// The StatsD recorder port to send metrics to
+    pub statsd_port: u16,
```
joeykraut, line 291: nit: maybe less white space, following the rest of the struct? I understand this to be a stylistic difference, but I'd suggest maintaining convention when one is established

joeykraut (approved):

config/src/lib.rs
akirillo: oh ya i agree, i think maybe my formatter config did this

config/src/lib.rs
```diff
@@ -174,6 +174,18 @@ struct Cli {
     /// Whether or not to enable Datadog-formatted logs
     #[clap(long, value_parser)]
     pub datadog_enabled: bool,
+
+    /// Whether or not to enable metrics collection
+    #[clap(long, value_parser)]
+    pub metrics_enabled: bool,
```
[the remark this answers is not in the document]
akirillo: hmm, in the previous pr, we just used `otlp` and `datadog`. i would say i agree with this comment more, though, that switches should be verbs - so i'll make this change and also change to `enable-otlp` and `enable-datadog`

PR #459, opened 2024-03-03
workers: add basic metrics

akirillo:
This PR adds metrics collection for the following:
- \# local peers
- \# remote peers
- \# new wallets
- \# / volume of deposits
- \# / volume of withdrawals
- volume of trades

Notably, this required:
1. Implementing a method on the `Token` type for calculating a human-readable amount of a token given its `decimals` value. This is because the `metrics` crate can only increment counters/gauges using `u64` or `f64`, so a raw token fixed-point amount could not be recorded. May as well record the human-readable amount directly, however this suffers from loss of precision for amount with more than 53 significant digits.
2. Opening the `MatchResult` in the match-settle flow after the collaborative proof has been verified so that the volume traded could be recorded. This adds a bit of bandwidth to the MPC, but is more straightforward in implementation than inferring the traded volume from the difference in wallet balances.

I tested that metrics are properly recorded by deploying the relayer w/ a metric attached to the ping endpoint - this is not included in this PR.


common/src/lib.rs
```diff
@@ -15,6 +15,7 @@ use tokio::sync::RwLock as TokioRwLock;
 
 pub mod default_wrapper;
 pub mod keyed_list;
+pub mod metrics_helpers;
```
joeykraut, line 18: I'd suggest we make this its own crate. `common` is convenient, but it's likely that this functionality will outgrow the `common` crate, maybe `renegade-metrics`?

common/src/types/token.rs
```diff
@@ -701,4 +704,15 @@ impl Token {
                 )
             })
     }
+
+    /// Gets the amount of the token as an f64, accounting for the associated
+    /// number of decimals.
+    ///
+    /// Note that due to conversion to f64, the result may lose precision.
+    pub fn get_amount_f64(&self, amount: u128) -> Result<f64, String> {
+        let decimals = self.get_decimals().ok_or(DECIMALS_NOT_FOUND_ERR.to_string())?;
```
joeykraut, line 713: I'd suggest we just `unwrap_or_default` here, so that we get zero decimals. That way this method can be infallible

common/src/types/token.rs
```diff
@@ -701,4 +704,15 @@ impl Token {
                 )
             })
     }
+
+    /// Gets the amount of the token as an f64, accounting for the associated
+    /// number of decimals.
+    ///
+    /// Note that due to conversion to f64, the result may lose precision.
+    pub fn get_amount_f64(&self, amount: u128) -> Result<f64, String> {
```
joeykraut, line 712: nit: I'd suggest renaming to `apply_decimal_adjustment`, `convert_decimals`, or similar. `get_amount` seems to imply a getter of a property of the token, but this method is really about applying a transformation

workers/gossip-server/src/peer_discovery/peers.rs
```diff
@@ -15,6 +15,13 @@ use crate::{
     server::GossipProtocolExecutor,
 };
 
+/// Metric describing the number of local peers the relayer
+/// is connected to
+const NUM_LOCAL_PEERS_METRIC: &str = "num_local_peers";
+/// Metric describing the number of remote peers the relayer
+/// is connected to
+const NUM_REMOTE_PEERS_METRIC: &str = "num_remote_peers";
```
joeykraut, line 23: why do we centralize metrics names for other metrics, but define them locally in this case? I feel we should do one or the other. And if we make `renegade-metrics` its own crate, it'd probably make sense to centralize them there

common/src/metrics_helpers.rs
```diff
@@ -0,0 +1,82 @@
...
+/// lossy f64 conversion via the associated number of decimals
+fn get_asset_and_volume(mint: &BigUint, amount: u128) -> Result<(String, f64), String> {
+    let token = Token::from_addr_biguint(mint);
+    let asset = token.get_ticker().unwrap_or(&biguint_to_hex_string(mint)).to_string();
+    let volume = token.get_amount_f64(amount)?;
+
+    Ok((asset, volume))
+}
+
+/// Record the volume of an asset movement (e.g. deposit, withdrawal, trade)
```
joeykraut, line 38: nit: wording here feels a bit weird

```suggestion
/// Record a volume metric (e.g. deposit, withdrawal, trade)
```

joeykraut: Also we can make the method it calls infallible via the suggestion above, so we can rename this method

workers/gossip-server/src/peer_discovery/peers.rs
```diff
@@ -102,11 +109,19 @@ impl GossipProtocolExecutor {
                 .collect_vec()
         }; // locked_expiry_cache released
 
+        let my_cluster_id = self.global_state.get_cluster_id()?;
+        let num_local_peers =
+            filtered_peers.iter().filter(|peer| peer.cluster_id == my_cluster_id).count();
+        let num_remote_peers = filtered_peers.len() - num_local_peers;
+
```
joeykraut, line 116: nit: move this above the location it's used, no reason to put unrelated lines between

workers/gossip-server/src/peer_discovery/peers.rs
```diff
@@ -102,11 +109,19 @@ impl GossipProtocolExecutor {
...
         // Add all filtered peers to the network manager's address table
         self.add_new_addrs(&filtered_peers)?;
         // Add all filtered peers to the global peer index
         self.global_state.add_peer_batch(filtered_peers)?;
 
+        metrics::gauge!(NUM_LOCAL_PEERS_METRIC).increment(num_local_peers as f64);
+        metrics::gauge!(NUM_REMOTE_PEERS_METRIC).increment(num_remote_peers as f64);
```
joeykraut, line 123: it looks like we never decrement this metric? We need to do so when peers expire o/w this metric only grows

workers/task-driver/src/tasks/create_new_wallet.rs
```diff
@@ -34,6 +34,9 @@ use crate::helpers::find_merkle_path;
 /// The task name to display when logging
 const NEW_WALLET_TASK_NAME: &str = "create-new-wallet";
 
+/// Metric describing the number of new wallets created
+const NUM_NEW_WALLETS_METRIC: &str = "num_new_wallets";
```
joeykraut, line 38: same here, why not centralize this definition?

joeykraut (approved):

workers/gossip-server/src/peer_discovery/peers.rs
akirillo: woah, strange - i definitely added a decrement at expiry, I must've lost in when rebasing or something. thank you for the call out!

workers/gossip-server/src/peer_discovery/peers.rs
akirillo: ah, this was to avoid cloning `filtered_peers` since it is moved in the call to `self.global_state.add_peer_batch`. it's likely an inconsequential clone (and maybe the compiler is smart enough to avoid it) so i'll just invoke it and move these lines

PR #460, opened 2024-03-04
task-driver: tasks: pay-fees: Define initial `pay-fees` task

joeykraut:
### Purpose
This PR defines the skeleton of the `PayFees` task, which will settle the relayer and protocol fee for a given balance in a given wallet. Implementation is deferred to a follow-up

### Testing
- Unit tests pass

### DIFF
--- a/workers/task-driver/src/tasks/pay_fees.rs
+++ b/workers/task-driver/src/tasks/pay_fees.rs
@@ -1,0 +1,195 @@
+//! The `PayFees` task is responsible for settling the fees due for a given
+//! wallet
+
+// TODO: Remove this lint allowance
+#![allow(unused)]
+
+use std::{
+    error::Error,
+    fmt::{Display, Formatter, Result as FmtResult},
+};
+
+use arbitrum_client::client::ArbitrumClient;
+use async_trait::async_trait;
+use common::types::{tasks::PayFeesTaskDescriptor, wallet::WalletIdentifier};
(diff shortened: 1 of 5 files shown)


workers/task-driver/src/tasks/pay_fees.rs
```diff
@@ -0,0 +1,195 @@
...
+    SubmittingProtocolPayment,
+    /// The task is updating the validity proofs for the wallet
+    UpdatingValidityProofs,
+    /// The task has finished
+    Completed,
+}
+
+impl TaskState for PayFeesTaskState {
+    fn commit_point() -> Self {
+        PayFeesTaskState::ProvingRelayerPayment
```
akirillo, line 54: semantically, shouldn't this be either `SubmittingPayment` or `SubmittingProtocolPayment`? and since either of the variants could be the commit point, it seems to justify either splitting this out into two tasks, or having the proving/submitting state be agnostic of whether this is a relayer or protocol fee payment (and have that be captured in the task descriptor)

akirillo (approved):
this looks mostly like scaffolding code so i'll approve, but unsure about the way the commit point is selected

workers/task-driver/src/tasks/pay_fees.rs
joeykraut: Oops yeah, wrote this down before I defined all the states.

It's most convenient to make it one task per balance. This is the granularity at which fees are generated, so it makes caller logic straightforward.

PR #464, opened 2024-03-05
docker: sequencer: use custom devnet chain info

akirillo:
This PR adds a custom chain configuration for the sequencer, allowing us to use the dedicated devnet chain ID defined in https://github.com/renegade-fi/renegade-contracts/pull/233.

This chain config is a copy of the one used for the [Arbitrum Stylus devnet](https://github.com/OffchainLabs/stylus/blob/stylus/cmd/chaininfo/arbitrum_chain_info.json#L148-L182), but with our own chain ID and chain name.

I ran the `arbitrum-client` integration tests with the new sequencer (against the `andrew/no-verify-guards` branch where these changes are reflected contract-side) and they pass, indicating that the verification guards are satisfied appropriately in the relayer integration tests and the chain IDs line up.

### DIFF
--- a/docker/sequencer/Dockerfile
+++ b/docker/sequencer/Dockerfile
@@ -53,6 +53,9 @@ RUN RUSTFLAGS="-Clink-arg=-zstack-size=131072 -Zlocation-detail=none -C opt-leve
     -Z build-std=std,panic_abort \
     -Z build-std-features=panic_immediate_abort
 
+# Copy over devnet chain info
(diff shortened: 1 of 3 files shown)


joeykraut (approved):
Nice work

I believe we have, but double checking: have we checked that the no verify flag fails on a different chain id?

akirillo:
> Nice work
> 
> I believe we have, but double checking: have we checked that the no verify flag fails on a different chain id?

Yes we have

PR #468, opened 2024-03-07
task-driver: pay-relayer-fee: Implement task

joeykraut:
### Purpose
This PR implements the `pay-relayer-fee` task which handles redeeming a fee from a user's wallet into the locally managed relayer's wallet.

Testing is deferred to a follow-up

### Testing
- Unit tests pass


common/src/types/wallet/derivation.rs
```diff
@@ -0,0 +1,204 @@
+//! Helpers for deriving wallet keychain types
```
joeykraut, line 1: renamed this file, didn't change the contents

config/src/lib.rs
```diff
@@ -552,6 +543,17 @@ fn set_contract_from_file(config: &mut RelayerConfig, file: Option<String>) -> R
     Ok(())
 }
 
+/// Parse the relayer's decryption key from a string
+pub fn parse_decryption_key(key_str: Option<String>) -> Result<DecryptionKey, String> {
+    if let Some(k) = key_str {
+        DecryptionKey::from_hex_str(&k)
+    } else {
+        // Must print here as logger is not yet setup
+        println!("{}\n", "WARN: No fee decryption key provided, generating one".yellow());
```
akirillo, line 552: I see why it's hard to ensure that logging is set up before we call this function, but on the note of using `colored` here - in Datadog, this will just produce wonky text encodings of the color being used. Not really an issue since we can always search for this log, but FYI

state/src/storage/tx/node_metadata.rs
```diff
@@ -22,6 +26,11 @@ const CLUSTER_ID_KEY: &str = "cluster-id";
 const NODE_KEYPAIR_KEY: &str = "node-keypair";
 /// The name of the local peer addr in the node metadata table
 const LOCAL_ADDR_KEY: &str = "local-addr";
+/// The key for the local relayer's wallet ID in the node metadata table
+const LOCAL_WALLET_ID_KEY: &str = "local-wallet-id";
+/// The key for the local relayer's fee decryption key in the node metadata
+/// table
+const LOCAL_RELAYER_FEE_KEY: &str = "local-relayer-fee";
```
akirillo, line 33: tiny nit can we have the key be `"local-relayer-fee-key"`

workers/task-driver/src/tasks/pay_relayer_fee.rs
```diff
@@ -189,23 +239,164 @@ impl Task for PayRelayerFeeTask {
...
+        sender_wallet: &Wallet,
+        recipient_wallet: &Wallet,
+    ) -> Result<(Wallet, Wallet), PayRelayerFeeTaskError> {
+        let mut new_sender_wallet = sender_wallet.clone();
+        let mut new_recipient_wallet = recipient_wallet.clone();
+
+        let balance = new_sender_wallet
             .get_balance_mut(mint)
             .ok_or_else(|| PayRelayerFeeTaskError::State(ERR_BALANCE_MISSING.to_string()))?;
+        let mint = balance.mint.clone();
```
akirillo, line 387: nit: I don't think this line is necessary, we already have `mint` from the args

akirillo (approved):

circuit-types/src/elgamal.rs
```diff
@@ -64,6 +77,33 @@ impl DecryptionKey {
...
 
+impl Serialize for DecryptionKey {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let mut bytes = Vec::new();
+        self.serialize_compressed(&mut bytes).unwrap();
+
+        // Serialize explicitly as a sequence to avoid issues with serde formats
```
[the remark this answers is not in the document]
joeykraut: Unlikely, haven't checked

config/src/lib.rs
```diff
@@ -152,10 +153,9 @@ struct Cli {
     /// Defaults to the devnet pre-funded key
     #[clap(long = "arbitrum-pkey", value_parser, default_value = "0xb6b15c8cb491557369f3c7d2c287b053eb229daa9c22138887752191c9520659")]
     pub arbitrum_private_key: String,
-    /// A file holding a json representation of the wallets the local node
-    /// should manage
-    #[clap(short, long, value_parser)]
-    pub wallet_file: Option<String>,
```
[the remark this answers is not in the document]
joeykraut: This was just old, not really subsumed in any way, we now have a lookup wallet task for this instead.

state/src/storage/tx/node_metadata.rs
joeykraut: did this in the follow up!

PR #469, opened 2024-03-07
task-driver: integration: pay-relayer-fee: Impl integration tests

joeykraut:
### Purpose
This PR adds integration tests for the `PayRelayerFee` task in the `task-driver`. These test both the case in which the relayer wallet holds a zero and non-zero balance of the fee mint. 

### Testing
- All unit tests pass crate wide
- All `task-driver` integration tests pass


common/src/types/tasks.rs
```diff
@@ -376,8 +376,8 @@ pub struct PayProtocolFeeTaskDescriptor {
 
 impl PayProtocolFeeTaskDescriptor {
     /// Constructor
-    pub fn new(wallet_id: WalletIdentifier, balance_mint: BigUint) -> Self {
-        PayProtocolFeeTaskDescriptor { wallet_id, balance_mint }
+    pub fn new(wallet_id: WalletIdentifier, balance_mint: BigUint) -> Result<Self, String> {
```
akirillo, line 379: note to self: why do we need this?

workers/task-driver/integration/main.rs
```diff
@@ -181,6 +182,9 @@ fn setup_arbitrum_client_mock(test_args: &CliArgs) -> ArbitrumClient {
 
 /// Setup code for the integration tests
 fn setup_integration_tests(test_args: &CliArgs) {
+    // Setup token remaps, this is used in metrics collection, will panic if not set
+    TOKEN_REMAPS.set(Default::default()).unwrap();
```
akirillo, line 186: hmm, should we gate metrics collection behind a cargo feature? the current mentality is to defer all configuration of the node to the config file, but this seems annoying (and maybe other users will want to completely disable for perf reasons)

workers/task-driver/integration/tests/pay_relayer_fee.rs
```diff
@@ -0,0 +1,189 @@
...
+        relayer_share_seed,
+        &test_args,
+    )
+    .await
+}
+integration_test_async!(test_pay_relayer_fee__zero_initial_balance);
+
+/// Tests paying a relayer fee when the relayer initially has a non-zero balance
+#[allow(non_snake_case)]
+async fn test_pay_relayer_fee__non_zero_initial_balance(
```
akirillo, line 140: seems like the only difference between the implementation of this test and the one above is the initial relayer wallet, perhaps we could absorb some of the common functionality into helpers?

akirillo (approved):

workers/task-driver/integration/main.rs
joeykraut: I think this is fine for now, no need to complicate things

PR #471, opened 2024-03-08
config, util: telemetry: use env vars to configure ddog tags

akirillo:
This PR configures [Unified Service Tagging](https://docs.datadoghq.com/getting_started/tagging/unified_service_tagging) for the telemetry options when `datadog_enabled` is true. It does so by reading the associated values (`service`, `env`, `version`) from the canonical environment variables used by DataDog. This obviates the need for the `otlp_env` config option.

Ideally, we wouldn't need to do this - the DataDog agent _should_ be able to automatically tag all incoming telemetry resources in a way that is configured outside of application-level code (by setting [Autodiscovery](https://docs.datadoghq.com/getting_started/containers/autodiscovery/) labels on the relayer container), but this has not been working with our infrastructure setup and I have an ongoing support thread about it. However, they are automatically applied to logs, as expected. Yay.

In the meanwhile, we will break the "only configure via the config file" pattern only for this case and set the Unified Service Tags manually.

I have tested that this works by running a relayer and datadog agent locally and ensuring that the tags are consistent across logs, metrics, and traces then. I have also pushed these changes to staging to test that tagging is applied correctly for the deployed relayers as well.


util/src/telemetry/datadog/mod.rs
```diff
@@ -0,0 +1,44 @@
...
+pub mod formatter;
+
+/// The name of the environment variable used to set the
+/// service name for DataDog unified service tagging
+pub const DD_SERVICE_ENV_VAR: &str = "DD_SERVICE";
+/// The name of the environment variable used to set the
+/// environment for DataDog unified service tagging
+pub const DD_ENV_ENV_VAR: &str = "DD_ENV";
+/// The current relayer version, used to tag telemetry resources
+pub const VERSION: &str = env!("CARGO_PKG_VERSION");
```
akirillo, line 16: @joeykraut thoughts on using this to replace the `version` cli / config file option?

joeykraut: sure

joeykraut (approved):

PR #472, opened 2024-03-08
task-driver: settle-internal-match, settle-match: Automatically pay fees after settlement

joeykraut:
### Purpose
This PR changes the implementation of the `settle-match` and `settle-internal-match` tasks to automatically pay fees after match settlement. This happens at the end of the task when the state updates.

### Todo
- Switch relayer fee settlement to use the offline fee path. This way we can rely on the task queue to ensure that relayer settlement does not mange the relayer's wallet. This requires the task to be implemented.

### Testing
- Unit tests pass
- Integration tests pass
- Added fee settlement test cases to the integration tests


workers/task-driver/integration/tests/settle_match.rs
```diff
@@ -42,9 +48,11 @@ use uuid::Uuid;
 /// The price at which the mock trade executes at
 const EXECUTION_PRICE: f64 = 9.6;
 /// The amounts that each order is for
-const BUY_ORDER_AMOUNT: u128 = 10;
+const BUY_ORDER_AMOUNT: u128 = 100_000;
```
joeykraut, line 51: Switched to an amount that generates a non-zero fee

akirillo (approved):

PR #474, opened 2024-03-08
task-driver: redeem-realyer-fee: Define task skeleton

joeykraut:
### Purpose
This PR implements the skeleton of the `redeem-relayer-fee` task. This task will be used by relayers to settle notes from chain state into their own wallets.

Implementation is deferred to a follow-up

### Testing
- All unit tests pass


common/src/types/tasks.rs
```diff
@@ -422,6 +428,31 @@ impl From<PayRelayerFeeTaskDescriptor> for TaskDescriptor {
...
 }
 
+/// The task descriptor for redeeming a relayer note
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct RedeemRelayerFeeTaskDescriptor {
+    /// The wallet ID of the relayer's wallet
+    ///
+    /// Technically this should be static and not needed here, but we include it
+    /// to allow the descriptor struct to compute its own task queue key
+    pub wallet_id: WalletIdentifier,
```
akirillo, line 438: why can't we use a `static` in computing the task queue key?

workers/task-driver/src/tasks/redeem_relayer_fee.rs
```diff
@@ -0,0 +1,164 @@
...
+
+/// Defines the redeem relayer fee task
+pub struct RedeemRelayerFeeTask {
+    /// The note to redeem
+    pub note: Note,
+    /// The proof of `VALID FEE REDEMPTION` used to pay the fee
+    pub proof: Option<FeeRedemptionBundle>,
+    /// The arbitrum client used for submitting transactions
+    pub arbitrum_client: ArbitrumClient,
+    /// A hand to the global state
```
akirillo, line 119:
```suggestion
    /// A handle to the global state
```

akirillo (approved):

common/src/types/tasks.rs
joeykraut: this value isn't static, it may be generated on startup

akirillo: ah, blanked out - we'd have to fetch this from state, right, which is async and doesn't play well with the `get_task_queue_key` interface?

akirillo: i suppose we could put it in a `OnceCell` on startup, but eh, inconsequential

PR #475, opened 2024-03-08
task-driver: redeem-relayer-fee: Implement task logic

joeykraut:
### Purpose
This PR implements the `redeem-realyer-fee` logic. Note that we don't need to update validity proofs because relayer wallets don't currently holds orders. 

We also automatically enqueue this task on the relayer's wallet when a relayer fee is settled offline.

Testing is deferred to a follow-up.

### Testing
- Workspace unit tests pass

### DIFF
--- a/workers/task-driver/src/helpers.rs
+++ b/workers/task-driver/src/helpers.rs
@@ -9,6 +9,7 @@ use circuit_types::{
         compute_wallet_private_share_commitment, create_wallet_shares_from_private, reblind_wallet,
         wallet_from_blinded_shares,
     },
+    note::Note,
     order::Order,
     r#match::OrderSettlementIndices,
     SizedWallet,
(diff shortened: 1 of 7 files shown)


workers/task-driver/src/helpers.rs
```diff
@@ -393,3 +394,12 @@ fn enqueue_protocol_fee_settlement_task(
         PayOfflineFeeTaskDescriptor::new_protocol_fee(wallet_id, mint).expect("infallible");
     state.append_task(descriptor.into()).map_err(|e| e.to_string()).map(|_| ())
 }
+
+/// Enqueue a job to redeem a relayer fee into the relayer's wallet
+pub(crate) fn enqueue_relayer_redeem_job(note: Note, state: &State) -> Result<(), String> {
+    let relayer_wallet_id = state.get_relayer_wallet_id()?.ok_or("relayer wallet not found")?;
```
akirillo, line 400: nit: we can use `ERR_WALLET_MISSING` here

akirillo (approved):

PR #477, opened 2024-03-10
api-server: http: wallet: Fixup shares after overwriting wallet fields

joeykraut:
### Purpose
This PR fixes up the public shares of the wallet after overwriting the `managing_cluster` and `match_fee` with locally configured values in `POST /v0/wallet`.

### Testing
- Unit tests pass
- I am able to create a new wallet through the API

### DIFF
--- a/workers/api-server/src/http/wallet.rs
+++ b/workers/api-server/src/http/wallet.rs
@@ -1,7 +1,10 @@
 //! Groups wallet API handlers and definitions
 
 use async_trait::async_trait;
-use circuit_types::{balance::Balance, order::Order};
+use circuit_types::{
+    balance::Balance, native_helpers::create_wallet_shares_from_private, order::Order,
+    SizedWallet as SizedCircuitWallet,
+};
 use common::types::{
     tasks::{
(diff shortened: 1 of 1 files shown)


workers/api-server/src/http/wallet.rs
```diff
@@ -171,6 +174,16 @@ impl TypedHandler for CreateWalletHandler {
         // and create proofs of validity
         let wallet_id = req.wallet.id;
         let mut wallet: Wallet = req.wallet.try_into().map_err(|e: String| bad_request(e))?;
+
+        // Modify the public of the new wallet to reflect the overwritten fields
```
sehyunc, line 178: Should be: `// Modify the public shares of the new wallet to reflect the overwritten fields`

sehyunc (changes requested):
Small change

workers/api-server/src/http/wallet.rs
joeykraut: yes, fixed

sehyunc (approved):

PR #480, opened 2024-03-15
workers: price-reporter: expose constants for retry logic

akirillo:
This PR makes the constants used in retrying exchange connections public so that the price reporter may use them. Additionally, I fix a small bug in which the retry delay was being interpreted as seconds, not milliseconds.


joeykraut:
I don't see the point in exposing these, why not just redefine them in the price reporter server. To me this is a very leaky interface -- implementation details like "how long do I sleep before retrying" don't make sense on the public interface.

PR #483, opened 2024-03-18
core: setup: Set `local-wallet-id` metadata entry at startup

joeykraut:
### Purpose
This PR fixed a bug wherein the relayer's wallet ID is never setup in the node metadata table. This prevents the relayer from redeeming fees after a match.

### Testing
- Unit tests pass
- Integration tests pass
- @sehyunc Please check that your flows work with this fix


akirillo (approved):

sehyunc (approved):
This fixes the `cannot find local-wallet-id` issue.

PR #484, opened 2024-03-18
workers: price-reporter: canonical (Binance) price reports

akirillo:
This PR scopes down the price reporter worker to just reporting a "canonical" price for a given pair, and replaces the reporting of a median price to the reporting of the Binance price.

This abstracts away the exact method by which the relayer's view of a pair's price is derived (e.g. median vs Binance), and removes any reporting of 1) the median price, and 2) exchange-specific prices. The latter can be removed because it was only used to stream prices to the frontend, which will now stream directly from the external price reporter. I have left the associated HTTP/WS endpoints to ease the integration effort on the frontend, with TODO comments to remove the endpoints.

There were some slight changes to the logic by which we determine a canonical (Binance) price to be valid, namely when we construct a median price against which to check deviation, we also filter out stale prices from exchanges.

All unit tests pass.

In the following PR, I will establish a separate worker that reports the canonical price by streaming it from an external price reporter.


external-api/src/http/mod.rs
```diff
@@ -4,7 +4,7 @@ use serde::{Deserialize, Serialize};
 
 pub mod network;
 pub mod order_book;
-pub mod price_report;
+pub mod price_report; // TODO: REMOVE AFTER FE INTEGRATION W/ STANDALONE PRICE REPORTER
```
joeykraut, line 7: are we removing this? Re: our discussion this morning, I though we plan to leave it as an option (though with the new paired down interface)? I'll say it's helpful for debugging occasionally

workers/price-reporter/src/reporter.rs
```diff
@@ -247,73 +243,73 @@ impl Reporter {
...
-        // Collect all non-zero PriceReports and ensure that we have enough.
-        let (non_zero_prices, timestamps): (Vec<Price>, Vec<u64>) = ALL_EXCHANGES
-            .iter()
-            .filter_map(|exchange| self.exchange_info.read_price(exchange))
-            .filter(|(price, _)| *price != Price::default() && price.is_finite())
-            .unzip();
+        // Fetch the most recent Binance price
+        match self.exchange_info.read_price(&Exchange::Binance) {
+            None => PriceReporterState::NotEnoughDataReported(0),
+            Some((price, ts)) => {
```
joeykraut, line 267: maybe breakout this arm into a helper?

joeykraut (approved):

external-api/src/http/mod.rs
akirillo: Ah, if it's helpful for debugging, I can leave it.
I was thinking we'd just leave access to the relayer's view of the price via the websocket endpoint that allows subscribing to system bus topics, but I can also leave the HTTP endpoint and just have it return the "canonical" price alone instead of also returning the per-exchange prices

PR #485, opened 2024-03-19
workers: external-price-reporter: scaffold external price reporter worker

akirillo:
This PR adds the necessary scaffolding for a separate worker responsible for driving the connection to the external price reporter service. This worker is spawned instead of the native price reporter worker if the associated config option is set. It will be sharing some types/functionality (e.g. in this PR, the `PriceReporterJob`) as appropriate with the native price reporter.

I will fill in the implementation of the external price reporter worker stemming from its execution loop in the following PR.

### DIFF
--- a/workers/price-reporter/src/worker.rs
+++ b/workers/price-reporter/src/worker.rs
@@ -93,6 +93,7 @@ impl Worker for PriceReporter {
 
     fn start(&mut self) -> Result<(), Self::Error> {
         // Spawn a tokio thread pool to run the manager in
+        // NOTE(@akirillo): Do we need this?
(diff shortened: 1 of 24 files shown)


workers/price-reporter/src/worker.rs
```diff
@@ -93,6 +93,7 @@ impl Worker for PriceReporter {
 
     fn start(&mut self) -> Result<(), Self::Error> {
         // Spawn a tokio thread pool to run the manager in
+        // NOTE(@akirillo): Do we need this?
```
akirillo, line 96: @joeykraut out of scope for this PR, but it appears that we construct the Tokio runtime below and never use it. I'm happy to remove this if I'm not missing anything

PR #486, opened 2024-03-19
arbitrum-client, util: Lookup protocol fee and key from contract

joeykraut:
### Purpose
This PR fetches the protocol's encryption key and take rate from the contract at startup and stores them in a `OnceCell`. To make testing simpler, if the `mocks` feature flag is enabled, the implementation of the `OnceCell` getter methods will fall back to a default.

### Testing
- Unit tests pass
- Ran a relayer, verified that it fetched the fee and encryption key

### DIFF
--- a/util/src/arbitrum.rs
+++ b/util/src/arbitrum.rs
@@ -1,12 +1,9 @@
 //! Utils relating to Starknet interaction
 
-use std::{fs::File, io::Read};
+use std::{fs::File, io::Read, sync::OnceLock};
 
-use circuit_types::elgamal::EncryptionKey;
-use constants::PROTOCOL_ENCRYPTION_KEY;
+use circuit_types::{elgamal::EncryptionKey, fixed_point::FixedPoint};
 use eyre::{eyre, Result};
-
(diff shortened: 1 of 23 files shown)


util/src/arbitrum.rs
```diff
@@ -31,8 +32,42 @@ pub fn parse_addr_from_deployments_file(file_path: &str, contract_key: &str) ->
...
+    let key = PROTOCOL_PUBKEY.get();
+
+    // If the mocks feature is enabled we unwrap to a default
+    #[cfg(feature = "mocks")]
+    {
+        use circuit_types::elgamal::DecryptionKey;
+        use rand::thread_rng;
+
+        let mut rng = thread_rng();
+        *key.unwrap_or(&DecryptionKey::random(&mut rng).public_key())
```
akirillo, line 66: should we `.set` the protocol pubkey so it's a constant pubkey when it's mocked?

akirillo (approved):

util/src/arbitrum.rs
joeykraut: good suggestion

PR #488, opened 2024-03-19
workers: price-reporter: reorganize price reporter modules for external executor

akirillo:
This PR introduces some reorganization of the `manager` module in the price reporter worker ahead of the implementation of the `ExternalPriceReporterExecutor`, which will be used to manage the worker in the case that the relayer opts to connect to an external price reporting service.

This mostly consists of moving the existing `PriceReporterExecutor` into the `native_executor` module, while keeping some utilities which will be shared in the `manager` module.

The next PR will have the implementation of the `ExternalPriceReporterExecutor`.

### DIFF
--- a/workers/price-reporter/src/manager.rs
+++ b/workers/price-reporter/src/manager.rs
@@ -1,20 +1,63 @@
 //! Defines the PriceReporterExecutor, the handler that is responsible
 //! for executing individual PriceReporterJobs.
-use common::default_wrapper::{DefaultOption, DefaultWrapper};
-use common::types::exchange::PriceReporterState;
-use common::types::token::Token;
-use common::types::CancelChannel;
-use common::{new_async_shared, AsyncShared};
-use job_types::price_reporter::{PriceReporterJob, PriceReporterReceiver};
-use std::{collections::HashMap, thread::JoinHandle};
-use tokio::runtime::Runtime;
-use tokio::sync::oneshot::Sender as TokioSender;
-use tracing::{error, info, info_span, warn, Instrument};
-use util::err_str;
 
-use crate::errors::{ExchangeConnectionError, PriceReporterError};
+use std::{
+    collections::HashMap,
+    sync::{
+        atomic::{AtomicU64, Ordering},
+        Arc,
+    },
+    thread::JoinHandle,
+};
 
-use super::{reporter::Reporter, worker::PriceReporterConfig};
+use atomic_float::AtomicF64;
+use common::types::{
+    exchange::{Exchange, PriceReport, PriceReporterState},
+    token::Token,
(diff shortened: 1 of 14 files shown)


workers/price-reporter/src/manager.rs
```diff
@@ -1,20 +1,63 @@
...
+use crate::{errors::PriceReporterError, worker::PriceReporterConfig};
+
+// pub mod external_executor;
+pub mod native_executor;
+
+// -------------
+// | Constants |
+// -------------
+
+/// If Binance has not reported an update within
```
joeykraut, line 31: nit: I'd avoiding wording these types of things as "if binance...", and keep it abstract. While currently we use Binance prices the code is written more extensibly than that and our language should reflect that!

workers/price-reporter/src/manager.rs
```diff
@@ -1,20 +1,63 @@
...
+/// recent price. Note that this threshold cannot be too aggressive, as certain
+/// long-tail asset pairs legitimately do not update that often.
+const MAX_REPORT_AGE_MS: u64 = 20_000; // 20 seconds
+/// If we do not have at least MIN_CONNECTIONS reports, we pause matches until
+/// we have enough reports. This only applies to Named tokens, as Unnamed tokens
+/// simply use UniswapV3.
+const MIN_CONNECTIONS: usize = 1;
+/// If a Binance PriceReport is more than MAX_DEVIATION (as a fraction) away
+/// from the midpoint, then we pause matches until the prices stabilize.
+const MAX_DEVIATION: f64 = 0.02;
```
joeykraut, line 42: 2% is quite high, is this just a dummy value? we'd like to push this lower IMO

workers/price-reporter/src/manager.rs
```diff
@@ -24,174 +67,112 @@ pub struct PriceReporter {
...
+            exchange != &Exchange::Binance
+                && *price != Price::default()
+                && price.is_finite()
+                && !ts_too_stale(*ts).0
+        })
+        .map(|(_, (price, _))| *price)
+        .collect();
+
+    // If we have enough data to create a median, check for deviation against it
+    if non_zero_prices.len() >= MIN_CONNECTIONS {
```
joeykraut, line 157: nit: I'd switch these cases, slightly less control flow:
```rust
if non_zero_prices.len() < MIN_CONNECTIONS {
    return ...
}

// Median checks
```

joeykraut (approved):

workers/price-reporter/src/manager.rs
akirillo: This is a dummy, value, though Chris and I discussed using 1%, so I'll set that here

PR #490, opened 2024-03-20
workers: price-reporter: add external price reporter executor

akirillo:
This PR implements a secondary executor for the price reporter worker which connects to an external price reporter service, subscribing to all price updates over a single websocket connection. It spawns a separate task which manages the websocket connection and forwards incoming/outgoing messages between it and the executor over a pair of channels.

I have tested running the relayer ad-hoc with both the native and external executor, ensuring that it starts smoothly and that websocket subscriptions are established.

### DIFF
--- a/workers/price-reporter/src/manager/external_executor.rs
+++ b/workers/price-reporter/src/manager/external_executor.rs
@@ -44,7 +44,10 @@ use crate::{
     worker::PriceReporterConfig,
 };
 
-use super::{compute_price_reporter_state, AtomicPriceStreamState, PRICE_REPORT_INTERVAL_MS};
+use super::{
+    compute_price_reporter_state, compute_supported_exchanges_for_pair, AtomicPriceStreamState,
+    PRICE_REPORT_INTERVAL_MS,
+};
 
 /// A type alias for the shared state of the price streams
 type SubscriptionStates = AsyncShared<HashMap<(Token, Token), AtomicPriceStreamState>>;
@@ -84,7 +87,7 @@ pub struct ExternalPriceReporterExecutor {
 
 impl ExternalPriceReporterExecutor {
     /// Creates the executor for the PriceReporter worker.
-    pub(super) fn new(
+    pub(crate) fn new(
         job_receiver: PriceReporterReceiver,
         config: PriceReporterConfig,
         cancel_channel: CancelChannel,
@@ -98,7 +101,7 @@ impl ExternalPriceReporterExecutor {
     }
 
     /// The execution loop for the price reporter
-    pub(super) async fn execution_loop(mut self) -> Result<(), PriceReporterError> {
+    pub(crate) async fn execution_loop(mut self) -> Result<(), PriceReporterError> {
         let mut job_receiver = self.job_receiver.take().unwrap();
         let mut cancel_channel = self.cancel_channel.take().unwrap();
 
@@ -205,11 +208,10 @@ impl ExternalPriceReporterExecutor {
             .map_err(err_str!(ExchangeConnectionError::InvalidMessage))?;
         let ts = get_current_time_seconds();
 
-        let mut subscriptions = self.subscription_states.write().await;
-        subscriptions
-            .entry((base_token, quote_token))
-            .or_insert_with(|| AtomicPriceStreamState::new_from_exchanges(ALL_EXCHANGES))
-            .new_price(exchange, price, ts);
+        let subscriptions = self.subscription_states.write().await;
+        // We unwrap here as we should only get price updates for which we've already
+        // mapped a subscription
+        subscriptions.get(&(base_token, quote_token)).unwrap().new_price(exchange, price, ts);
 
         Ok(())
     }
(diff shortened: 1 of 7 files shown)


workers/price-reporter/src/manager/external_executor.rs
```diff
@@ -0,0 +1,467 @@
...
+    fn spawn_ws_handler_loop(
+        &self,
+    ) -> (UnboundedSender<WebsocketMessage>, UnboundedReceiver<PriceMessage>) {
+        let price_reporter_url: Url = self
+            .config
+            .price_reporter_url
+            .clone()
+            .unwrap()
+            .parse()
+            .expect("Invalid price reporter URL");
```
joeykraut, line 156: Maybe we can parse this in the config or something so that we don't introduce a panic here for a misconfigured relayer?

workers/price-reporter/src/manager/external_executor.rs
```diff
@@ -0,0 +1,467 @@
...
+            msg_out_rx,
+            msg_in_tx,
+        ));
+
+        (msg_out_tx, msg_in_rx)
+    }
+
+    /// Spawns the price streamer loop, which periodically publishes price
+    /// reports to the system bus for the given token pair
+    fn spawn_price_streamer_loop(self, base: Token, quote: Token) {
```
joeykraut, line 175: Why not just push to the bus every time we receive a new price from the API? Simpler this way and doesn't require a separate loop? 

Ideally we get prices exactly on the cadence they're delivered from the exchange. We don't want to introduce extra latency via this loop's sleep. We can restructure the subscription states to hold entire price reports, the choice to make them atomic was a premature optimization. Though this second point isn't that important

workers/price-reporter/src/manager/external_executor.rs
```diff
@@ -0,0 +1,467 @@
...
+                    if let PriceReporterState::Nominal(report) =
+                        self.get_state(base.clone(), quote.clone()).await
+                    {
+                        self.config
+                            .system_bus
+                            .publish(topic_name.clone(), SystemBusMessage::PriceReport(report));
+                    }
+                }
+
+                tokio::time::sleep(Duration::from_millis(PRICE_REPORT_INTERVAL_MS)).await;
```
joeykraut, line 190: nit: could be cleaner to construct the duration once at the function start, even if negligable cost it's a solid habit

workers/price-reporter/src/manager/external_executor.rs
joeykraut: Now that I'm looking at the use below we wrap the atomic structure in a lock anyways so we don't save anything by having it atomic

workers/price-reporter/src/manager/external_executor.rs
```diff
@@ -0,0 +1,467 @@
...
+/// re-establishing connections indefinitely in case of failure
+async fn ws_handler_loop(
+    price_reporter_url: Url,
+    subscription_states: SubscriptionStates,
+    mut msg_out_rx: UnboundedReceiver<WebsocketMessage>,
+    msg_in_tx: UnboundedSender<PriceMessage>,
+) -> Result<(), PriceReporterError> {
+    // Outer loop handles retrying the websocket connection to the external price
+    // reporter in case of some failure
+    loop {
```
joeykraut, line 338: I'd break some of this out into helpers, it's 8 nested control flow statements deep at its max

workers/price-reporter/src/worker.rs
```diff
@@ -59,13 +61,20 @@ impl PriceReporterConfig {
...
         let disabled = self.disabled_exchanges.contains(&exchange);
-        let configured = match exchange {
-            Exchange::Coinbase => {
-                self.exchange_conn_config.coinbase_api_key.is_some()
-                    && self.exchange_conn_config.coinbase_api_secret.is_some()
-            },
-            Exchange::UniswapV3 => self.exchange_conn_config.eth_websocket_addr.is_some(),
-            _ => true,
+
+        let configured = if self.price_reporter_url.is_some() {
```
joeykraut, line 65: maybe add a `uses_external_reporter` to the config? Since we use this pattern twice: here and once below. Pretty minor so at your discretion

joeykraut (approved):

workers/price-reporter/src/manager/external_executor.rs
akirillo: yeah fair point, i'll remove the price streaming loop. I'm not sure we can have the subscription states store price reports since we still need to store the latest price from each of the exchanges somewhere to calculate a price report for a pair. I'll make sure that we only acquire a write lock when adding a new token pair to the top-level mapping, so we should still get something out of atomics.

workers/price-reporter/src/manager/external_executor.rs
akirillo: yeah, fair criticism lol

workers/price-reporter/src/worker.rs
akirillo: ah, I thought about it, but I'd like to capture one intent with one config option

PR #493, opened 2024-03-22
common: wallet: orders: Remove orders entirely from `KeyedList`

joeykraut:
### Purpose
This PR updates the way in which we remove orders from a wallet. Specifically, we no longer replace the order with a default but have the order list behave more like an append-centric vector. This makes for a more sensible API client-side.

This comes as a result of circuit changes that allow for more flexibility in updating orders.

### Testing
- Used the changes to create and cancel orders in the `renegade-cli` and verified its functionality
- Unit tests pass

### DIFF
--- a/common/src/types/wallet/mod.rs
+++ b/common/src/types/wallet/mod.rs
@@ -115,32 +115,6 @@ mod test {
         assert_eq!(wallet.orders.index_of(&id2), Some(1));
     }
 
-    /// Tests adding an order that overwrites a default order
-    #[test]
-    fn test_add_order_overwrite() {
-        let mut wallet = mock_empty_wallet();
-
-        // Fill the orders with default orders
-        for _ in 0..MAX_ORDERS {
-            let id = Uuid::new_v4();
-            let order = mock_order();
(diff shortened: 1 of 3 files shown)


sehyunc (approved):
Regarding client-side changes, I'm assuming orders should simply get appended to the end of the array. For removals, I'm assuming the order gets removed and the other orders shift down, e.g. if the array is [O, O, Z, O, O] where Z is a zeroed order and O is an order with an amount and Z gets removed is the resulting array [O, O, null, O, O] or [O, O, O, O, _]?

joeykraut:
> Regarding client-side changes, I'm assuming orders should simply get appended to the end of the array. For removals, I'm assuming the order gets removed and the other orders shift down, e.g. if the array is `[O, O, Z, O, O]` where Z is a zeroed order and O is an order with an amount and Z gets removed is the resulting array `[O, O, null, O, O]` or `[O, O, O, O, _]`?

The ladder. The order vector should behave somewhat like a common sense vector would; i.e. removing an element in the middle shifts over the elements that were to its right. Adding an order appends it _always_, etc.

PR #494, opened 2024-03-22
external-api: task: Add descriptor to `TaskStateResponse`

joeykraut:
### Purpose
This PR changes the API format of the task status to give a cleaner interface to the client. As well, for update wallet tasks we name the tasks by the update they perform.

### Testing
- Unit tests pass
- Tested the create + cancel order flows
- Opened a websocket on one of the tasks, verified the shape of the JSON

### DIFF
--- a/workers/task-driver/src/tasks/create_new_wallet.rs
+++ b/workers/task-driver/src/tasks/create_new_wallet.rs
@@ -70,10 +70,11 @@ impl TaskState for NewWalletTaskState {
 impl Display for NewWalletTaskState {
     fn fmt(&self, f: &mut Formatter<'_>) -> FmtResult {
         match self {
-            NewWalletTaskState::SubmittingTx { .. } => {
-                write!(f, "SubmittingTx")
-            },
-            _ => write!(f, "{self:?}"),
+            NewWalletTaskState::Pending => write!(f, "Pending"),
+            NewWalletTaskState::Proving => write!(f, "Proving"),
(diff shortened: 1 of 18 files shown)


workers/task-driver/src/tasks/create_new_wallet.rs
```diff
@@ -70,10 +70,11 @@ impl TaskState for NewWalletTaskState {
...
     fn fmt(&self, f: &mut Formatter<'_>) -> FmtResult {
         match self {
-            NewWalletTaskState::SubmittingTx { .. } => {
-                write!(f, "SubmittingTx")
-            },
-            _ => write!(f, "{self:?}"),
+            NewWalletTaskState::Pending => write!(f, "Pending"),
+            NewWalletTaskState::Proving => write!(f, "Proving"),
+            NewWalletTaskState::SubmittingTx => write!(f, "Submitting Tx"),
+            NewWalletTaskState::FindingMerkleOpening => write!(f, "Finding Merkle Opening"),
```
sehyunc, line 76: Would be nice to rename `FindingMerkleOpening` to `FindingOpening` for consistency with UpdateWalletTask, but not necessary.

sehyunc (approved):
Looks good to me

workers/task-driver/src/tasks/create_new_wallet.rs
joeykraut: Will do, nice catch

PR #496, opened 2024-03-23
common, external-api: wallet: Do not filter default elements

joeykraut:
### Purpose
This PR makes a few changes to the API and the serialization we impose on it:
- We do not remove default elements from the wallet when delivering a wallet to the API. These, importantly, should remain in place when a wallet is updated (unless they are overridden) so removing them from the API response makes it difficult or impossible for a client to correctly construct a wallet update.
- We serialize and deserialize fixed point values by the decimal string of their `repr` field.
- I also simplified some of the withdrawal logic in the balance API

### Testing
- Unit tests pass
- Tested the API through our CLI

### DIFF
--- a/circuit-types/src/fixed_point.rs
+++ b/circuit-types/src/fixed_point.rs
@@ -5,14 +5,14 @@
 use std::ops::{Add, Mul, Neg, Sub};
 
 use ark_ff::{BigInteger, Field, PrimeField};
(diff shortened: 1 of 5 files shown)


sehyunc (approved):
Tested simple creating new wallet, depositing, withdrawing, and placing orders in frontend.

PR #497, opened 2024-03-26
common: types: token: clean up token statics

akirillo:
This PR accomplishes the following:
1. Pares down any token metadata hardcoded in the relayer to contain only those tokens which are supported by Binance, and are currently on the hosted token remap in Github. This latter condition is temporary, I'll hammer out canonical token list w/ @ChrisBender tomorrow & update hosted token remap accordingly.
2. Simplifies the `ERC20_DATA` static to only contain ticker renaming information across exchanges, deferring the token address & decimals to the remap used by the relayer as the source of truth.

Also, this PR temporarily switches the quote asset for the default pairs from USDC to USDT. This is because many assets do not have a USDC pair on Binance. When stablecoin price conversion is implemented, we can go back to default USDC quotes w/ price conversion beneath the hood. But for now, this allows the relayer to run against the deployed price reporter w/o issue and with effectively identical prices to what will be implemented, unblocking testing of the frontend.

Using this, the relayer successfully returns a price report over http both when using the native and external executor for the price reporter worker. For the external executor, I ran a price reporter locally over VPN that made use of the changes in this PR. I have also deployed a staging price reporter that imports the changes in this PR and tested the relayer locally against it, confirming that it can return price reports.

When this PR is merged, the `andrew/external-price-reporting-clean` branch will be ready for merge with `main`, ahead of the stablecoin price conversion logic.

### DIFF
--- a/workers/handshake-manager/src/manager/price_agreement.rs
+++ b/workers/handshake-manager/src/manager/price_agreement.rs
@@ -22,9 +22,9 @@ const ERR_NO_PRICE_STREAM: &str = "price report not available for token pair";
 // | Quote Tokens |
 // ----------------
 
-/// USDC ticker
+/// USDT ticker
 ///
-/// For now we only stream prices quoted in USDC
+/// For now we only stream prices quoted in USDT
 const USDT_TICKER: &str = "USDT";
 
 // ---------------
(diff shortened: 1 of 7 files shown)


workers/handshake-manager/src/manager/price_agreement.rs
```diff
@@ -25,7 +25,7 @@ const ERR_NO_PRICE_STREAM: &str = "price report not available for token pair";
 /// USDC ticker
 ///
 /// For now we only stream prices quoted in USDC
```
joeykraut, line 27: nit: update this comment

joeykraut (approved):

PR #498, opened 2024-03-28
workers: price-reporter: external_executor: resubscribe when reconnecting

akirillo:
This PR fixes a bug in the external price reporter executor in which we would not re-subscribe to previously subscribed-to price streams when establishing a new connection with the external price reporter due to it failing over.

In a deployed setting, this manifested in continual websocket hangup errors w/ the external price reporter occurring in the relayer, because nothing was being sent in either direction on the websocket and AWS ALB idle timeout was triggered, killing the connection.

Additionally, in a deployed setting, price reports were not showing up as stale when the above error pattern occurred, because the staleness window was expressed in milliseconds but interpreted as seconds. This PR updates that to express the staleness window in seconds.

I've confirmed that this works via an ad-hoc test of the following nature:
1. Run a local price reporter and relayer, and hit the `/price_report` api route to check for a nominal price report
2. Tear down the local price reporter, and hit the api route to check for a "not enough data reported" report (as the subscription is no longer active)
3. Spin the local price reporter back up, and hit the api route to check for a nominal price report once again

I've also confirmed that this test fails for the version of the relayer currently running in staging.

### DIFF
--- a/workers/price-reporter/src/manager/external_executor.rs
+++ b/workers/price-reporter/src/manager/external_executor.rs
@@ -8,7 +8,7 @@ use common::{
     default_wrapper::DefaultOption,
     new_async_shared,
     types::{
-        exchange::{Exchange, PriceReporterState, ALL_EXCHANGES},
+        exchange::{Exchange, PriceReporterState},
         token::{is_pair_named, Token},
         CancelChannel, Price,
     },
@@ -22,7 +22,7 @@ use futures::{
     stream::{SplitSink, SplitStream},
     SinkExt, StreamExt,
 };
-use job_types::price_reporter::{PriceReporterJob, PriceReporterReceiver};
+use job_types::price_reporter::{PriceReporterJob, PriceReporterQueue, PriceReporterReceiver};
 use serde::{Deserialize, Serialize};
 use tokio::{
     net::TcpStream,
@@ -109,7 +109,7 @@ impl ExternalPriceReporterExecutor {
             tokio::select! {
                 // Process price update from external price reporter
                 Some(price_message) = msg_in_rx.recv() => {
-                    self.handle_price_update(price_message).await.map_err(PriceReporterError::ExchangeConnectionError)?;
+                    self.handle_price_update(price_message).await.map_err(PriceReporterError::ExchangeConnection)?;
                 }
 
                 // Dequeue the next job from elsewhere in the local node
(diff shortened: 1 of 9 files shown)


workers/price-reporter/src/manager/external_executor.rs
```diff
@@ -326,34 +331,39 @@ async fn ws_handler_loop(
...
                 },
             }
         }
+
+        // Breaking from the loop above indicates a failure in the websocket connection.
+        // As such, we will have to re-subscribe to all the price streams that
+        // were previously subscribed to on the re-established connection, so we
+        // enqueue the re-subscription jobs here
+        info!("Enqueueing re-subscriptions to prior price streams...");
+        resubscribe_to_prior_streams(&price_reporter_job_queue, &subscription_states).await?;
```
joeykraut, line 355: Why not just have the already existing retry logic re-subscribe? Feels clunky to go through the job queue when we already have retry logic executed immediate after this (i.e. at the top of the next loop).

Feel like we can just pass `subscription_states` to `connect_with_retries` (maybe rename this to `connect_and_resubscribe`?

akirillo: Oh, haha - I did this because I thought it was cleaner than the alternative 😅 guess we shot past each other on this one.

I err against doing resubscriptions in the `connect_with_retries` method because on the first pass of the loop, we would double-subscribe to all price streams. this is because the `StreamPrice` jobs from the handshake manager are typically received before the `ws_handler_loop` is spawned (I have observed this), the outgoing subscription messages on the websocket are already enqueued & the subscription states map written to by the time the websocket is established. currently, the external price reporter is configured to handle double-subscriptions idempotently, but this doesn't feel like an assumption that should be made here.

As far as using the job queue vs the outbound websocket message channel directly, we already have the logic for subscribing to a new price stream well-defined in the handler for the price stream subscription job, and that generally is the interface for new subscriptions, so why not use it?

But I digress, I'm not deeply opinionated, and if you feel this violates patterns that have already been established in the codebase, I'll defer to consistency

akirillo: Something else I considered when deciding on this approach: when implementing stablecoin-quote price conversion, I'll want the following logic:
If we get a `StreamPrice` job for e.g. UNI-USDC, we'll want to ensure that we are also subscribed to both UNI-USDT and USDC-USDT for the conversion.
I was planning on just sending `StreamPrice` jobs for these additional pairs to the job queue, similarly to how we do here. I feel this centralizes & simplifies the means by which we subscribe to price streams instead of having yet another code path responsible for this

joeykraut: The fact that we're designing this with the implementation of the API in mind means the API interface is bad. And playing games like "which message arrives first" is not a principled approach to design.

Way more readable to run a method than go through the queue

akirillo: yeah, fair points. I'll go ahead and restructure w/o the queue

joeykraut (approved):

PR #500, opened 2024-03-28
workers: price-reporter: per-exchange stream states

akirillo:
This PR is introduces some preliminary changes to facilitate the implementation of stablecoin-quote price conversion. Primarily:
1. Adds a static set of stablecoin tokens, which are candidates for price conversion when they are a quote asset
2. Adds a static mapping from exchange -> most liquid stable quote asset for that exchange, this will be the pair through which price is converted when the pair in the order is quoted with a different stablecoin
3. Lifts the mapping from exchange -> atomic (price, timestamp) out of the `AtomicPriceStreamState`. This makes the subsequent implementation of kicking off price streams for potentially different pairs than what was requested, as well as computing the price via conversion through the most liquid stable quote asset, much smoother. For example, when requesting prices for UNI<>USDC, that will mean a UNI<>USDT price stream on Binance, and a UNI<>USD price stream on Kraken. Because of a potentially non-uniform quote on the underlying price streams with exchanges, it no longer makes sense to couple the price stream state for all exchanges under a single pair.

These changes are a work in progress, and will continue to take shape in subsequent PRs for price conversion. I wanted to get this PR up as I am fairly confident in this direction and wanted to split up review.

### DIFF
--- a/common/src/types/token.rs
+++ b/common/src/types/token.rs
@@ -322,13 +322,13 @@ lazy_static! {
         exchange_tickers
     };
     /// The set of stablecoin quote assets
-    static ref STABLECOIN_QUOTES: HashSet<Token> = {
+    pub static ref STABLECOIN_QUOTES: HashSet<Token> = {
         let stablecoin_tickers = [USDC_TICKER, USDT_TICKER];
         stablecoin_tickers.iter().map(|ticker| Token::from_ticker(ticker)).collect()
     };
     /// The mapping between exchanges and the stablecoin quote asset on which they are generally
     /// the most liquid
-    static ref MOST_LIQ_STABLE: HashMap<Exchange, Token> = {
+    pub static ref MOST_LIQ_STABLE: HashMap<Exchange, Token> = {
         let mut most_liquid_stable = HashMap::<Exchange, Token>::new();
 
         // Most liquid stablecoin quote on Binance is generally USDT
--- a/common/src/types/token.rs
+++ b/common/src/types/token.rs
@@ -28,6 +28,64 @@ use crate::biguint_to_str_addr;
 
 use super::exchange::{Exchange, ALL_EXCHANGES};
 
+// ----------------
+// | Quote Tokens |
+// ----------------
+
+/// USDC ticker
+pub const USDC_TICKER: &str = "USDC";
+/// USDT ticker
+pub const USDT_TICKER: &str = "USDT";
+/// USD ticker
+///
+/// We don't actually allow USD as a quote asset since it's not an ERC20,
+/// but it is used as a quote in some exchanges, so we must be able to
+/// stream prices for it.
+pub const USD_TICKER: &str = "USD";
+
+// ---------------
+// | Base Tokens |
+// ---------------
+
+/// WBTC ticker
+pub const WBTC_TICKER: &str = "WBTC";
+/// WETH ticker
+pub const WETH_TICKER: &str = "WETH";
+/// BNB ticker
+pub const BNB_TICKER: &str = "BNB";
+/// MATIC ticker
+pub const MATIC_TICKER: &str = "MATIC";
+/// LDO ticker
+pub const LDO_TICKER: &str = "LDO";
+/// LINK ticker
+pub const LINK_TICKER: &str = "LINK";
+/// UNI ticker
+pub const UNI_TICKER: &str = "UNI";
+/// CRV ticker
+pub const CRV_TICKER: &str = "CRV";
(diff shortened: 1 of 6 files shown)


common/src/types/token.rs
```diff
@@ -62,126 +113,155 @@ pub static TICKER_NAMES: &[(
...
     ),
     // LSDs
-    ("LDO", ExchangeTicker::Same, ExchangeTicker::Same, ExchangeTicker::Same, ExchangeTicker::Same),
-    // Stables
     (
-        "USDC",
+        LDO_TICKER,
+        ExchangeTicker::Same,
         ExchangeTicker::Same,
-        ExchangeTicker::Renamed("USD"), // USDC has 1:1 peg with USD on Coinbase
```
akirillo, line 98: The USD- & USDC-quoted markets are literally the same on Coinbase, they are aliased to one another in the API. So it's simpler to just use the USDC ticker in Coinbase (and we don't need to convert a USDC-quoted pair through a USD quote down the line)

common/src/types/token.rs
```diff
@@ -240,6 +321,29 @@ lazy_static! {
...
+        most_liquid_stable.insert(Exchange::Coinbase, Token::from_ticker(USDC_TICKER));
+        // Most liquid "stablecoin" quote on Kraken is generally USD
+        most_liquid_stable.insert(Exchange::Kraken, Token::from_ticker(USD_TICKER));
+        // Most liquid stablecoin quote on Okx is generally USDT
+        most_liquid_stable.insert(Exchange::Okx, Token::from_ticker(USDT_TICKER));
+
+        // We currently omit UniswapV3 here, as it requires more involved routing logic
+
+        most_liquid_stable
+    };
```
joeykraut, line 346: this should probably just be a method, we can make it static if we like, but we should mimize the scope of these `lazy_statics` as much as possible. I.e. something like:
```rust
default_exchange_stable(exchange: Exchange) -> Token {
    match exchange {
        Exchange::Binance => Token::from_ticker(USDT_TICKER),
        ...
    }
}
```

Also notice the rename, "most liquid" is _currently_ how we set the value, but "default" captures a more general notion

workers/price-reporter/src/manager.rs
```diff
@@ -70,49 +70,56 @@ pub struct PriceReporter {
...
+#[derive(Debug, Default)]
 pub struct AtomicPriceStreamState {
-    /// The price information for each exchange, updated by the
-    /// `ConnectionMuxer`
-    price_map: HashMap<Exchange, Arc<AtomicF64>>,
-    /// A map indicating the time at which the last price was received from each
-    /// exchange
-    last_received: HashMap<Exchange, Arc<AtomicU64>>,
+    /// The price of the pair on the exchange
+    price_map: AtomicF64,
```
joeykraut, line 78: nit: This isn't a map anymore, just `price`

common/src/types/token.rs
```diff
@@ -240,6 +321,29 @@ lazy_static! {
         }
         exchange_tickers
     };
+    /// The set of stablecoin quote assets
+    pub static ref STABLECOIN_QUOTES: HashSet<Token> = {
```
joeykraut, line 325: I think we should remove this, it isn't used anywhere. A couple other points on this:
1. We don't need a hashset for two elements, it's almost certainly slower than a vector
2. I don't think we need to lazily initialize anything about our quote tokens, especially given that there are two of them. `Token::from_ticker` is totally fine to do at runtime, and I'd far prefer that to more `lazy_static`s.

workers/price-reporter/src/manager.rs
```diff
@@ -70,49 +70,56 @@ pub struct PriceReporter {
     pub(super) manager_executor_handle: Option<JoinHandle<PriceReporterError>>,
 }
 
-/// The state streamed from the connection multiplexer to the price reporter
+/// The state of a price stream for a given (exchange, base, quote).
 /// Uses atomic primitives to allow for hardware synchronized update streaming
-#[derive(Clone, Debug)]
+#[derive(Debug, Default)]
 pub struct AtomicPriceStreamState {
-    /// The price information for each exchange, updated by the
```
joeykraut, line 77: clean, i like this decision

joeykraut (approved):
very clean

common/src/types/token.rs
akirillo: fair, I'll just do a constant array of quote tickers. but this will be used in the next PR! got a bit ahead of myself in this one and decided to cut it short

PR #502, opened 2024-03-30
workers: price-reporter: stable quote price conversion in native executor

akirillo:
This PR implements stable-quote price conversion when the relayer uses the native price reporter. We do this by lifting much of the shared functionality around initializing price stream states for the pairs in the conversion route, and the price conversion itself, into a `SharedPriceStreamStates` struct, which is now used by both the external and native executors.

Much of the rest of the structure of the native executor is the same, i.e. there is still a separate `Reporter` per pair, that still spawns a `ConnectionMuxer`, except now the `Reporter` handles price conversion for the pair under the hood, and the `ConnectionMuxer` also manages connections to the pairs in the conversion route.

This was tested ad-hoc by running a relayer locally, using both the native and external price reporter executor, and ensuring that price conversion is taking place by logging in the `convert_through_default_stable` method.

**EDIT:**
Subsequent to the first PR review, I've made the following changes to the code structure to cleanly ensure that no duplicate streams are maintained:
1. There is no longer a `Reporter` struct. Since the price for a requested (base, quote) pair now may go through other token pairs, it no longer makes sense to group streaming logic by the requested (base, quote)
2. Similarly, it no longer makes sense to mux a streamed pair across all exchanges. It is much simpler to handle each (exchange, base, quote) stream separately. To this end, the `ConnectionMuxer` no longer muxes, it just handles streaming prices & retrying connections, and has been renamed to `ExchangeConnectionManager`. It is managed directly by the `PriceReporterExecutor`.
3. Because there is no longer a `Reporter`, we no longer maintain a price streamer loop. We compute price reports and publish them to the system bus when updates come through the exchange connections, as in the external executor.
    - However, I noticed the way we do this is flawed. For example, when we get a price update on a (stable quote, default stable quote), we don't actually want to compute a price report for it. We also need to know which requested / top-level pairs this conversion pair is being used for so we can compute price reports for them. The fix for this is coming in the next PR.


workers/price-reporter/src/manager.rs
```diff
@@ -97,29 +98,127 @@ impl AtomicPriceStreamState {
...
-impl ExchangeStates {
-    /// Construct a new exchange states instance from a set of exchanges
-    pub fn new_from_exchanges(exchanges: &[Exchange]) -> Self {
-        let mut exchange_states = ExchangeStates(HashMap::new());
-        for exchange in exchanges {
-            exchange_states.0.insert(*exchange, Arc::new(AtomicPriceStreamState::default()));
-        }
-        exchange_states
+pub struct SharedPriceStreamStates(
+    AsyncShared<HashMap<(Exchange, Token, Token), AtomicPriceStreamState>>,
```
joeykraut, line 105: no real point in making the state atomic if it's wrapped in a lock, it might actually be net slower if atomic

workers/price-reporter/src/reporter.rs
```diff
@@ -255,7 +196,7 @@ impl ConnectionMuxer {
 
                 // New price streamed from an exchange
```
joeykraut, line 197: we should really break this loop up, control flow depth of 6 is pretty hard to read

workers/price-reporter/src/reporter.rs
```diff
@@ -300,26 +244,73 @@ impl ConnectionMuxer {
...
+                .await?;
+
+                // Connect to the price stream for the quote / default stable
+                // pair
+                self.initialize_stream(
+                    *exchange,
+                    self.quote_token.clone(),
+                    default_stable,
+                    &mut stream_map,
+                )
```
joeykraut, line 279: similiar thing to the other PR, lots of repeated logic here

workers/price-reporter/src/reporter.rs
```diff
@@ -300,26 +244,73 @@ impl ConnectionMuxer {
     /// `StreamMap` for multiplexing
     async fn initialize_connections<'a>(
```
joeykraut, line 245: I find it odd that the `ConnectionMuxer` handles price conversion. It's job is to manage connections and forward data, I feel the layers on top of this should do price conversion. 

It makes sense to do price conversion where we do aggregation; neither of which belong in the connection muxer.

joeykraut: Plus putting it in the connection muxer means we subscribe to the `quote/default_quote` topic for _every_ new asset. Lots of wasted work + bandwidth

workers/price-reporter/src/manager.rs
```diff
@@ -97,29 +98,127 @@ impl AtomicPriceStreamState {
...
-
-impl ExchangeStates {
-    /// Construct a new exchange states instance from a set of exchanges
-    pub fn new_from_exchanges(exchanges: &[Exchange]) -> Self {
-        let mut exchange_states = ExchangeStates(HashMap::new());
-        for exchange in exchanges {
-            exchange_states.0.insert(*exchange, Arc::new(AtomicPriceStreamState::default()));
-        }
-        exchange_states
+pub struct SharedPriceStreamStates(
```
joeykraut, line 104: See my comment below about the connection muxer, I don't think it should handle price conversion. Another reason is that we pay for it here. We need a more complex structure under the assumption that the connection muxer should be handling possibly multiple token pairs. 

And that is only to facilitate _two_ pairs (the base pair and the conversion pair); seems like far to heavy a change to handle two pairs

workers/price-reporter/src/manager.rs
akirillo: we only ever acquire a write lock on this object when adding an entry to it, when updating the atomic state it's just a read lock, so we don't add locking overhead to updating the atomic

workers/price-reporter/src/reporter.rs
akirillo: The latter is a good point, currently all the default pairs are on the same quote, and this will generally be true...

I'll take a stab at having the `Reporter` (where price conversion currently occurs via `get_state`) be responsible for spawning multiple `ConnectionMuxer`s, maintaining the functionality that each `ConnectionMuxer` streams the same pair directly from all of the exchanges that support it, as opposed to spawning any connections necessary for reporting the price for a requested pair

workers/price-reporter/src/manager.rs
akirillo: I would much rather use the `SharedPriceStreamStates` in the `Reporter` than not, it centralizes the logic used to aggregate prices (including price conversion, deviation checks, etc.) between the native and external price reporter. I'll explore a sensible way to avoid passing the struct down to the `ConnectionMuxer`, but I think overall, it largely simplifies the price reporting code.

workers/price-reporter/src/manager.rs
```diff
@@ -201,3 +372,41 @@ fn ts_too_stale(ts: u64) -> (bool, u64) {
     let time_diff = get_current_time_seconds() - ts;
     (time_diff > MAX_REPORT_AGE, time_diff)
 }
+
+/// Returns the (exchange, base, quote) tuples for which price streams are
+/// required to accurately compute the price for the (requested_base,
+/// requested_quote) pair
+pub fn required_streams_for_pair(
+    requested_base: Token,
```
joeykraut, line 380: clippy errors here and below

joeykraut (approved):

PR #503, opened 2024-03-30
common: wallet: balances, orders: Make order/balance append consistent

joeykraut:
### Purpose
This PR modifies the logic for balance and order updates to make them mutually consistent:
- Orders are now overwritten if they are zero'd (amount is zero)
- Balances are now overwritten if they are zero'd (amount and fees are zero)
- Balances are added to the wallet in the same fashion orders are, i.e:
    - Overwriting the first zero'd balance if it exists
    - Appending if possible
    - Erroring otherwise

### Testing
- Unit tests pass

### DIFF
--- a/common/src/types/wallet/balances.rs
+++ b/common/src/types/wallet/balances.rs
@@ -82,22 +82,20 @@ impl Wallet {
             return Ok(());
         }
 
-        // Otherwise, add the balance
-        if self.balances.len() < MAX_BALANCES {
-            self.balances.insert(balance.mint.clone(), balance);
-            return Ok(());
+        if let Some(index) = self.find_first_replaceable_balance() {
+            self.balances.replace_at_index(index, balance.mint.clone(), balance);
+        } else if self.balances.len() < MAX_BALANCES {
+            self.balances.append(balance.mint.clone(), balance);
+        } else {
+            return Err(ERR_BALANCES_FULL.to_string());
         }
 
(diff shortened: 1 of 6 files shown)


common/src/types/wallet/balances.rs
```diff
@@ -82,24 +82,22 @@ impl Wallet {
             return Ok(());
         }
 
-        // Otherwise, add the balance
-        if self.balances.len() < MAX_BALANCES {
-            self.balances.insert(balance.mint.clone(), balance);
-            return Ok(());
+        if let Some(index) = self.find_first_replaceable_balance() {
+            self.balances.insert_at_index(index, balance.mint.clone(), balance);
```
sehyunc, line 86: KeyedList's `[insert_at_index](https://github.com/renegade-fi/renegade/blob/joey/balance-append-logic/common/src/keyed_list.rs#L155)` method inserts an element at position index within the vector, shifting all elements after it to the right. My understanding was the behavior we want is to replace the zero balance (and zero order).

sehyunc: Should we use `replace_at_index` instead if this is indeed the correct behavior?

sehyunc: Actually, I remember you mentioning that balances must maintain their order, so we cannot overwrite/replace for balances.

sehyunc (approved):
Tested on frontend, able to settle internal matches through the interface.

PR #508, opened 2024-04-09
workers: task-driver: driver, state: applicator: task_queue: await queue pausing & reject double-pauses

akirillo:
This PR changes the applicator logic for the task queue such that preemption is not supported when a queue that a task attempts to preempt is already paused. This way, only one preemptive task for a queue can be running at a given time. This should mitigate the observed error w/ conflicting preemptive tasks causing double-resumes, spawning duplicates of the task at the top of the queue. This requires modification of the logic in the task driver to await the successful pausing of queues for a preemptive task.

This PR also includes some added telemetry which aided in the debugging of the issue described above, it generally seemed good to keep around.

These changes are running in staging, I will keep an eye on whether this issue gets observed again.

### DIFF
--- a/arbitrum-client/src/client/contract_interaction.rs
+++ b/arbitrum-client/src/client/contract_interaction.rs
@@ -18,7 +18,7 @@ use constants::Scalar;
 use contracts_common::types::MatchPayload;
 use renegade_crypto::fields::{scalar_to_u256, u256_to_scalar};
 use tracing::{info, instrument};
-use util::err_str;
+use util::{err_str, telemetry::helpers::backfill_trace_field};
 
 use crate::{
     conversion::{
@@ -133,7 +133,7 @@ impl ArbitrumClient {
         .await?;
 
         let tx_hash = format!("{:#x}", receipt.transaction_hash);
-        tracing::Span::current().record("tx_hash", &tx_hash);
+        backfill_trace_field("tx_hash", &tx_hash);
         info!("`new_wallet` tx hash: {}", tx_hash);
 
         Ok(())
@@ -174,7 +174,7 @@ impl ArbitrumClient {
         .await?;
 
         let tx_hash = format!("{:#x}", receipt.transaction_hash);
-        tracing::Span::current().record("tx_hash", &tx_hash);
+        backfill_trace_field("tx_hash", &tx_hash);
         info!("`update_wallet` tx hash: {}", tx_hash);
 
         Ok(())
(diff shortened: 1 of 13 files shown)


arbitrum-client/src/client/contract_interaction.rs
```diff
@@ -111,6 +114,8 @@ impl ArbitrumClient {
     ///
     /// Awaits until the transaction is confirmed on-chain
     #[instrument(skip_all, err, fields(
+        proof_calldata,
+        statement_calldata,
```
joeykraut, line 118: do we really want to log all calldata? That's quite a lot of data to be sending to DD

state/src/applicator/task_queue.rs
```diff
@@ -168,6 +179,7 @@ impl StateApplicator {
     // -----------
 
     /// Start a task if the current peer is the executor
+    #[instrument(skip_all, err, fields(task_id = %task.id, task = %task.descriptor.display_description()))]
```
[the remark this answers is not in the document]
joeykraut: No l don't think we need to log that

arbitrum-client/src/client/contract_interaction.rs
joeykraut: Plus if it's a hex string that's almost entirely useless in a trace except to check it against another hex string. That case is unlikely to come up (under working circumstances all calldata is mostly different). I'd say dumping calldata into DD doesn't warrant the benefit it gives

workers/task-driver/src/driver.rs
```diff
@@ -260,19 +256,24 @@ impl TaskExecutor {
...
+                    state
+                        .pause_task_queue(wallet_id)
+                        .expect("error proposing task queue pause for wallet {wallet_id}")
+                        .await
+                        .expect("error pausing task queue for wallet {wallet_id}");
+                }
+
+                // Add the task to the preemptive tasks list so that notification requests can
+                // be registered for it
+                preemptive_tasks.write().unwrap().insert(task_id);
```
joeykraut, line 275: now that there's a decent chunk of logic in this block, probably makes sense to break this out into its own `async` method, easier to follow that way

joeykraut (approved):

arbitrum-client/src/client/contract_interaction.rs
akirillo: I did use this when debugging the `verification failed` issue from before, I can decode this calldata in a unit test that uses it to invoke contract code. This is generally something I could get from a transaction receipt, but of course that's only when the transaction succeeds.

My rationale here is that the testing we're doing is surfacing issues that can be hard to reproduce, so it's quite helpful to be equipped w/ test data when an issue is observed, and I figure we can eat the DD bill for a while.

But that said, this does feel clunky, and is a lot of data that we do pay for - generally the observed issues are reproducible if you just give the load testing a bit of time to snag them again.

So I'm ok with removing right now

PR #512, opened 2024-04-16
circuits, test-helpers: Ingest interface changes from `ark-mpc`

joeykraut:
### Purpose
This PR adjusts the interfaces used in `circuits` and `test-helpers` to match changes in `ark-mpc`. Specifically these changes force the `OfflinePhase` (formerly `BeaverSource`) to return MACs with the shared values.

These changes also get all tests passing in the workspace.

Finally, the changes ingest a few optimizations that were made while integrating the offline phase, resulting in a significant speedup in match-settle latency.

### Testing

### DIFF
--- a/test-helpers/src/mpc_network/mocks.rs
+++ b/test-helpers/src/mpc_network/mocks.rs
@@ -6,9 +6,10 @@ use std::{
 };
 
 use ark_mpc::{
-    beaver::SharedValueSource,
+    algebra::ScalarShare,
     error::MpcNetworkError,
     network::{MpcNetwork, NetworkOutbound, PartyId},
+    offline_prep::OfflinePhase,
 };
 use async_trait::async_trait;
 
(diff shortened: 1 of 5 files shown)


test-helpers/src/mpc_network/mocks.rs
```diff
@@ -37,27 +38,54 @@ impl PartyIDBeaverSource {
...
+        (
+            ScalarShare::new(a_share, a_mac),
+            ScalarShare::new(b_share, b_mac),
+            ScalarShare::new(c_share, c_mac),
+        )
     }
 
-    fn next_shared_inverse_pair(&mut self) -> (Scalar, Scalar) {
-        (Scalar::from(self.party_id), Scalar::from(self.party_id))
+    fn next_shared_inverse_pair(
```
akirillo, line 78: is this a valid inverse pair for party id = 0?

akirillo (approved):

PR #518, opened 2024-04-22
core: setup, state: interface: increase election & heartbeat intervals

akirillo:
This PR increases the Raft heartbeat interval to 1s and the election timeout to be bounded between 10-15s. This ensures that Raft elections are not instigated unnecessarily when the network manager is handling bursts of load, in which case Raft heartbeats take longer to dequeue and execute.

We also move the node setup to after the gossip server is spun up, so that nodes are able to gossip raft messages, receive heartbeats, and establish leadership accordingly before the node setup delay kicks in - otherwise this gives nodes enough time to self-elect before heartbeating and thus never have a consistent view of leadership (as they would self-elect as leaders w/ the same term and # of log entries before ever heartbeating, leading to a stalemate)

I tested this by running a 2-node cluster locally in debug mode, and am able to successfully go from empty state to two wallets that match and settle an order internally.

### DIFF
--- a/core/src/main.rs
+++ b/core/src/main.rs
@@ -180,22 +180,6 @@ async fn main() -> Result<(), CoordinatorError> {
         mpsc::channel(1 /* buffer_size */);
     watch_worker::<ProofManager>(&mut proof_manager, &proof_manager_failure_sender);
 
-    // Setup the relayer wallet once the task driver and proof manager are running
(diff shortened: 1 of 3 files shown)


joeykraut (approved):

joeykraut:
nice work debugging this one

PR #520, opened 2024-04-23
state: applicator: Add return types to applicator

joeykraut:
### Purpose
This PR adds return types to the state applicator. These return types allow the applicator to return a value to the waiters of a proposal. This will be used for allocating preprocessing values to a given node in the MPC prep.

I also changed the semantics of the `ProposalWaiter` so that they can no longer be joined. This simplifies the implementation of the return value semantics as we avoid buffering results of partially ready joins. This functionality was not used anywhere -- just vestigial complexity.

### Testing
- Unit tests pass crate-wide

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,12 +14,8 @@ jobs:
       uses: arduino/setup-protoc@v1
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
-    - name: Build
-      uses: actions-rs/cargo@v1
-      with:
-        command: build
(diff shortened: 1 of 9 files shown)


joeykraut:
> if we find ourselves needing to return from other state transitions, may be nice to define a state transition trait where it can apply itself given a handle to a `StateApplicator`, and have an associated type for the return. we could make `ProposalWaiter`s generic over this

Yeah I thought about this, it gets complex trying to store these proposal waiters, but could be worth it if we need to return a variety of values, like you suggest.

PR #521, opened 2024-04-23
state: interface: mpc-preprocessing: Add mpc preprocessing interface

joeykraut:
### Purpose
This PR adds an interface for adding and consuming preprocessing values. The implementation in the applicator is deferred for a follow up, as well as testing. This PR essentially lays out the scaffolding needed for mpc preprocessing interaction with consensus.

### Testing
- Unit tests pass crate wide


state/src/storage/tx/mpc_preprocessing.rs
```diff
@@ -28,6 +32,26 @@ impl<'db, T: TransactionKind> StateTxn<'db, T> {
...
+    /// Get the size of the preprocessing value for a cluster
+    pub fn get_mpc_prep_size(
+        &self,
+        cluster: &ClusterId,
+    ) -> Result<Option<PreprocessingSlice>, StorageError> {
+        let prep = res_some!(self.get_mpc_prep(cluster)?).values;
+        let slice = PreprocessingSlice {
+            num_bits: prep.random_bits.len(),
+            num_values: prep.random_values.len(),
+            num_input_masks: usize::min(
```
akirillo, line 45: it makes sense to use `usize::min` here regardless, but IIUC there should never be a case where the input masks have different lengths, right?

akirillo (approved):

state/src/storage/tx/mpc_preprocessing.rs
joeykraut: that's correct

PR #522, opened 2024-04-23
state: applicator: mpc-preprocessing: Implement state transitions

joeykraut:
### Purpose
This PR implements the state transition logic for consuming and appending MPC preprocessing values to the preprocessing store for a given cluster pair. 

### Testing
- Unit tests pass
- Tested the applicator methods as well as through raft consensus


akirillo (approved):
looking clean

PR #523, opened 2024-04-24
state: storage: order-metadata: Setup order metadata storage primitive

joeykraut:
### Purpose
This PR sets up storage primitives for order metadata and order history. For each wallet, we store a list of historical orders with states. This store is denormalized from the wallet store.

### Testing
- Unit tests pass

### DIFF
--- a/state/src/storage/tx/mod.rs
+++ b/state/src/storage/tx/mod.rs
@@ -9,6 +9,7 @@
 pub mod mpc_preprocessing;
 pub mod node_metadata;
 pub mod order_book;
+pub mod order_metadata;
 pub mod peer_index;
 pub mod raft_log;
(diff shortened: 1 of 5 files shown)


state/src/storage/tx/mod.rs
```diff
@@ -96,6 +97,7 @@ impl<'db> StateTxn<'db, RW> {
             CLUSTER_MEMBERSHIP_TABLE,
             PRIORITIES_TABLE,
             ORDERS_TABLE,
+            ORDER_HISTORY_TABLE,
```
akirillo, line 100: We need to bump the `NUM_TABLES` const right

akirillo (approved):

PR #525, opened 2024-04-24
state: applicator: wallet-index: Update order history on wallet update

joeykraut:
### Purpose
This PR updates order history for new and cancelled orders in the wallet update handlers. We modify order history here to hide the wallet index / order history denormalization from the interface. That is, when a wallet is updated, the order history is brought to a consistent state without the caller having to separately update order history.

### Testing
- Unit tests pass
- Tested order history via wallet update interface

### DIFF
--- a/state/src/applicator/wallet_index.rs
+++ b/state/src/applicator/wallet_index.rs
@@ -1,13 +1,22 @@
 //! Applicator methods for the wallet index, separated out for discoverability
 
-use common::types::{network_order::NetworkOrder, wallet::Wallet};
+use common::types::{
+    network_order::NetworkOrder,
+    wallet::{
+        order_metadata::{OrderMetadata, OrderState},
+        Wallet,
+    },
+};
 use external_api::bus_message::{wallet_topic, SystemBusMessage};
 use itertools::Itertools;
 use libmdbx::RW;
 
 use crate::storage::tx::StateTxn;
 
-use super::{Result, StateApplicator};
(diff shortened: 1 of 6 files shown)


state/src/applicator/wallet_index.rs
```diff
@@ -80,20 +76,85 @@ impl StateApplicator {
...
+        // Change the state of each active order in the wallet to `Created`
+        // to reflect that validity proofs have not been created yet
+        let wallet_id = wallet.wallet_id;
+        for id in wallet.get_nonzero_orders().into_keys() {
+            let mut md = tx
+                .get_order_metadata(wallet_id, id)?
+                .ok_or(StateApplicatorError::MissingEntry(ERR_NO_METADATA))?;
+            md.state = OrderState::Created;
+            self.update_order_metadata_with_tx(tx, md)?;
+        }
```
akirillo, line 100: This means that all existing orders (not new or cancelled) will have their states overwritten to `OrderState::Created` - I could see some confusion on the frontend seeing an order go from `OrderState::Matching` to `OrderState::Created`. Thoughts on adding a state for `OrderState::GeneratingProofs` (or something)? If not let's make sure @sehyunc is aware of this so it's represented well on the FE

akirillo (approved):
just some thoughts about going from `Matching` -> `Created`

state/src/applicator/wallet_index.rs
joeykraut: Created can be displayed as "GeneratingProofs" if we like.

PR #531, opened 2024-04-25
state: applicator: only run matching engine on executor of last task

akirillo:
This PR adjusts the conditions under which we execute the internal matching engine after popping a task, adding the clause that the engine should only be run by the executor of the last task. This is currently the most straightforward way to ensure that the engine is only run on one node.

Additionally, this PR removes the `expect`ation of successful task queue pausing when running a preemptive task, as this can error healthily in cases such as attempting to preempt a committed task, or pause an already-paused queue.

The combination of these two changes should remediate the issue we were seeing in `dev` where nodes would panic after a match was found.

### DIFF
--- a/state/src/applicator/task_queue.rs
+++ b/state/src/applicator/task_queue.rs
@@ -68,17 +68,14 @@ impl StateApplicator {
         let task = tx.pop_task(&key)?.ok_or_else(|| StateApplicatorError::TaskQueueEmpty(key))?;
 
         // If the queue is non-empty, start the next task
-        let tasks = tx.get_queued_tasks(&key)?;
-        if let Some(task) = tasks.first() {
+        let remaining_tasks = tx.get_queued_tasks(&key)?;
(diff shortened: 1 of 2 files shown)


state/src/applicator/task_queue.rs
```diff
@@ -74,11 +74,8 @@ impl StateApplicator {
             self.maybe_start_task(task, &tx)?;
         }
 
-        // If the queue is empty and the task was a wallet task, run the matching engine
-        // on all orders that are ready
-        // TODO: We should only have one node execute the internal matching engine,
-        // though this is okay for the moment
-        if tasks.is_empty() && task.descriptor.is_wallet_task() {
+        if Self::should_run_matching_engine(&tasks, &task, &tx)? {
```
joeykraut, line 77: nit: can we rename `tasks` to `remaining_tasks` or something? `tasks` next to `task` reads a little odd

joeykraut (approved):
clean, nice

PR #533, opened 2024-04-26
workers: task-driver, api-server: Settle fees only on withdrawal

joeykraut:
### Purpose
This PR moves fee payment out of the match-settle flow. Instead, fee payment is done when a user withdraws. This removes the period after a match in which a wallet is ineligible for another match.

### Testing
- Tested against a local relayer

### DIFF
--- a/workers/api-server/src/http/wallet.rs
+++ b/workers/api-server/src/http/wallet.rs
@@ -7,8 +7,8 @@ use circuit_types::{
 };
 use common::types::{
     tasks::{
-        LookupWalletTaskDescriptor, NewWalletTaskDescriptor, TaskDescriptor, TaskIdentifier,
-        UpdateWalletTaskDescriptor,
+        LookupWalletTaskDescriptor, NewWalletTaskDescriptor, PayOfflineFeeTaskDescriptor,
+        TaskDescriptor, TaskIdentifier, UpdateWalletTaskDescriptor,
     },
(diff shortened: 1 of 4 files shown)


workers/api-server/src/http/wallet.rs
```diff
@@ -69,6 +69,32 @@ async fn append_task_and_await(
...
+/// Enqueue tasks to pay all fees for a wallet
+///
+/// Modifies the wallet in place to remove fees
+async fn pay_wallet_fees(wallet: &mut Wallet, state: &State) -> Result<(), ApiServerError> {
+    let wallet_id = wallet.wallet_id;
+    for (mint, balance) in wallet.balances.iter_mut() {
+        if balance.relayer_fee_balance > 0 {
+            // Pay relayer fee
+            let task = PayOfflineFeeTaskDescriptor::new_relayer_fee(wallet_id, mint.clone())
+                .expect("infallible");
```
akirillo, line 81: nit but why do these constructors return results

akirillo (approved):

workers/api-server/src/http/wallet.rs
joeykraut: Some tasks _are_ fallible; I wanted interface consistently but didn't want to go so far as to force it into a trait. We can remove it down the line if we like

PR #536, opened 2024-04-26
state: applicator: task-queue: Add preemptive tasks to queue

joeykraut:
### Purpose
This PR adds preemptive tasks to the queue and pops them when the queue is resumed. This will allow:
1. Task history to be made aware of matches on the wallet
2. Task simulation at the API layer to be aware of match updates to a wallet

### Testing
- Unit tests pass
- Updated task queue unit tests to check for the preemptive task when paused

### DIFF
--- a/state/src/interface/task_queue.rs
+++ b/state/src/interface/task_queue.rs
@@ -122,9 +122,25 @@ impl State {
         self.send_proposal(StateTransition::TransitionTask { task_id, state })
     }
 
-    /// Pause a task queue
-    pub fn pause_task_queue(&self, key: &TaskQueueKey) -> Result<ProposalWaiter, StateError> {
-        self.send_proposal(StateTransition::PreemptTaskQueue { key: *key })
+    /// Pause a task queue placing the given task at the front of the queue
+    #[instrument(name = "propose_preempt_task_queue", skip_all, err, fields(task_id, task = %task.display_description()))]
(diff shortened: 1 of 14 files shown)


state/src/interface/task_queue.rs
```diff
@@ -122,9 +122,25 @@ impl State {
...
-        self.send_proposal(StateTransition::PreemptTaskQueue { key: *key })
+    /// Pause a task queue placing the given task at the front of the queue
+    #[instrument(name = "propose_preempt_task_queue", skip_all, err, fields(task_id, task = %task.display_description()))]
+    pub fn pause_task_queue(
+        &self,
+        key: &TaskQueueKey,
+        task: TaskDescriptor,
+    ) -> Result<ProposalWaiter, StateError> {
+        // Pick a task ID and create a task from the description
+        let id = TaskIdentifier::new_v4();
```
akirillo, line 133: hmm, we should have already created a task ID for the preemptive task before we propose pausing the queue. At the very least for a consistent view of traces, could we have that be passed in here and used? Not sure if it breaks some assumptions to use the same task ID on two different queues, though

akirillo (approved):

state/src/interface/task_queue.rs
joeykraut: yeah that works, done

PR #537, opened 2024-04-26
workers: api-server: Add `/v0/wallet/:id/back_of_queue` endpoint

joeykraut:
### Purpose
This PR adds an endpoint to query the "back of queue" wallet: that is, the wallet after all queued tasks have been speculatively applied to it. This allows clients to sign multiple queued state transitions by having the relayer simulate their in-flight state transitions.

We use this functionality to construct `UpdateWalletTaskDescriptor` types so that the API layer correctly constructs the  updated wallet from the back of queue.

I also fixed a bug that results from being able to enqueue multiple tasks: the `UpdateWalletTask` needs to fetch the most recent version of the `old_wallet` in its constructor to -- in particular -- update to the most recent Merkle path.

### Testing
- Unit tests pass
- Implemented the client logic in the CLI and tested there. Queued up multiple tasks of all kinds and verified their transitions were successful and correct.

### DIFF
--- a/workers/task-driver/src/tasks/update_wallet.rs
+++ b/workers/task-driver/src/tasks/update_wallet.rs
@@ -10,7 +10,6 @@ use std::fmt::{Display, Formatter, Result as FmtResult};
 use arbitrum_client::client::ArbitrumClient;
 use async_trait::async_trait;
 use circuit_types::transfers::ExternalTransferDirection;
-use circuit_types::{native_helpers::wallet_from_blinded_shares, SizedWallet};
 use circuits::zk_circuits::valid_wallet_update::{
     SizedValidWalletUpdateStatement, SizedValidWalletUpdateWitness,
 };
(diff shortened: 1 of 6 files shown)


workers/task-driver/src/tasks/update_wallet.rs
```diff
@@ -330,22 +332,6 @@ impl UpdateWalletTask {
     // | Helpers |
     // -----------
 
-    /// Check the construction of a wallet's shares, i.e. that the shares match
-    /// the wallet as a whole
-    fn check_wallet_shares(new_wallet: &Wallet) -> Result<(), UpdateWalletTaskError> {
```
joeykraut, line 335: handled in the descriptor constructor

workers/task-driver/src/tasks/update_wallet.rs
```diff
@@ -184,12 +183,15 @@ impl Task for UpdateWalletTask {
     type Descriptor = UpdateWalletTaskDescriptor;
 
     async fn new(descriptor: Self::Descriptor, ctx: TaskContext) -> Result<Self, Self::Error> {
-        // Safety check, the new wallet's secret shares must recover the new wallet
-        Self::check_wallet_shares(&descriptor.new_wallet)?;
+        // Pull in the most recent version of the `old_wallet`
+        let old_wallet = ctx
+            .state
+            .get_wallet(&descriptor.old_wallet.wallet_id)?
```
akirillo, line 189: Should the descriptor just contain the `old_wallet_id` then instead of the whole `old_wallet`? Looks like that's the only field on the `old_wallet` that we use anywhere

akirillo (approved):

workers/task-driver/src/tasks/update_wallet.rs
joeykraut: good suggestion, done

PR #539, opened 2024-04-27
state: order-history: Add all order fields to stored history

joeykraut:
### Purpose
This PR updates the order history to store the order itself under the `data` field. As well, this PR fixes a bug wherein the settlement logic failed to update the `filled` field on the order metadata.

### Testing
- Unit tests pass
- @sehyunc will test frontend integration

### DIFF
--- a/workers/task-driver/src/helpers.rs
+++ b/workers/task-driver/src/helpers.rs
@@ -11,7 +11,7 @@ use circuit_types::{
     },
     note::Note,
     order::Order,
-    r#match::OrderSettlementIndices,
+    r#match::{MatchResult, OrderSettlementIndices},
     SizedWallet,
 };
 use circuits::zk_circuits::{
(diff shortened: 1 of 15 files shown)


workers/task-driver/src/helpers.rs
```diff
@@ -76,12 +76,22 @@ pub fn transition_order_settling(order_id: OrderIdentifier, state: &State) -> Re
 }
 
 /// Update an order's state to `Filled`
-pub fn transition_order_filled(order_id: OrderIdentifier, state: &State) -> Result<(), String> {
+pub fn transition_order_filled(
```
akirillo, line 79: nit: maybe we should rename this / fix the doc comment, it implies that we're definitely marking the order as filled. perhaps `record_order_fill` or something?

akirillo (approved):

PR #543, opened 2024-04-27
chore: pin serde to 1.0.197 for contracts compatibility

akirillo:
This PR pins the version of `serde` used throughout the relayer to `1.0.197`, to ensure it is the same version used by all of the deployed contracts. This doesn't limit current usage of serde functionality and seems to be causing invalid signature errors in the contracts.

### DIFF
--- a/circuit-types/Cargo.toml
+++ b/circuit-types/Cargo.toml
@@ -36,7 +36,7 @@ renegade-crypto = { path = "../renegade-crypto" }
 byteorder = "1.5"
 itertools = "0.10"
 lazy_static = "1.4"
-serde = { version = "1.0.139", features = ["serde_derive"] }
+serde = { workspace = true, features = ["serde_derive"] }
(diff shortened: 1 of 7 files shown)


circuit-types/Cargo.toml
```diff
@@ -36,7 +36,7 @@ renegade-crypto = { path = "../renegade-crypto" }
 byteorder = "1.5"
 itertools = "0.10"
 lazy_static = "1.4"
-serde = { version = "1.0.139", features = ["serde_derive"] }
+serde = { version = "=1.0.197", features = ["serde_derive"] }
```
joeykraut, line 39: Let's just inherit the workspace version in all these crates

joeykraut (approved):

circuit-types/Cargo.toml
akirillo: ack

PR #544, opened 2024-04-28
state: applicator: task-queue: Append to task history on queue pop

joeykraut:
### Purpose
This PR handles the state transition logic to update task history; specifically by placing the popped task in the history when the queue is popped. This PR also passes success/failure information into the pop state transition so that the task history may store which tasks were successful vs not.

### Testing
- Unit tests pass

### DIFF
--- a/state/src/applicator/mod.rs
+++ b/state/src/applicator/mod.rs
@@ -76,7 +76,7 @@ impl StateApplicator {
             },
             StateTransition::UpdateOrderMetadata { meta } => self.update_order_metadata(meta),
             StateTransition::AppendTask { task } => self.append_task(&task),
-            StateTransition::PopTask { task_id } => self.pop_task(task_id),
+            StateTransition::PopTask { task_id, success } => self.pop_task(task_id, success),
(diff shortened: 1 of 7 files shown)


state/src/applicator/mod.rs
```diff
@@ -76,7 +76,7 @@ impl StateApplicator {
             },
             StateTransition::UpdateOrderMetadata { meta } => self.update_order_metadata(meta),
             StateTransition::AppendTask { task } => self.append_task(&task),
-            StateTransition::PopTask { task_id } => self.pop_task(task_id),
+            StateTransition::PopTask { task_id, success } => self.pop_task(task_id, success),
```
akirillo, line 79: this lgtm, but down the line should we just include the `QueuedTaskState` directly instead of a bool? would be more extensible if we put more states in that enum

akirillo (approved):

state/src/applicator/mod.rs
joeykraut: Yeah that's a good suggestion, i'll make that change

PR #547, opened 2024-04-29
api-server: websocket: task: Add task history websocket handler

joeykraut:
### Purpose
This PR adds a websocket handler for the task history endpoint. This pushes updates about historical tasks as well as active tasks and their state. In this way, this endpoint is a superset of the current task status websocket.

### Testing
- Unit tests pass
- Tested websockets locally

### DIFF
--- a/state/src/applicator/task_queue.rs
+++ b/state/src/applicator/task_queue.rs
@@ -5,8 +5,9 @@ use common::types::{
     wallet::WalletIdentifier,
 };
 use external_api::{
-    bus_message::{task_topic, SystemBusMessage},
+    bus_message::{task_history_topic, task_topic, SystemBusMessage},
     http::task::ApiTaskStatus,
+    types::ApiHistoricalTask,
 };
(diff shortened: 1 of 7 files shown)


state/src/applicator/task_queue.rs
```diff
@@ -162,7 +160,7 @@ impl StateApplicator {
         // was paused
         tx.resume_task_queue(&key)?;
         let task = tx.pop_task(&key)?.expect("expected preemptive task");
-        if let Some(mut t) = HistoricalTask::from_queued_task(task, key) {
+        if let Some(mut t) = HistoricalTask::from_queued_task(task.clone(), key) {
```
akirillo, line 163: similar to `pop_task`, should we mutate the task state outside of this conditional so that the updated state is published to the system bus correctly?

akirillo (approved):

state/src/applicator/task_queue.rs
joeykraut: yes, good catch, done

PR #548, opened 2024-04-29
arbitrum-client: Use `NonceManager` ethers middleware

joeykraut:
### Purpose
This PR updates the `arbitrum-client` to use the `ethers` `NonceManager` middleware. This allows the client to submit multiple concurrent transactions in a nonce-consistent way, rather than relying on a nonce-fetch which may be out of date with respect to a transaction yet to be executed.

### Testing
- Unit tests pass
- Replicated the out of date nonce bug on a local relayer targeting the staging sequencer. Verified that the bug was fixed after this change


akirillo (approved):
big hype for this

PR #550, opened 2024-04-29
state: storage: order-book: Delete orders when nullified

joeykraut:
### Purpose
This PR explicitly deletes orders after their wallets have been nullified. Keeping these orders around after they've been nullified only results in state bloat. Orders that are still valid will be re-gossiped by their managing cluster anyways.

### Testing
- Unit tests pass
- Tested submitting and cancelling orders in a wallet. Verified that the `/v0/order_book/orders` route reflected these changes

### DIFF
--- a/state/src/storage/tx/order_book.rs
+++ b/state/src/storage/tx/order_book.rs
@@ -205,25 +205,21 @@ impl<'db> StateTxn<'db, RW> {
         self.write_order(&order)
     }
 
-    /// Cancel an order in the order book
-    pub fn cancel_order(&self, order_id: &OrderIdentifier) -> Result<(), StorageError> {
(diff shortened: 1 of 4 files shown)


state/src/storage/tx/order_book.rs
```diff
@@ -205,25 +205,21 @@ impl<'db> StateTxn<'db, RW> {
         self.write_order(&order)
     }
 
-    /// Cancel an order in the order book
-    pub fn cancel_order(&self, order_id: &OrderIdentifier) -> Result<(), StorageError> {
-        let mut order = self.get_order_info_or_err(order_id)?;
-        order.state = NetworkOrderState::Cancelled;
```
akirillo, line 211: note to self: does this state need to exist anymore?

akirillo (approved):

PR #551, opened 2024-04-30
workers: price-reporter: exchange: per-exchange pair support API calls

akirillo:
This PR adds a new method, `supports_pair`, to the `ExchangeConnection` trait, which we use to check if a given exchange supports a given pair by making the requisite API call for that exchange.

We use this in the `connect` methods of the `ExchangeConnection` implementors to error on connecting to a price stream for a pair that an exchange doesn't support. This change is absorbed by the price reporter, as well, since it invokes the same method.

We also expose a helper method for the external price reporter to use in its own validation logic before trying to `connect`.

We don't use this method for any pre-connection checks in the relayer, since the implementation of price conversion through "default" stables for each exchange (paired with the assumption that quote tokens are stables) lets us assume that each exchange has a market between any listed base token and the default stable, an assumption we have already been leveraging.

### DIFF
--- a/workers/price-reporter/src/exchange/connection.rs
+++ b/workers/price-reporter/src/exchange/connection.rs
@@ -142,10 +142,16 @@ pub trait ExchangeConnection: Stream<Item = PriceStreamType> + Unpin + Send {
     ) -> Result<Self, ExchangeConnectionError>
     where
         Self: Sized;
+
     /// Send a keepalive signal on the connection if necessary
     async fn send_keepalive(&mut self) -> Result<(), ExchangeConnectionError> {
         Ok(())
     }
+
+    /// Check whether the exchange supports the given pair
+    async fn supports_pair(base_token: &Token, quote_token: &Token) -> Result<bool, ExchangeConnectionError>
+    where
+        Self: Sized;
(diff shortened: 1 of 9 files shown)


workers/price-reporter/src/exchange/connection.rs
```diff
@@ -142,10 +142,19 @@ pub trait ExchangeConnection: Stream<Item = PriceStreamType> + Unpin + Send {
...
         Ok(())
     }
+
+    /// Check whether the exchange supports the given pair
+    async fn supports_pair(
+        base_token: &Token,
+        quote_token: &Token,
+    ) -> Result<bool, ExchangeConnectionError>
+    where
+        Self: Sized;
```
joeykraut, line 157: nit: might be worth just floating the `Sized` trait bound up to the trait signature now that it's on a few methods

joeykraut (approved):
clean

workers/price-reporter/src/exchange/connection.rs
akirillo: ah, this makes it impossible for the trait to be object-safe, which we need

PR #554, opened 2024-04-30
state: replication: create `Worker` for Raft replication node

akirillo:
This PR implements the `Worker` trait for the `ReplicationNode` so that it can gracefully handle failures on the Raft thread. We already find ourselves needing to implement this and the most sensible approach is to hook into the worker management patterns that are already present.

All the unit tests pass, and I also tested locally that the node appropriately shuts down when the Raft thread panics (which we see occasionally in `dev`).

### DIFF
--- a/state/src/interface/mod.rs
+++ b/state/src/interface/mod.rs
@@ -14,12 +14,10 @@ pub mod wallet_index;
 
 use std::sync::{Arc, RwLock};
 
+use common::default_wrapper::default_option;
 use config::RelayerConfig;
 use external_api::bus_message::SystemBusMessage;
-use job_types::{
-    handshake_manager::HandshakeManagerQueue, network_manager::NetworkManagerQueue,
-    task_driver::TaskDriverQueue,
-};
+use job_types::network_manager::NetworkManagerQueue;
 use system_bus::SystemBus;
 use util::err_str;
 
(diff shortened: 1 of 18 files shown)


state/src/interface/mod.rs
```diff
@@ -179,24 +96,36 @@ impl State {
...
+        &self,
+        replication_config: &mut ReplicationNodeConfig<GossipRaftNetwork>,
+        network_outbound: NetworkManagerQueue,
+        raft_inbound: RaftMessageReceiver,
+    ) {
+        let network =
+            GossipRaftNetwork::new(network_outbound, raft_inbound, self.translation_map.clone());
+
+        self.fill_replication_config_with_network(replication_config, network)
+    }
```
joeykraut, line 112: This business of creating one object, using it to edit a config, then creating another is pretty messy in my opinion. This is a fine pattern if we internalize this to the `state` crate. So I'd suggest we publish a helper to start the raft worker and return a state instance. Essentially what is done in the `core` crate, but defined here to encapsulate the coupling.

This will make it worlds easier for tests to interact with state.

akirillo: after thinking about it somewhere, I think it would be cleaner to just have `State::new` return the handles to the Raft worker & cancel channel than go through all this back and forth. I can rename the method so it doesn't imply that it only returns a `State` object.

joeykraut (approved):

state/src/interface/mod.rs
```diff
@@ -79,23 +76,24 @@ pub struct State {
 }
 
 impl State {
-    /// Create a new state handle using a `GossipRaftNetwork`
-    pub fn new(
+    /// Spawn a new state object using a `GossipRaftNetwork`, returning the
+    /// handles to the state, Raft worker, and Raft cancel sender
+    pub fn spawn(
```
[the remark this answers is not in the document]
akirillo: ah sure, i thought `new` might be confusing because it returns multiple objects, but happy to keep that

PR #558, opened 2024-05-02
api-server: websocket: Do not trace websocket `handle_connection`

joeykraut:
### Purpose
This PR removes the `handle_connection` traces in the relayer's websocket server and downgrades connection errors to `info`. The `handle_connection` ends up creating many long running traces that have no detailed information. As well, if a client hangs up, we report an error logs which distracts from actual errors in traces.


akirillo (approved):
this was long overdue, thank you

PR #559, opened 2024-05-05
docker: release: add bootloader script

akirillo:
This PR adds a basic "bootloader" script to the Docker image which fetches the node's config file from S3, and fills it in with values set via environment variables.


joeykraut (approved):
nice, has this been tested?

akirillo:
> nice, has this been tested?

yes, this has been running in dev, nodes spin up without issue. have yet to test a proper config change and ensure it is ingested on container restart

PR #565, opened 2024-05-08
state: replicationv2: state-machine: Implement state machine

joeykraut:
### Purpose
This PR implements the `RaftStateMachine` trait for the new consensus engine. Additionally this PR fixes a few bugs with storage that were discovered while integrating.

Not all tests pass, snapshotting needs to be implemented before the `openraft` suite passes

### Testing
- Non-state tests pass

### DIFF
--- a/state/src/storage/tx/mod.rs
+++ b/state/src/storage/tx/mod.rs
@@ -26,7 +26,7 @@ use crate::{
     TASK_HISTORY_TABLE, TASK_QUEUE_TABLE, TASK_TO_KEY_TABLE, WALLETS_TABLE,
 };
 
-use self::raft_log::RAFT_METADATA_TABLE;
+use self::raft_log::{RAFT_LOGS_TABLE, RAFT_METADATA_TABLE};
 
 use super::{
     cursor::DbCursor,
@@ -107,6 +107,7 @@ impl<'db> StateTxn<'db, RW> {
             MPC_PREPROCESSING_TABLE,
             NODE_METADATA_TABLE,
             RAFT_METADATA_TABLE,
+            RAFT_LOGS_TABLE,
         ]
         .iter()
         {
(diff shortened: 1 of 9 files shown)


state/src/storage/tx/mod.rs
```diff
@@ -107,6 +107,7 @@ impl<'db> StateTxn<'db, RW> {
             MPC_PREPROCESSING_TABLE,
             NODE_METADATA_TABLE,
             RAFT_METADATA_TABLE,
+            RAFT_LOGS_TABLE,
```
akirillo, line 110: do we need to increment `NUM_TABLES`?

state/src/storage/tx/raft_log.rs
```diff
@@ -233,13 +234,11 @@ impl<'db> StateTxn<'db, RW> {
...
-        let first_log = self.first_raft_log_index()?.unwrap_or(0);
-
-        // Purge the logs
         let mut log_cursor = self.logs_cursor()?;
+
+        let idx_lsn = lsn_to_key(index);
         log_cursor.seek(&idx_lsn)?;
-        for _ in first_log..=index {
+        while log_cursor.has_current()? {
             log_cursor.delete()?;
```
akirillo, line 242: just want to confirm that `log_cursor.delete()` will not progress the cursor's position, right?

akirillo (approved):

state/src/storage/tx/mod.rs
joeykraut: no, this was already created elsewhere i added it here for automatic creation

state/src/storage/tx/raft_log.rs
joeykraut: it _does_ move the cursor, this is the desired behavior and there is a test for this

PR #566, opened 2024-05-08
state: replicationv2: snapshot: Implement snapshot procedure

joeykraut:
### Purpose
This PR implements the procedure for snapshotting relayer state. This proceeds as follows:
1. A read tx is initiated to read-lock the DB
2. The DB's data file is copied to the snapshot location
3. gzip the data file 

### Todo
- Support excluding tables from the snapshot

### Testing
- Unit tests pass
- Tested snapshot and recovery

### DIFF
--- a/state/src/replicationv2/snapshot.rs
+++ b/state/src/replicationv2/snapshot.rs
@@ -1,0 +1,131 @@
+//! Raft snapshots implementation
+use std::fs;
+use std::fs::File;
+use std::io::BufReader;
+use std::path::PathBuf;
+
+use flate2::write::GzEncoder;
+use flate2::Compression;
+use util::err_str;
+
+use crate::storage::{db::DB, tx::raft_log::RAFT_LOGS_TABLE};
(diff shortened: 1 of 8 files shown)


state/src/replicationv2/snapshot.rs
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    /// Tests snapshotting and recovering a DB from a snapshot
+    #[tokio::test]
+    async fn test_snapshot() {
+        const TABLE: &str = "test-table";
+        let (k, v) = ("key".to_string(), "value".to_string());
+
+        let db_path = tmp_db_path();
+        let snap_path = db_path.clone();
```
akirillo, line 114: it looks like we save the snapshot to the same location as the db path, this feels like unintended usage - perhaps we make `snap_path` a different temp file?

akirillo (approved):

state/src/replicationv2/snapshot.rs
joeykraut: for tests its fine, wanted to avoid having to create new directories. it should not matter whether it is the same directory

PR #568, opened 2024-05-08
state: replicationv2: snapshot: Build full snapshot from zip file

joeykraut:
### Purpose
This PR sets up the full snapshot reported to the consensus engine by the `StateMachine`. This includes the snapshot metadata as well as the archive file backing the snapshot. In the current `openraft` implementation, the consensus engine will actually immediately drop the snapshot data, closing the file, and only re-request it when sending the snapshot to a follower.

I also internalized this logic to the `StateMachine` to give a cleaner dependency access pattern.

### Testing
- Unit tests pass


state/src/replication/log_store.rs
```diff
@@ -268,104 +268,4 @@ mod test {
...
         assert_eq!(snap_res.get_metadata(), snap.get_metadata());
     }
-
-    // -------------------
-    // | Log Entry Tests |
-    // -------------------
-
-    /// Tests fetching the first and last entries from an empty log
-    #[test]
-    fn test_empty_log() {
```
joeykraut, line 278: old code that now errors

akirillo (approved):

PR #569, opened 2024-05-08
workers: task-driver: tasks: update_wallet: check reblind progression

akirillo:
This PR adds a check in the creation of an `UpdateWalletTask` which ensures that the new wallet is "one reblind on top of" the old wallet. It also tweaks the error handling of spawning tasks to notify listeners if creating a task fails (not just running it).

I tested this by running a local relayer and asserting that a valid deposit passes this check, and that a doubly-reblinded deposit does not.

### DIFF
--- a/workers/task-driver/src/driver.rs
+++ b/workers/task-driver/src/driver.rs
@@ -397,14 +397,18 @@ impl TaskExecutor {
         notifications: TaskNotificationMap,
     ) -> Result<(), TaskDriverError> {
         // Create the task
-        let mut task = RunnableTask::<T>::from_descriptor(immediate, id, descriptor, ctx).await?;
+        let combined_res =
+            match RunnableTask::<T>::from_descriptor(immediate, id, descriptor, ctx).await {
+                Ok(mut task) => {
+                    // Run the task
+                    let res = Self::run_task_to_completion(&mut task, args).await;
 
-        // Run the task
-        let res = Self::run_task_to_completion(&mut task, args).await;
-
-        // Cleanup
-        let cleanup_res = task.cleanup(res.is_ok()).await;
-        let combined_res = res.and(cleanup_res);
+                    // Cleanup
+                    let cleanup_res = task.cleanup(res.is_ok()).await;
+                    res.and(cleanup_res)
+                },
+                Err(e) => Err(e),
+            };
 
         // Notify any listeners that the task has completed
         let str_res = combined_res.clone().map_err(|e| e.to_string());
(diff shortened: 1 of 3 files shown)


common/src/types/wallet/shares.rs
```diff
@@ -29,6 +29,17 @@ impl Wallet {
         circuit_wallet == recovered_wallet
     }
 
+    /// Check that the wallet's blinder and private shares are the result of
+    /// applying a reblind to the old wallet
+    pub fn check_reblind_progression(&self, old_wallet: &Wallet) -> bool {
```
joeykraut, line 34: i don't think this should be an instance method of the `Wallet` type, the wallet isn't really checking anything. Let's move this to the descriptor, see my second comment

workers/task-driver/src/driver.rs
```diff
@@ -397,14 +397,18 @@ impl TaskExecutor {
...
         // Create the task
-        let mut task = RunnableTask::<T>::from_descriptor(immediate, id, descriptor, ctx).await?;
-
-        // Run the task
-        let res = Self::run_task_to_completion(&mut task, args).await;
-
-        // Cleanup
-        let cleanup_res = task.cleanup(res.is_ok()).await;
-        let combined_res = res.and(cleanup_res);
+        let combined_res =
```
joeykraut, line 400: disagree with this approach, let's validate this at the API layer. The whole point of this change is to give safety to events created through the API, this is too indirect of a way of returning an error to the API. Think about it from the client's perspective; the http task succeeds but they have to poll the task status to figure out whether their request was valid -- that's just a poor interface.

I'd suggest adding a safety check in the task descriptor so that the http call itself fails with `400` when we incorrectly reblind

joeykraut (changes requested):

workers/task-driver/src/driver.rs
joeykraut: And when validating at the API layer we should use the back of queue wallet, it should be available already via a helper at the API layer

akirillo: The reason for this approach is to have certainty that we're checking a reblind progression on the exactly correct wallet, without having to make assumptions about the correctness of the wallet simulation functionality.

At the API layer we already apply reblinds on top of the back of queue wallet, so doing this check there would only tell us if we forgot to call `reblind` before invoking the descriptor constructor. However one of the issues we saw is that this did not accurately represent the state of the wallet after the queue had processed.

I agree that this is somewhat confusing from the client's perspective but it should be able to see that the task failed, this is something it has to handle regardless.

joeykraut: talked off line, orienting this around preventing a footgun is more important than letting the client know early of a relayer mistake. approved

joeykraut (approved):

PR #570, opened 2024-05-09
state: replicationv2: snapshot: Apply snapshot to state machine

joeykraut:
### Purpose
This PR implements the logic to update a `StateMachine` from a snapshot that has finished being received. We do so by opening a cursor on each table and copying out of the cursor exhaustively.

I also restructured some of how DB tables are allocated to make the table index more explicit and centrally defined. 

### Testing
- Tested DB copy operation and snapshot application
- Unit tests pass generally
- Deferring passing the `openraft` suite to a follow-up

### DIFF
--- a/state/src/replicationv2/snapshot.rs
+++ b/state/src/replicationv2/snapshot.rs
@@ -4,6 +4,7 @@ use std::fs::File;
 use std::io::BufReader;
 use std::path::PathBuf;
 
+use flate2::read::GzDecoder;
 use flate2::write::GzEncoder;
 use flate2::Compression;
 use openraft::{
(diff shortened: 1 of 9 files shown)


state/src/replicationv2/snapshot.rs
```diff
@@ -157,53 +171,118 @@ impl StateMachine {
...
+    }
+
+    /// Update the local DB using a snapshot DB
+    pub async fn update_from_snapshot(
+        &mut self,
+        meta: &SnapshotMeta<NodeId, Node>,
+        snapshot_db: DB,
+    ) -> Result<(), ReplicationV2Error> {
+        self.last_applied_log = meta.last_log_id;
+        self.last_membership = meta.last_membership.clone();
```
akirillo, line 218: do we want to carry over membership from the snapshot? I could see this changing drastically between when a snapshot was taken and when it is applied

akirillo: ah, nvm - this is just the `last_membership` metadata right, not the nodes actual view of the membership (which is in an excluded table)?

akirillo (approved):

state/src/replicationv2/snapshot.rs
joeykraut: yeah the last membership change the node is aware of. we do want this; a leader will only send a snapshot if a node is far enough behind to not replicate logs directly. So yes, it may change a lot but only to a more recent value.

PR #571, opened 2024-05-09
state: replicationv2: Fix misc bugs to pass `openraft` test suite

joeykraut:
### Purpose
This PR implements various bug fixes to get the `openraft` test suite passing on our implementation of the log store and state machine.

### Testing
- `openraft` suite passes

### DIFF
--- a/state/src/storage/tx/raft_log.rs
+++ b/state/src/storage/tx/raft_log.rs
@@ -230,11 +230,19 @@ impl<'db> StateTxn<'db, RW> {
     /// Purge all logs before the given index (inclusive)
     pub fn purge_logs(&self, index: u64) -> Result<(), StorageError> {
         let mut log_cursor = self.logs_cursor()?;
+        log_cursor.seek_first()?;
 
-        let idx_lsn = lsn_to_key(index);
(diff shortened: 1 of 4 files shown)


state/src/storage/tx/raft_log.rs
```diff
@@ -230,11 +230,19 @@ impl<'db> StateTxn<'db, RW> {
     /// Purge all logs before the given index (inclusive)
     pub fn purge_logs(&self, index: u64) -> Result<(), StorageError> {
         let mut log_cursor = self.logs_cursor()?;
-
-        let idx_lsn = lsn_to_key(index);
-        log_cursor.seek(&idx_lsn)?;
-        while log_cursor.has_current()? {
-            log_cursor.delete()?;
```
akirillo, line 237: I see, was the issue here that when deleting, the cursor's position is changed (whereas we thought it would stay the same)?

akirillo (approved):

state/src/storage/tx/raft_log.rs
joeykraut: no delete was implemented correctly. The issue is that the method deletes the wrong set of values

PR #574, opened 2024-05-10
state: replicationv2: Implement mock network and initial raft tests

joeykraut:
### Purpose
This PR adds an implementation of a mock local network and defines the initial tests for the new consensus engine.

### Todo
- Connect network to gossip layer

### Testing
- Unit tests pass

### DIFF
--- a/state/src/replicationv2/mod.rs
+++ b/state/src/replicationv2/mod.rs
@@ -9,7 +9,7 @@ mod network;
 mod snapshot;
 mod state_machine;
 
-use openraft::{EmptyNode, RaftTypeConfig};
+use openraft::{EmptyNode, Raft as RaftInner, RaftTypeConfig};
 
 use crate::StateTransition;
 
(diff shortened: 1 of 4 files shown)


state/src/replicationv2/mod.rs
```diff
@@ -81,4 +230,48 @@ mod test {
...
+    #[tokio::test]
+    async fn test_raft_singleton() {
+        let node = MockRaft::create_singleton_raft().await;
+        let wallet = mock_empty_wallet();
+
+        // Propose a new wallet to the raft
+        let id = wallet.wallet_id;
+        node.get_raft()
+            .client_write(Box::new(StateTransition::AddWallet { wallet }))
+            .await
```
akirillo, line 243: this is nice, we can remove `ProposalWaiter`s then, right?

akirillo (approved):

state/src/replicationv2/mod.rs
joeykraut: Not quite, going to add back in proposal waiters afterward to allow for fire and forget semantics if a client doesn't want to await a transition

PR #577, opened 2024-05-11
workers: proof-manager: worker: increase stack size & frontload circuit preprocessing

akirillo:
This PR makes the following changes to the proof manager:
1. Name all threads (why not)
2. Increase the thread stack size to 10MB (default is 2MB)
3. Preprocess all the circuit keys & layouts when the proof manager starts up

Empirically (when testing on an EC2 instance w/ a dev build) this no longer triggers the stack overflows we were seeing in the dev environment, which seemed to be coming from the proof manager, typically when generating the preprocessed circuit keys.

The way the orderbots/quoters all hit the relayer at ~the same time, it's possible there was redundant work being done in generating the keys - this wouldn't directly explain the stack overflow, but regardless it is helpful to have this be done at startup.

We are no longer seeing stack overflows in dev.

### DIFF
--- a/workers/proof-manager/src/worker.rs
+++ b/workers/proof-manager/src/worker.rs
@@ -40,8 +40,6 @@ impl Worker for ProofManager {
     where
         Self: Sized,
     {
-        // TODO: Initialize pkeys/vkeys for all circuits
-
         // Build a thread pool for the worker
         let proof_generation_thread_pool = ThreadPoolBuilder::new()
             .thread_name(|i| format!("{}-{}", WORKER_THREAD_PREFIX, i))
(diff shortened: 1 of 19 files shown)


workers/proof-manager/src/worker.rs
```diff
@@ -36,8 +40,12 @@ impl Worker for ProofManager {
     where
         Self: Sized,
     {
+        ProofManager::preprocess_circuits()?;
```
joeykraut, line 43: propose moving this to the start of the execution loop, perhaps before we start receiving jobs. Problem with putting this in `new` is we now explicitly block the main thread which is probably not desirable

joeykraut (approved):
one thing i think we should take care of, but o/w lgtm

workers/proof-manager/src/worker.rs
akirillo: sure thing, good call out

PR #578, opened 2024-05-11
state: replicationv2: raft: Node join and leave

joeykraut:
### Purpose
This PR supports node membership updates in the raft client interface and adds tests for joining and leaving clusters, scaling up and down.

### Todo
- Proposal waiters
- Connection to gossip network
- Connect raft client to state interface

### Testing
- All unit tests pass

### DIFF
--- a/state/src/replicationv2/mod.rs
+++ b/state/src/replicationv2/mod.rs
@@ -35,20 +35,23 @@ pub type SnapshotData = <TypeConfig as RaftTypeConfig>::SnapshotData;
 /// A raft using our type config
 pub type Raft = RaftInner<TypeConfig>;
 
+// -------------
+// | Mock Raft |
+// -------------
+
 #[cfg(test)]
 pub mod test_helpers {
-    use std::{sync::Arc, time::Duration};
+    use std::{collections::HashMap, sync::Arc, time::Duration};
 
-    use futures::stream::StreamExt;
+    use common::{new_async_shared, AsyncShared};
     use itertools::Itertools;
-    use tokio_stream::{wrappers::UnboundedReceiverStream, StreamMap};
 
     use crate::{applicator::test_helpers::mock_applicator, storage::db::DB};
 
     use super::{
         log_store::LogStore,
         network::{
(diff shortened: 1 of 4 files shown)


state/src/replicationv2/mod.rs
```diff
@@ -124,61 +127,69 @@ pub mod test_helpers {
...
         }
 
         /// Create a mock raft network with `n_nodes` nodes and return the rafts
         /// in use
-        pub async fn create_raft(n_nodes: usize) -> Vec<MockRaftNode> {
-            let mut receivers = Vec::with_capacity(n_nodes);
-            let mut nodes = Vec::with_capacity(n_nodes);
+        ///
+        /// `init` indicates whether to initialize the cluster with knowledge of
+        /// other nodes
```
akirillo, line 156: nit: we should prob remove this as we don't accept an `init` argument

state/src/replicationv2/mod.rs
```diff
@@ -273,51 +311,265 @@ mod test {
...
+        tokio::time::sleep(Duration::from_millis(100)).await;
+
+        // Check the DB of the new node
+        let db = raft.get_db(new_nid).await;
+        let tx = db.new_read_tx().unwrap();
         let wallet = tx.get_wallet(&wallet_id).unwrap();
         assert!(wallet.is_some());
+
+        // Promote the learner, propose a new wallet, ensure consensus is reached
+        client.promote_learner(new_nid).await.unwrap();
```
akirillo, line 394: is it important for the new learner to be promoted for these tests? it should receive new proposals regardless, right? which leads me to the question: in these join/leave tests, would it be helpful to test that after promotion, the new node can successfully make new proposals of its own?

akirillo (approved):

state/src/replicationv2/mod.rs
joeykraut: To your first question: it does not matter whether the node is a learner or voter for this test

To your second: Technically only the leader can create proposals. We just do the extra step of having each client forward proposals to the leader. So it doesn't matter whether the node is a follower or learner here either.

PR #580, opened 2024-05-12
state: replicationv2: Add proposal notification map

joeykraut:
### Purpose
This PR implements the emission of proposal notification from the state machine when a proposal is applied. These notifications will be setup in the interface methods to allow clients on the interface to await proposals.

### Testing
- Tested awaiting a proposal on the state machine
- Unit tests pass


state/src/replicationv2/mod.rs
```diff
@@ -12,13 +12,13 @@ mod state_machine;
...
 
-use crate::StateTransition;
+use crate::Proposal;
 
 // Declare the types config for the raft
 openraft::declare_raft_types! (
     /// The type config for the raft
     pub TypeConfig:
-        D = Box<StateTransition>,
+        D = Proposal,
```
akirillo, line 21: I'm assuming we updated the `Proposal` struct to contain a `Box<StateTransition>` instead of just a `StateTransition` and that's just not in this PR?

akirillo (approved):

state/src/replicationv2/mod.rs
joeykraut: yeah the rest of it's file is in an intermediate refactor state so I left it out

PR #581, opened 2024-05-12
state: interface: Connect state interface to raft client

joeykraut:
### Purpose
This PR sets up the connection between the state interface and the raft client. This involves spawning a raft inside the state constructor and tying together the notifications queue, dbs, state machine, etc.

We no longer run the raft as a separate worker, this is handled by `openraft`

### Testing
- Unit tests pass
- Tested some basic state transitions through the interface

### DIFF
--- a/state/src/interface/mod.rs
+++ b/state/src/interface/mod.rs
@@ -2,58 +2,44 @@
 //! proposing state transitions and reading from state
 
 pub mod error;
-pub mod mpc_preprocessing;
+// pub mod mpc_preprocessing;
 pub mod node_metadata;
-pub mod notifications;
-pub mod order_book;
-pub mod order_history;
-pub mod peer_index;
-pub mod raft;
-pub mod task_queue;
+// pub mod order_book;
+// pub mod order_history;
+// pub mod peer_index;
+// pub mod raft;
+// pub mod task_queue;
 pub mod wallet_index;
 
-use std::sync::{Arc, RwLock};
+use std::sync::Arc;
 
-use ::raft::prelude::Config as RaftConfig;
-use common::{
-    types::{gossip::WrappedPeerId, new_cancel_channel},
-    worker::Worker,
-};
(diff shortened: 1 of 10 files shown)


state/src/interface/mod.rs
```diff
@@ -147,63 +84,60 @@ impl State {
...
+        let notifications = OpenNotifications::new();
+        let sm_config = StateMachineConfig::new(config.raft_snapshot_path.clone());
+        let sm = StateMachine::new(sm_config, notifications.clone(), applicator);
 
-        // Start the Raft worker
-        let mut raft_worker = ReplicationNodeWorker::new(replication_config)
-            .expect("failed to build Raft replication node");
-        raft_worker.start().expect("failed to start Raft replication node");
+        // Start a raft
+        let raft = RaftClient::new(raft_config, db.clone(), network, sm)
```
akirillo, line 102: If the `openraft` thread ever fails/panics, how do we propagate this to the node?

state/src/lib.rs
```diff
@@ -22,19 +22,21 @@ use common::types::{
     tasks::{QueuedTask, QueuedTaskState, TaskIdentifier, TaskQueueKey},
     wallet::{order_metadata::OrderMetadata, OrderIdentifier, Wallet},
 };
-use interface::notifications::ProposalResultSender;
-use replication::RaftPeerId;
+use notifications::ProposalId;
+use replicationv2::NodeId;
+// use replication::RaftPeerId;
```
akirillo, line 27: dead code?

akirillo (approved):

state/src/interface/mod.rs
joeykraut: Not much we can do, caller would panic given that it's async. ITT we should observe it and see if that's a problem, panics in libraries should be rare and I'm wary of going out of our way to detect panics unless we believe it's a problem

PR #582, opened 2024-05-12
state: interface: Connect interface to new raft

joeykraut:
### Purpose
This PR migrates the rest of the state interfaces to the new consensus engine and deletes some now-unused code. The full cleanup will be done separately once the rest of testing is done.

### Testing
- All unit tests pass


state/src/interface/raft.rs
akirillo, line 1: this is lower prio but since we never actually use the `PeerIdTranslationMap` as a map anymore we can just call `fxhash(peer_id)` directly where we need to

akirillo (approved):

state/src/interface/raft.rs
joeykraut: Did this in a follow up

PR #583, opened 2024-05-13
state: interface: Use async state interface

joeykraut:
### Purpose
This PR changes the interface of the state to use async methods all around. Callers of the state interface are all async, so we wish to avoid blocking their worker threads unnecessarily -- as mdbx may do.

Instead, we spawn all state access methods on blocking tasks that manage a read or write transaction, and await this blocking task's completion. 

### Testing
- All unit tests pass for the state library

### DIFF
--- a/state/src/interface/peer_index.rs
+++ b/state/src/interface/peer_index.rs
@@ -11,11 +11,12 @@ use common::types::gossip::{ClusterId, PeerInfo, WrappedPeerId};
 use external_api::bus_message::{SystemBusMessage, NETWORK_TOPOLOGY_TOPIC};
 use gossip_api::request_response::heartbeat::HeartbeatMessage;
 use itertools::Itertools;
+use util::runtime::block_current;
 
 use crate::{
     error::StateError,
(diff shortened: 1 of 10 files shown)


state/src/interface/peer_index.rs
```diff
@@ -125,70 +138,76 @@ impl<N: NetworkEssential> StateHandle<N> {
...
+
+                // Add the peer to the store
+                tx.write_peer(&peer)?;
+                tx.add_to_cluster(&peer.peer_id, &peer.cluster_id)?;
+
+                // If the peer belongs in the same cluster, add it to the raft group
+                let my_cluster_id = tx.get_cluster_id()?;
+                if peer.cluster_id == my_cluster_id {
+                    let raft_id = PeerIdTranslationMap::get_raft_id(&peer.peer_id);
+                    block_current(this.raft.add_learner(raft_id))?;
```
akirillo, line 160: unable to find `block_current` in this PR, if the idea is that it blocks the current runtime until this task is complete, is that so that any other concurrent proposals are sent to the right set of peers?

akirillo (approved):

PR #584, opened 2024-05-13
state: replicationv2: network: Add gossip network implementation

joeykraut:
### Purpose
This PR sets up the connection between the gossip network and the new consensus engine. This is done by giving the network implementation queue access to the network manager. The network shim forwards a request to the queue with a response channel on which it expects the corresponding `RaftResponse`.

### Todo
- Handle these message types in the network manager

### Testing
- Unit tests pass in `state`


state/src/replicationv2/network/gossip.rs
```diff
@@ -0,0 +1,86 @@
...
+};
+
+use super::{P2PRaftNetwork, P2PRaftNetworkWrapper, RaftRequest, RaftResponse};
+
+/// The error message emitted when a response type is invalid
+const ERR_INVALID_RESPONSE: &str = "invalid response type from raft peer";
+
+/// The network shim
+#[derive(Clone)]
+pub struct GossipNetwork {
```
akirillo, line 23: I understand it's a bit out of scope for our codebase, but could we have some quick documentation about how this struct is managed via the `RaftNetworkFactory` trait? IIUC every time there's a membership change, new `GossipNetwork` instances will be spawned for each node in the membership via `RaftNetworkFactory::new_client`. This is initially unclear w/o looking at the openraft guide and I was a bit confused as to why the whole `GossipNetwork` only has one `target`

akirillo (approved):

PR #585, opened 2024-05-13
workers: network-manager: Multithread task handlers

joeykraut:
### Purpose
We spawn network manager tasks on a tokio runtime instead of using a single thread. This is complicated by `Swarm` not implementing `Sync`, so we broker access to the `Swarm` via a `behavior_rx` queue. This queue is drained and its events handled on the main thread to obviate synchronization on the `Swarm`.

This is done ahead of the raft request/response integration to avoid processing raft messages (or otherwise awaiting their processing) on the main thread.

### Testing
- Deferred until the relayer builds

### DIFF
--- a/workers/network-manager/src/executor.rs
+++ b/workers/network-manager/src/executor.rs
@@ -1,41 +1,38 @@
 //! The network manager handles lower level interaction with the p2p network
+mod behavior;
 mod control_directives;
 mod identify;
 mod pubsub;
 mod request_response;
 
 use common::{
-    default_wrapper::DefaultWrapper,
-    types::{
-        gossip::{ClusterId, PeerInfo, WrappedPeerId},
-        CancelChannel,
-    },
+    default_wrapper::{DefaultOption, DefaultWrapper},
+    new_async_shared,
+    types::{gossip::WrappedPeerId, CancelChannel},
+    AsyncShared,
 };
 use ed25519_dalek::Keypair as SigKeypair;
 use futures::StreamExt;
-use gossip_api::pubsub::{orderbook::ORDER_BOOK_TOPIC, PubsubMessage};
+use gossip_api::pubsub::PubsubMessage;
 use job_types::{
     gossip_server::GossipServerQueue,
     handshake_manager::HandshakeManagerQueue,
     network_manager::{NetworkManagerJob, NetworkManagerReceiver},
 };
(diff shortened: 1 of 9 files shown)


workers/network-manager/src/executor.rs
```diff
@@ -47,6 +44,8 @@ const ERR_BROKER_MPC_NET: &str = "failed to broker MPC network";
 /// The multiaddr protocol of the transport in libp2p
 const TRANSPORT_PROTOCOL_NAME: &str = "udp";
 
+/// The
```
akirillo, line 47: nit: dead code

workers/network-manager/src/executor.rs
```diff
@@ -149,31 +88,34 @@ struct BufferedPubsubMessage {
...
-    warmup_buffer: Vec<BufferedPubsubMessage>,
-    /// The underlying swarm that manages low level network behavior
-    swarm: Swarm<ComposedNetworkBehavior>,
-    /// The channel to receive outbound requests on from other workers
+    warmup_buffer: AsyncShared<Vec<BufferedPubsubMessage>>,
+    /// The behavior channel receiver, used to sequence access to the underlying
+    /// swarm
+    behavior_rx: DefaultOption<BehaviorReceiver>,
+    /// The behavior channel sender, used to send behaviors directly to the
+    /// swarm from within the network manager's job handlers The channel to
```
akirillo, line 112: nit: move "The channel to" at the end of this line to the line below

workers/network-manager/src/executor.rs
```diff
@@ -225,30 +166,46 @@ impl NetworkManagerExecutor {
...
-    pub async fn executor_loop(mut self) -> NetworkManagerError {
+    pub async fn executor_loop(
+        mut self,
+        mut swarm: Swarm<ComposedNetworkBehavior>,
+    ) -> NetworkManagerError {
         info!("Starting executor loop for network manager...");
         let mut cancel_channel = self.cancel.take().unwrap();
         let mut job_channel = self.job_channel.take().unwrap();
+        let mut behavior_channel = self.behavior_rx.take().unwrap();
+        // let mut swarm = self.swarm.take().unwrap();
```
akirillo, line 177: nit: dead code

akirillo (approved):

PR #587, opened 2024-05-14
meta: Fixup workspace to use new async state interface

joeykraut:
### Purpose
This PR fixes up various workers that call out to the state interface, which is now async.

The next PR will get the relayer building and working with basic raft functionality.

### Testing
- Most unit tests pass


workers/task-driver/src/driver.rs
```diff
@@ -281,121 +247,89 @@ impl TaskExecutor {
...
-        let args = self.runtime_config;
-        let task_notifications = self.task_notifications.clone();
-
-        Self::start_task(immediate, task_id, descriptor, ctx, args, task_notifications)
-    }
-
     /// Spawn a new task in the driver
     ///
     /// Returns the success of the task
     async fn start_task(
```
akirillo, line 282: can we re-attach the tracing that was lost from the `TaskDriverJob::Run` arm of the match statement in `run` to `start_task` here? Namely:
```suggestion
    #[instrument(name = "task", skip_all, err, fields(
        task_id = %id,
        task = %task.display_description(),
        queue_key = %task.queue_key(),
    )]
    async fn start_task(
```
this is by far the most important trace we produce. And applying it this way to `start_task` will also give us more descriptive traces for preemptive tasks, though I would also request that we add a similar `#[instrument]` attribute to `start_preemptive_task` so that we can trace any errors in pausing/unpausing the queue appropriately.

akirillo (approved):
need to add back some tracing but o/w lgtm

workers/task-driver/src/driver.rs
joeykraut: done

PR #588, opened 2024-05-15
state: replicationv2: network: Dynamically dispatch all network impls

joeykraut:
### Purpose
This PR changes the raft networking implementations to use trait objects internally. This is key to avoiding generics that color all encapsulating interfaces, including those beyond the `state` crate. This change also vastly simplifies the mocking and testing approach in the integration tests of other crates.

The performance implications should be minimal -- certainly dwarfed by the I/O bound nature of these traits' implementation. 

### Testing
- Unit tests pass

### DIFF
--- a/state/src/replicationv2/raft.rs
+++ b/state/src/replicationv2/raft.rs
@@ -10,7 +10,9 @@ use crate::{notifications::ProposalId, storage::db::DB, Proposal, StateTransitio
 use super::{
     error::ReplicationV2Error,
     log_store::LogStore,
-    network::{P2PRaftNetwork, RaftRequest, RaftResponse},
+    network::{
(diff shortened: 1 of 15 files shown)


state/src/replicationv2/raft.rs
```diff
@@ -136,6 +133,18 @@ impl<N: NetworkEssential> RaftClient<N> {
         self.raft.current_leader().await
     }
 
+    /// Get the node info an ID for the current leader
+    pub async fn leader_info(&self) -> Option<(NodeId, RaftNode)> {
```
akirillo, line 137: clippy complaining that this doesn't need to be async

akirillo (approved):

PR #590, opened 2024-05-15
core: Bring relayer to functional state with new raft

joeykraut:
### Purpose
This PR fixes a host of bugs with the new consensus engine to get it into a working state. 

### Todo
- Gossip-centric setup
- Automatic learner promotion
- Snapshots

### Testing
- Unit tests pass crate wide
- Ran a raft of two nodes, went through the process of setting up two wallets (one on each node) and letting the internal matching engine run on the nodes

### DIFF
--- a/core/src/setup.rs
+++ b/core/src/setup.rs
@@ -41,7 +41,7 @@ pub async fn node_setup(
 
     // TODO: remove this once we have a more fleshed
     // out startup flow
-    tokio::time::sleep(Duration::from_secs(15)).await;
+    tokio::time::sleep(Duration::from_secs(5)).await;
 
     // Setup the contract constants
     fetch_contract_constants(client).await?;
@@ -81,7 +81,10 @@ async fn setup_relayer_wallet(
 
     // Set the node metadata entry for the relayer's wallet
     let wallet_id = derive_wallet_id(key).map_err(err_str!(CoordinatorError::Setup))?;
-    state.set_local_relayer_wallet_id(wallet_id).map_err(err_str!(CoordinatorError::Setup))?;
+    state
+        .set_local_relayer_wallet_id(wallet_id)
+        .await
+        .map_err(err_str!(CoordinatorError::Setup))?;
 
     // Attempt to find the wallet on-chain
     if find_wallet_onchain(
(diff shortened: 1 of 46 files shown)


core/src/setup.rs
```diff
@@ -41,7 +41,7 @@ pub async fn node_setup(
 
     // TODO: remove this once we have a more fleshed
     // out startup flow
-    tokio::time::sleep(Duration::from_secs(15)).await;
+    tokio::time::sleep(Duration::from_secs(5)).await;
```
joeykraut, line 44: going to remove this pretty soon anyways, we don't need to wait for a leader election, this is actually more important for the followers now.

mock-node/src/lib.rs
```diff
@@ -313,29 +305,29 @@ impl MockNodeController {
...
         self
     }
 
     /// Add a gossip server to the mock node
     pub fn with_gossip_server(mut self) -> Self {
         let config = &self.config;
         let state = self.state.clone().expect("State not initialized");
-        let local_peer_id = state.get_peer_id().expect("Failed to get peer id");
+        let rt = TokioRuntime::new().expect("Failed to create tokio runtime");
+        let local_peer_id = rt.block_on(state.get_peer_id()).expect("Failed to get peer id");
```
akirillo, line 330: nit: use `run_fut` here?

state/src/replicationv2/raft.rs
```diff
@@ -106,12 +109,8 @@ impl RaftClient {
 
         // Initialize the raft
         if config.init {
-            let mut initial_nodes = config.initial_nodes.clone();
-            if !initial_nodes.contains(&config.id) {
-                initial_nodes.push(config.id);
-            }
```
akirillo, line 112: why do we no longer do this? is the node expected to add itself to the cluster membership only after a round-trip of being discovered by another node in the gossip layer (triggering a proposal to add the local node as a learner)?

akirillo (approved):

state/src/replicationv2/raft.rs
joeykraut: Yes, it's expected a node will configure this value with its own node info. 

The reason this is messy at this layer is that the `initial_nodes` type is now `BTreeMap<NodeId, Node>` rather than `BTreeSet<NodeId>`. The ladder (and former implementation) fails to store the node info -- including the peer ID. This means that when membership is disseminated to new learners, they lose the information needed to dial the leader.

Instead, we need to construct a `BTreeMap` with the node info. This _could_ happen here, but it's cleaner to do at a higher layer where we have access to the relayer config and thereby the local peer's `PeerId`

PR #593, opened 2024-05-16
task-driver: tasks: node-startup: Define node startup task skeleton

joeykraut:
### Purpose
This PR sets up the skeleton for the node startup task. This task will handle a few responsibilities:
1. Giving the gossip layer time to warmup and discover peers
2. Join an existing cluster raft if one is found, otherwise create one anew
3. Manage the process of catching up to the raft; requesting snapshots and proposing self-promotion when the node is caught up.
4. Fetch the relayer wallet on-chain if it is not available in raft state

I also separated out the task descriptors into their own module with individual files. The single `descriptor.rs` file was becoming unwieldy.  

### Testing
- Unit tests pass


common/src/types/tasks/descriptors/mod.rs
```diff
@@ -0,0 +1,297 @@
...
+
+    use super::{
+        NewWalletTaskDescriptor, QueuedTask, QueuedTaskState, TaskDescriptor, TaskQueueKey,
+    };
+
+    /// Generate the wallet update signature for a new wallet
+    pub fn gen_wallet_update_sig(wallet: &Wallet, key: &SecretSigningKey) -> Vec<u8> {
+        // Serialize and hash the wallet commitment
+        let new_wallet_comm = wallet.get_wallet_share_commitment();
+        let digest = keccak256(new_wallet_comm.to_biguint().to_bytes_be());
```
akirillo, line 203: pls no, we should be importing `contracts_common::custom_serde::BytesSerializable` and doing:
```suggestion
        let digest = keccak256(new_wallet_comm.inner().serialize_to_bytes());
```

workers/task-driver/src/tasks/node_startup.rs
```diff
@@ -0,0 +1,187 @@
...
+/// Defines the state of the node startup task
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub enum NodeStartupTaskState {
+    /// The task is awaiting scheduling
+    Pending,
+    /// The task is waiting for the gossip layer to warm up
+    GossipWarmup,
+    /// The task is completed
+    Completed,
+}
```
akirillo, line 35: understand that this a skeleton, but do we wanna stub in the other steps now? e.g. `RaftCatchup`, `SetupRelayerWallet`?

akirillo (approved):
one important tweak, but doesn't require discussion so approving

workers/task-driver/src/tasks/node_startup.rs
joeykraut: not sure what the state are yet so i'll hold off til implementation

common/src/types/tasks/descriptors/mod.rs
joeykraut: done, pulled in changes from your PR on main

PR #594, opened 2024-05-16
task-driver: node-startup: Create new raft cluster after gossip warmup

joeykraut:
### Purpose
This PR implements the process by which a new raft is initialized if none is detected during gossip warmup. The node startup task waits for a configurable timeout to elapse before checking whether it has been added to an initialized (non-empty voters) raft cluster. If not, it will initialize a new raft cluster with the peer ids it knows to be in its cluster.

### Todo
- When joining an existing cluster, a node must:
    - Request a snapshot and begin catchup
    - Request promotion to a voter when it has replicated all logs

### Testing
- Unit tests pass crate wide
- Tested consistency after wallet updates between raft nodes given the following conditions:
    - Node 1 starts up, completes gossip warmup and initializes a new cluster. Node 2 starts up and learns about this raft after it is intialized.
    - Node 1 starts up; while warming up gossip Node 2 starts up and the two learn of one another. Node 1 finishes gossip warmup first and initializes a new raft with both peers.
    - Node 1 starts up; while warming up gossip, Node 2 starts up and _finishes its warmup before Node 1_. The nodes have learned of one another, so Node 2 initializes a raft with both peers

### DIFF
--- a/common/src/types/tasks/descriptors/node_startup.rs
+++ b/common/src/types/tasks/descriptors/node_startup.rs
@@ -2,7 +2,7 @@
 
 use serde::{Deserialize, Serialize};
 
-use super::TaskIdentifier;
+use super::{TaskDescriptor, TaskIdentifier};
 
 /// The task descriptor for the node startup task
 #[derive(Clone, Debug, Serialize, Deserialize)]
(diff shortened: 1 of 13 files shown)


akirillo (approved):
this lgtm, but iiuc this means we can remove the `assume_leader` config option, right? nodes will warm up into gossip, initialize a raft w/ each other included, and one of them will step up as leader, right?

joeykraut:
> this lgtm, but iiuc this means we can remove the `assume_leader` config option, right? nodes will warm up into gossip, initialize a raft w/ each other included, and one of them will step up as leader, right?
> 
> **EDIT:** nevermind - I see we only add raft learners when adding peers (which should happen in the gossip warmup period) if a raft is already initialized. so we still must have someone `assume_leader`

copying here for posterity: we decided to remove in a follow-up

PR #596, opened 2024-05-17
task-driver: node-startup: Lookup relayer wallet if first node online

joeykraut:
### Purpose
This PR adds protocol fee & relayer wallet lookup to the node startup task. If, after gossip warmup, a node has discovered no existing raft, it will create a new raft and lookup/create a relayer wallet. If instead the node has found an existing raft, it will do no lookup and instead learn about the relayer wallet through the replication subsystem.

### Raft Todo
- Test snapshotting
- Await promotion when joining a new cluster

### Testing
- Unit test pass
- Tested a two node cluster; both cases in which the first node online can find a wallet on-chain and when a new wallet must be created. Verified that both nodes had the wallet after the setup was finished

### DIFF
--- a/workers/task-driver/src/tasks/node_startup.rs
+++ b/workers/task-driver/src/tasks/node_startup.rs
@@ -9,17 +9,34 @@ use std::{error::Error, fmt::Display, time::Duration};
 
 use arbitrum_client::client::ArbitrumClient;
 use async_trait::async_trait;
-use common::types::tasks::NodeStartupTaskDescriptor;
+use common::types::{
+    tasks::{LookupWalletTaskDescriptor, NewWalletTaskDescriptor, NodeStartupTaskDescriptor},
+    wallet::{
+        derivation::{
+            derive_blinder_seed, derive_share_seed, derive_wallet_id, derive_wallet_keychain,
+        },
+        KeyChain, Wallet, WalletIdentifier,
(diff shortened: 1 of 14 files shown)


workers/task-driver/src/tasks/node_startup.rs
```diff
@@ -234,16 +308,100 @@ impl NodeStartupTask {
...
+    async fn find_wallet_onchain(
+        &self,
+        wallet_id: WalletIdentifier,
+        blinder_seed: Scalar,
+        share_seed: Scalar,
+        keychain: KeyChain,
+    ) -> Result<bool, NodeStartupTaskError> {
+        info!("Finding relayer wallet on-chain");
+        let descriptor =
+            LookupWalletTaskDescriptor::new(wallet_id, blinder_seed, share_seed, keychain)
```
akirillo, line 372: do we need this lookup task to bypass the queue as well? probably not, since the raft should already be initialized by this point, right?

akirillo (approved):

workers/task-driver/src/tasks/node_startup.rs
joeykraut: correct, once we're initialized the raft we're good to go with task queue. it's only the original task that sets up the raft which requires we bypass

PR #598, opened 2024-05-17
state: applicator: handle expected application failures

akirillo:
This PR enables the state machine to gracefully handle expected failures, dubbed "rejections", of proposal applications. This is achieved by adding the `ApplicationRejected` variant to the `ApplicatorReturnType`, which is returned by the applicator in a few choice scenarios.

Currently, we only return a rejection in the following cases:
1. Attempting to pop a task when the queue is paused
2. Attempting to transition a task when the queue is paused
3. Attempting to preempt a task queue when some task on it has already committed
4. Attempting to preempt a task queue when it is already paused

I have tested this locally by running a relayer in which the `append_task` state transition returns an `Ok(ApplicatorReturnType::ApplicationRejected(..))`, and one in which it returns an `Err(..)`, and have confirmed that only in the latter case does the error propagate to the Raft core.

Currently deploying this to dev for testing at scale.

### DIFF
--- a/state/src/applicator/return_type.rs
+++ b/state/src/applicator/return_type.rs
@@ -8,14 +8,18 @@
 
 use common::types::mpc_preprocessing::PairwiseOfflineSetup;
 
+use super::error::StateApplicatorError;
+
 /// The return type from the Applicator
-#[derive(Clone, Debug)]
+#[derive(Debug)]
 #[allow(clippy::large_enum_variant)]
 pub enum ApplicatorReturnType {
     /// A set of MPC preprocessing values that are ready to be used
     MpcPrep(PairwiseOfflineSetup),
     /// No return value
     None,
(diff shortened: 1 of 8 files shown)


state/src/applicator/return_type.rs
```diff
@@ -8,14 +8,18 @@
...
-#[derive(Clone, Debug)]
+#[derive(Debug)]
 #[allow(clippy::large_enum_variant)]
 pub enum ApplicatorReturnType {
     /// A set of MPC preprocessing values that are ready to be used
     MpcPrep(PairwiseOfflineSetup),
     /// No return value
     None,
+    /// The application of the proposal failed in an expected manner
+    ApplicationRejected(StateApplicatorError),
```
joeykraut, line 22: nit: maybe just rejected? less verbose

joeykraut (approved):
lgtm

PR #601, opened 2024-05-18
arbitrum-client: Decrease ethers polling interval for quicker tx updates

sehyunc:
This PR reduces the polling interval of the ethers provider used in the `arbitrum-client`. The default value is 7s, probably set for use on Ethereum where block times are ~12s (Arbitrum block time is ~260ms). This value is also used as the initial delay to wait before querying the chain for updates.

Tested in the fronted on local relayer, SubmittingTx now takes ~1s compared to ~4-5s before.

### DIFF
--- a/arbitrum-client/src/client/mod.rs
+++ b/arbitrum-client/src/client/mod.rs
@@ -1,7 +1,7 @@
 //! The definition of the Arbitrum client, which holds the configuration
 //! details, along with a lower-level handle for the darkpool smart contract
 
-use std::{str::FromStr, sync::Arc};
+use std::{str::FromStr, sync::Arc, time::Duration};
 
 use alloy_primitives::ChainId;
 use constants::{DEVNET_DEPLOY_BLOCK, TESTNET_DEPLOY_BLOCK};
@@ -38,6 +38,8 @@ pub struct ArbitrumClientConfig {
     pub rpc_url: String,
     /// The private key of the account to use for signing transactions
     pub arb_priv_key: LocalWallet,
+    /// The interval at which to poll for event filters and pending transactions
+    pub block_polling_interval_ms: u64,
 }
 
 /// A type alias for the RPC client, which is an ethers middleware stack that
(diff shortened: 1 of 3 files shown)


arbitrum-client/src/constants.rs
```diff
@@ -53,6 +53,8 @@ pub const SELECTOR_LEN: usize = 4;
 pub const WALLET_UPDATED_EVENT_NAME: &str = "WalletUpdated";
 /// The abi signature for the `NullifierSpent` event
 pub const NULLIFIER_SPENT_EVENT_NAME: &str = "NullifierSpent";
+/// The interval at which to poll for event filters and pending transactions
+pub const BLOCK_POLLING_INTERVAL: u64 = 0;
```
sehyunc, line 57: Open to pushback on setting this to 0, not sure of the networking drawbacks of querying this quickly.

joeykraut: the chain events worker polls using these same parameters [here](https://github.com/renegade-fi/renegade/blob/main/workers/chain-events/src/listener.rs#L86-L96). I believe the stream it constructs will poll under the hood. This will go on continuously so this is probably too frequent for chain events.

Can we expose this as a config option on the client and build a separate client for `chain-events`?

joeykraut: and for this one can we do like 100ms or something? not a fan of spin loops like this

sehyunc: Sounds good, will add `polling_interval` to the config and instantiate a new client for `chain-events`. 7s is too long for the `chain-events` client as well right?

joeykraut: lets set it to 7s for now, the behavior `chain-events` exposes is only important for multi-cluster so we can tune it then

arbitrum-client/src/client/mod.rs
```diff
@@ -38,6 +38,8 @@ pub struct ArbitrumClientConfig {
     pub rpc_url: String,
     /// The private key of the account to use for signing transactions
     pub arb_priv_key: LocalWallet,
+    /// The interval at which to poll for event filters and pending transactions
+    pub block_polling_interval: u64,
```
joeykraut, line 42: nit: maybe name this `block_polling_interval_ms` to make the unit clear, same with the constants.

joeykraut (approved):
lgtm, thanks for the change, excited to see the speedups

PR #602, opened 2024-05-18
task-driver: node-startup: Await leader election and voter promotion

joeykraut:
### Purpose
This PR adds two wait conditions to each of the raft setup paths:
1. If starting a new raft, the task will wait for leader election before continuing. This is especially important when a raft is initialized with a number of other peers together.
2. If joining an existing raft, the task will await its sync with the leader and subsequent promotion from learner to follower (voter)

### Testing
- Unit tests pass crate-wide
- Tested both cases, joining an existing raft and creating one anew

### DIFF
--- a/state/src/replicationv2/raft.rs
+++ b/state/src/replicationv2/raft.rs
@@ -3,9 +3,10 @@
 use std::{
     collections::{BTreeMap, BTreeSet},
     sync::Arc,
+    time::Duration,
 };
 
-use openraft::{ChangeMembers, Config as RaftConfig, Membership, RaftMetrics};
+use openraft::{ChangeMembers, Config as RaftConfig, Membership, RaftMetrics, ServerState};
 use tracing::info;
 use util::err_str;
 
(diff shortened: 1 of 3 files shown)


state/src/replicationv2/raft.rs
```diff
@@ -178,6 +186,32 @@ impl RaftClient {
...
 
+    // -------------------
+    // | Wait Conditions |
+    // -------------------
+
+    /// Await the promotion of the local peer to a voter
+    pub async fn await_promotion(&self) -> Result<(), ReplicationV2Error> {
+        let timeout = Duration::from_millis(DEFAULT_PROMOTION_TIMEOUT_MS);
+        self.raft
+            .wait(Some(timeout))
```
akirillo, line 197: that is so neat

akirillo (approved):

state/src/replicationv2/raft.rs
joeykraut: yeah definitely a really neat abstraction to learn from

PR #610, opened 2024-05-21
workers: task-driver: driver: Await task queue pause + resume

joeykraut:
### Purpose
This task fixes a bug in which concurrent matches were allowed to process. This ends up panicking the state machine and thereby the raft as we try to pop a preemptive task twice. The root cause is that two match tasks _both_ attempt to pause the task queue. Only one succeeds, but neither waits for the pause state transition to finish before starting the task.

I changed the driver's implementation to await the preemption proposal before continuing.

While fixing this codepath, I also fixed a few other bugs that have not yet come up but will:
- We now pause/resume task queues in an "all-or-none" fashion. This mitigates a bug in which we succeed to pause one task queue and not another. We will fail the driver's loop and never resume the first wallet's task queue.
- To more easily support the above bug fix, I had the state machine return an `Err` on rejection rather than `Ok(ApplicatorReturnType::Rejected)`. This simplifies the implementation and allows us to use try operators within a tx -- so that the tx will abort if one try fails, that is, if one pause/resume is rejected

### Testing
- Unit tests pass, added tests for pausing/resuming multiple queues
- Replicated the `"expected preemptive task"` bug locally, verified that this fix mitigates it -> we fail one of the task driver jobs

### DIFF
--- a/state/src/interface/task_queue.rs
+++ b/state/src/interface/task_queue.rs
@@ -167,15 +167,24 @@ impl State {
     }
 
     /// Pause a task queue placing the given task at the front of the queue
-    #[instrument(name = "propose_preempt_task_queue", skip_all, err, fields(task_id, task = %task.display_description()))]
     pub async fn pause_task_queue(
         &self,
         key: &TaskQueueKey,
         task_id: TaskIdentifier,
         task: TaskDescriptor,
     ) -> Result<ProposalWaiter, StateError> {
-        // Pick a task ID and create a task from the description
-        backfill_trace_field("task_id", task_id.to_string());
+        self.pause_multiple_task_queues(vec![*key], task_id, task).await
+    }
+
+    /// Pause multiple task queues, placing the given task at the front of each
+    /// queue
+    #[instrument(name = "propose_preempt_task_queues", skip_all, err, fields(queue_keys = ?keys, task_id = %task_id, task = %task.display_description()))]
+    pub async fn pause_multiple_task_queues(
+        &self,
+        keys: Vec<TaskQueueKey>,
+        task_id: TaskIdentifier,
+        task: TaskDescriptor,
+    ) -> Result<ProposalWaiter, StateError> {
         let self_id = self.get_peer_id().await?;
         let task = QueuedTask {
             id: task_id,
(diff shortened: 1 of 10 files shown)


state/src/interface/task_queue.rs
```diff
@@ -167,15 +167,26 @@ impl State {
...
+    /// Pause multiple task queues, placing the given task at the front of each
+    /// queue
+    #[instrument(name = "propose_preempt_task_queues", skip_all, err, fields(queue_keys, task_id = %task_id, task = %task.display_description()))]
+    pub async fn pause_multiple_task_queues(
+        &self,
+        keys: Vec<TaskQueueKey>,
+        task_id: TaskIdentifier,
+        task: TaskDescriptor,
+    ) -> Result<ProposalWaiter, StateError> {
+        backfill_trace_field("queue_keys", format!("{keys:?}"));
```
akirillo, line 188: nit: you can set the `queue_keys` trace field in the `#[instrument]` macro like this:
```suggestion
    #[instrument(name = "propose_preempt_task_queues", skip_all, err, fields(queue_keys = ?keys, task_id = %task_id, task = %task.display_description()))]
    pub async fn pause_multiple_task_queues(
        &self,
        keys: Vec<TaskQueueKey>,
        task_id: TaskIdentifier,
        task: TaskDescriptor,
    ) -> Result<ProposalWaiter, StateError> {
```

workers/task-driver/src/driver.rs
```diff
@@ -268,12 +272,12 @@ impl TaskExecutor {
             error!("error running immediate task: {e:?}");
         }
 
-        // Unpause the queues for the affected local wallets
-        for wallet_id in wallet_ids.iter() {
-            self.state().resume_task_queue(wallet_id, res.is_ok()).await?;
+        // Unpause the queues for the affected local wallets and remove
+        // from the preemptive tasks list
+        if wallet_ids.is_empty() {
```
akirillo, line 277: should this be:
```suggestion
        if !wallet_ids.is_empty() {
```

akirillo (approved):

workers/task-driver/src/driver.rs
joeykraut: good catch

PR #612, opened 2024-05-22
state: applicator: task_queue: record task queue lengths

akirillo:
This PR adds metrics gauging the lengths of tasks queues in the relayer. We do this in the applicator, despite the fact that metrics will be emitted from all nodes in the cluster - this can be easily deduplicated in DataDog. We don't record these metrics in the interface layer (i.e. when proposing state transitions) because we would be unable to accurately capture queue lengths when preemptive tasks are added / queues are cleared due to task failure.

### DIFF
--- a/state/Cargo.toml
+++ b/state/Cargo.toml
@@ -53,6 +53,7 @@ tracing-slog = "0.2"
 tui = "0.19"
 tui-logger = "0.8"
 uuid = "1.1.2"
+metrics = { workspace = true }
 
 [dev-dependencies]
 multiaddr = "0.17"
--- a/state/Cargo.toml
+++ b/state/Cargo.toml
@@ -7,6 +7,7 @@ edition = "2021"
 # Used to enable mocks from other crates in tests
 all-tests = ["common/mocks"]
 mocks = ["dep:tempfile", "dep:test-helpers"]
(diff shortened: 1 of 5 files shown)


state/Cargo.toml
```diff
@@ -53,6 +54,7 @@ tracing-slog = "0.2"
 tui = "0.19"
 tui-logger = "0.8"
 uuid = "1.1.2"
+metrics = { workspace = true }
```
joeykraut, line 57: nit: make this conditional on the `task-queue-len` feature

state/src/applicator/task_queue.rs
```diff
@@ -50,6 +54,23 @@ fn new_running_state() -> QueuedTaskState {
     QueuedTaskState::Running { state: PENDING_STATE.to_string(), committed: false }
 }
 
+/// Record the length of a task queue
+fn record_task_queue_length<T: TransactionKind>(key: &TaskQueueKey, tx: &StateTxn<'_, T>) {
```
joeykraut, line 58: compiler is _probably_ smart enough to do this, but i'd `#[inline]` this to avoid a function call when the feature is disabled

joeykraut (approved):

PR #614, opened 2024-05-23
arbitrum-client: use multiple arbitrum keypairs

akirillo:
This PR tweaks the `arbitrum_client` to be configured with a list of Arbitrum keypairs, from which it will randomly select one every time a darkpool contract client is needed. This required an extra compiler flag to gate usage of the `rand` crate so that the contracts repo, which import that `arbitrum_client` crate just to make use of their conversion utilities, can still be built to WASM.
The relayer wallet is derived from the first supplied keypair.

I tested this by running a local relayer w/ 2 keys, executing a few `new-wallet` tasks, and asserting that there were transactions submitted from both associated addresses.

### DIFF
--- a/arbitrum-client/src/client/mod.rs
+++ b/arbitrum-client/src/client/mod.rs
@@ -36,8 +36,10 @@ pub struct ArbitrumClientConfig {
     pub chain: Chain,
     /// HTTP-addressable RPC endpoint for the client to connect to
     pub rpc_url: String,
-    /// The private key of the account to use for signing transactions
-    pub arb_priv_key: LocalWallet,
+    /// The private keys of the accounts to use for signing transactions.
+    /// Multiple keys can be provided to mitigate nonce contention across a node
+    /// / cluster.
+    pub arb_priv_keys: Vec<LocalWallet>,
     /// The interval at which to poll for event filters and pending transactions
     pub block_polling_interval_ms: u64,
 }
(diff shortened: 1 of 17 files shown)


arbitrum-client/src/client/mod.rs
```diff
@@ -90,42 +94,60 @@ impl ArbitrumClientConfig {
...
+            .collect();
         let deploy_block = config.get_deploy_block();
 
-        Ok(Self { darkpool_contract, deploy_block })
+        Ok(Self { darkpool_clients, deploy_block })
     }
 
-    /// Get a reference to the underlying RPC client
+    /// Get a darkpool contract client
+    pub fn darkpool(&self) -> DarkpoolContract<MiddlewareStack> {
```
joeykraut, line 126: nit: `get_client`, `get_signer`, or similar would be much clearer. it is unclear in the callsites what it means to call `darkpool` on the `ArbitrumClient`

workers/task-driver/integration/helpers.rs
```diff
@@ -188,6 +187,12 @@ pub(crate) async fn setup_relayer_wallet(test_args: &IntegrationTestArgs) -> Res
     Ok(())
 }
 
+/// Get the address of the (assumed to be only) Arbitrum account
+/// with which the relayer is configured
+pub fn wallet_address(test_args: &IntegrationTestArgs) -> Address {
```
joeykraut, line 192: nit: maybe just implement this on the test args?

joeykraut (approved):

PR #616, opened 2024-05-23
workers: gossip-server: Notify cluster peers before expiring

joeykraut:
### Purpose
This PR is the first of two that improves the way in which we expire peers. Specifically, we implement a [SWIM](https://www.cs.cornell.edu/projects/Quicksilver/public_pdfs/SWIM.pdf) inspired propose/reject protocol wherein:
1. A node believes another to have failed, by timing out on heartbeats. This node sends a pubsub message to the cluster proposing that the cluster expire the candidate peer.
2. Each recipient of the pubsub message checks their own last known heartbeat, if it is within some tolerance they reject the proposal
3. If no rejections are sent within some time window, the cluster expires the peer

This PR implements the messaging half of that, but not the timeout, rejection, expiry logic.

### Testing
- Ran a cluster and killed a node, verified that the messages were sent correctly

### DIFF
--- a/workers/gossip-server/src/peer_discovery/heartbeat.rs
+++ b/workers/gossip-server/src/peer_discovery/heartbeat.rs
@@ -1,15 +1,21 @@
 //! Groups gossip server logic for the heartbeat protocol
 
 use common::types::gossip::WrappedPeerId;
-use gossip_api::request_response::{
-    heartbeat::{HeartbeatMessage, PeerInfoRequest},
-    orderbook::OrderInfoRequest,
-    GossipRequest,
+use gossip_api::{
+    pubsub::{
+        cluster::{ClusterManagementMessage, ClusterManagementMessageType},
+        PubsubMessage,
+    },
+    request_response::{
+        heartbeat::{HeartbeatMessage, PeerInfoRequest},
+        orderbook::OrderInfoRequest,
+        GossipRequest,
+    },
 };
 use job_types::network_manager::{NetworkManagerControlSignal, NetworkManagerJob};
 use renegade_metrics::helpers::record_num_peers_metrics;
 use tracing::info;
-use util::{err_str, get_current_time_seconds};
+use util::{err_str, get_current_time_millis};
 
 use crate::{errors::GossipError, server::GossipProtocolExecutor};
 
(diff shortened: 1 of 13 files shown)


workers/gossip-server/src/peer_discovery/peers.rs
```diff
@@ -51,6 +52,50 @@ impl GossipProtocolExecutor {
...
+        let info = match peer_info {
+            Some(info) => info,
+            None => return Ok(()),
+        };
+
+        // If the local peer has received a recent heartbeat from the candidate, notify
+        // the sender that the expiry should not proceed
+        let now = get_current_time_millis();
+        let time_since_last_heartbeat = now - info.last_heartbeat;
+        if time_since_last_heartbeat < HEARTBEAT_FAILURE_MS {
```
akirillo, line 74: not sure if we're planning on adjusting these windows later, but for now should this be check be using `CLUSTER_HEARTBEAT_FAILURE_MS` depending on if the peer is a cluster peer?

workers/gossip-server/src/peer_discovery/peers.rs
```diff
@@ -51,6 +52,50 @@ impl GossipProtocolExecutor {
...
+
+            // The peer should not expire yet, send a rejection
+            let msg = GossipRequest::RejectExpiry(RejectExpiryRequest {
+                peer_id,
+                last_heartbeat: info.last_heartbeat,
+            });
+
+            let job = NetworkManagerJob::request(sender, msg);
+            self.network_channel.send(job).map_err(err_str!(GossipError::SendMessage))?;
+        }
```
akirillo, line 85: If the peer does _not_ reject here, should they also follow through with expiry logic locally? Otherwise they have to wait for their own heartbeats with the expiry candidate to time out, though it seems semantically the decision to expire is made here

workers/gossip-server/src/peer_discovery/heartbeat.rs
```diff
@@ -123,18 +129,30 @@ impl GossipProtocolExecutor {
...
+        if same_cluster && last_heartbeat < CLUSTER_HEARTBEAT_FAILURE_MS {
             return Ok(());
-        } else if !same_cluster && last_heartbeat < HEARTBEAT_FAILURE_MS / 1000 {
+        } else if !same_cluster && last_heartbeat < HEARTBEAT_FAILURE_MS {
             return Ok(());
         }
 
+        // TODO: Give the cluster time to reject peer expiry
+
+        // Send a peer expiry proposal to the cluster
```
akirillo, line 144: per our convo internally, this pubsub message should only be sent if `same_cluster == true`, right?

akirillo (approved):
Have some qs but defer to ur understanding, approved

workers/gossip-server/src/peer_discovery/peers.rs
joeykraut: yeah i just filled this in to test, the actual logic is more complex than either

PR #617, opened 2024-05-23
workers: gossip-server: Track expiry candidates and invisibility window

joeykraut:
### Purpose
This PR adds a `PeerExpiryWindows` struct that tracks both the expiry candidacy and invisibility windows. These windows are updated through the expiry gossip protocol.

### Todo
- Move expiry rejection to pubsub and transition peers to expiry candidates locally when not rejecting an expiry proposal

### Testing
- Unit tests pass
- Tested expiry of a node; verified that it went into the candidate state and waited the appropriate time before expiry.

### DIFF
--- a/workers/gossip-server/src/peer_discovery/expiry_window.rs
+++ b/workers/gossip-server/src/peer_discovery/expiry_window.rs
@@ -44,12 +44,6 @@ impl<T: Hash + Eq + PartialEq> TimeWindowBuffer<T> {
         self.windows.write().await
     }
 
-    /// Cleanup all expired windows
-    async fn cleanup(&self) {
-        let mut this = self.write_windows().await;
-        this.retain(|_, expiry_time| *expiry_time > Instant::now());
-    }
-
     /// Whether or not the given key is present in the buffer
     pub async fn contains(&self, key: &T) -> bool {
         let this = self.read_windows().await;
@@ -59,29 +53,22 @@ impl<T: Hash + Eq + PartialEq> TimeWindowBuffer<T> {
     /// Check whether the window for a key has expired
     pub async fn is_expired(&self, key: &T) -> bool {
         let this = self.read_windows().await;
-        match this.get(key) {
-            Some(expiry_time) => *expiry_time < Instant::now(),
-            None => true,
-        }
+        this.get(key).map_or(false, |expiry_time| *expiry_time < Instant::now())
     }
 
     /// Adds a key to the buffer
+    ///
+    /// Does not replace existing windows
     pub async fn add(&self, key: T, dur: Duration) {
         let expiry_time = Instant::now() + dur;
         let mut this = self.write_windows().await;
         this.entry(key).or_insert(expiry_time);
-        drop(this); // unlock
-
-        self.cleanup().await;
     }
 
     /// Remove a key from the buffer
     pub async fn remove(&self, key: T) {
         let mut this = self.write_windows().await;
         this.remove(&key);
-        drop(this); // unlock
-
-        self.cleanup().await;
     }
 }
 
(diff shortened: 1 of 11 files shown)


workers/gossip-server/src/peer_discovery/expiry_window.rs
```diff
@@ -0,0 +1,169 @@
...
+        let this = self.read_windows().await;
+        this.contains_key(key)
+    }
+
+    /// Check whether the window for a key has expired
+    pub async fn is_expired(&self, key: &T) -> bool {
+        let this = self.read_windows().await;
+        match this.get(key) {
+            Some(expiry_time) => *expiry_time < Instant::now(),
+            None => true,
```
akirillo, line 64: thoughts on having this method return an `Option<bool>` and letting the callsite decide if a `None` implies the window has expired / was never added?

workers/gossip-server/src/peer_discovery/expiry_window.rs
```diff
@@ -0,0 +1,169 @@
...
+        }
+    }
+
+    /// Adds a key to the buffer
+    pub async fn add(&self, key: T, dur: Duration) {
+        let expiry_time = Instant::now() + dur;
+        let mut this = self.write_windows().await;
+        this.entry(key).or_insert(expiry_time);
```
akirillo, line 72: can we add to the doc comment that this will _not_ overwrite the window's duration if it already exists?

workers/gossip-server/src/peer_discovery/expiry_window.rs
akirillo: For example if we call `PeerExpiryWindows::should_expire` below with a `peer_id` that was never a candidate, we will expire it

workers/gossip-server/src/peer_discovery/expiry_window.rs
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    /// Remove a candidate from the expiry candidate list
+    pub async fn remove_expiry_candidate(&self, peer_id: WrappedPeerId) {
+        self.candidates.remove(peer_id).await;
+    }
+
+    /// Mark a peer as expired, adding it to the invisibility window
+    pub async fn mark_expired(&self, peer_id: WrappedPeerId) {
+        self.mark_invisible(peer_id).await;
```
akirillo, line 167: should this method also remove the peer from the candidate list?

workers/gossip-server/src/peer_discovery/heartbeat.rs
```diff
@@ -115,45 +110,106 @@ impl GossipProtocolExecutor {
...
+
+    // --- Peer Expiry --- //
+
+    /// Update the expiry status of a peer
+    async fn update_expiry_status(&self, peer_id: WrappedPeerId) -> Result<(), GossipError> {
+        if self.expiry_buffer.is_expiry_candidate(&peer_id).await {
+            self.maybe_expire_candidate(peer_id).await
+        } else {
+            self.maybe_expire_peer(peer_id).await
+        }
```
akirillo, line 136: IIUC, it looks like the expiration of a cluster peer now goes through two stages via `update_expiry_status` - if the peer hasn't heartbeated sufficiently recently, it gets marked as an expiry candidate, then if it is a candidate it will be expired on the next call to `update_expiry_status`. `update_expiry_status` is only called in `send_heartbeat`, so we basically may wait an extra heartbeat round, leaving a peer in a candidate state, before we actually expire it. This seems fine to me, just wanted to be sure I'm understanding correctly

akirillo: I suppose this is the reason why we add peers to the invisibility window when we mark them as candidates - to allow for the extra heartbeat round?

akirillo (approved):
mostly clarifying questions, lgtm

PR #620, opened 2024-05-24
arbitrum-client, workers: task-driver, proof-manager: add concurrency metrics

akirillo:
This PR adds metrics which will be used to measure the amounts of concurrency in the Arbitrum client, task driver, and proof manager across a node/cluster. Concretely these are pairs of counter metrics which are incremented when a {task, proof job, transaction submission} are started / stopped. The difference between the cumulative sums of these metrics gives us the number of concurrently running actions at a point in time. This approach is taken because StatsD does not support incrementing / decrementing a gauge metric, and is simpler than maintaining some global state for a gauge.

Additionally, in the case of the task driver, we have two separate metrics for tasks being stopped, and tasks completing. This is because tasks can be stopped when they are preempted. The start/stop metrics are used to track concurrency, and the completed metrics are used to track throughput.

These metrics are guarded behind feature flags, which are grouped together in the `core` crate under the `dev-metrics` feature flag.


joeykraut (approved):
worth acknowledging that timing in the metrics collector may slightly skew these, but with steady load it's probably a decent approximation

PR #623, opened 2024-05-25
renegade-metrics: gauge: add atomic gauge metric primitive

akirillo:
This PR introduces the `Gauge` struct, which manages an atomic, incrementable/decrementable gauge value from which metrics are reported. This is necessary as StatsD does not recognize the `gauge.increment()` / `gauge.decrement()` events, only `gauge.set()` - so we expose them through this wrapper.

The metrics from which in-flight tasks, proofs, and transactions were calculated were replaced with static gauges of this type.

### DIFF
--- a/workers/task-driver/src/driver.rs
+++ b/workers/task-driver/src/driver.rs
@@ -17,7 +17,6 @@ use common::{
     Shared,
 };
 use job_types::task_driver::{TaskDriverJob, TaskDriverReceiver, TaskNotificationSender};
-use renegade_metrics::helpers::{decr_inflight_tasks, incr_completed_tasks, incr_inflight_tasks};
 use serde::Serialize;
 use state::State;
 use tokio::runtime::Builder as TokioRuntimeBuilder;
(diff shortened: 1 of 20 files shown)


workers/task-driver/src/driver.rs
```diff
@@ -158,7 +160,7 @@ impl TaskExecutor {
         // Build a runtime
         let runtime = TokioRuntimeBuilder::new_multi_thread()
             .enable_all()
-            .worker_threads(TASK_DRIVER_N_THREADS)
+            // .worker_threads(TASK_DRIVER_N_THREADS)
```
joeykraut, line 163: dead code

joeykraut (approved):

workers/task-driver/src/driver.rs
akirillo: ah yeah, that's from my testing - but fwiw i think we should remove this and let tokio select the number of threads based on underlying available parallelism. thoughts?

PR #624, opened 2024-05-25
snapshot-sidecar: Read snapshot dir and cluster id from relayer config

joeykraut:
### Purpose
This PR changes the `snapshot-sidecar` process to read in the snapshot path and cluster id from the relayer config. The sidecar now also prepends the cluster id to the snapshot.

### Testing
- Unit tests pass
- Ran the sidecar alongside a relayer, had the relayer snapshot. I then downloaded the snapshot from s3 and recovered a relayer's state from it


config/src/cli.rs
```diff
@@ -0,0 +1,386 @@
+//! The relayer CLI and config definitions
```
joeykraut, line 1: essentially all of the config methods are the same, just broken out into files now

akirillo (approved):

PR #627, opened 2024-05-27
state: replication: state-machine: Recover from snapshot at startup

joeykraut:
### Purpose
This PR changes the state machine to check for a snapshot at startup and recover from the snapshot if it is present. This also involves cleanup logic -- which for now is just clearing all task queues so that wallets are not stuck. 

### Todo
- In the node startup task, check all wallets on-chain for updates since the snapshot was taken

### Testing
- Unit tests pass
- Tested recovering a node from a snapshot with a wallet in the state, verified that the state held the wallet without any lookup task

### DIFF
--- a/state/src/replication/state_machine/recovery.rs
+++ b/state/src/replication/state_machine/recovery.rs
@@ -1,0 +1,47 @@
+//! Recover the state machine from a snapshot
+
+use common::types::tasks::{QueuedTask, TaskQueueKey};
+use openraft::SnapshotMeta;
+use tracing::info;
+use uuid::Uuid;
+
+use crate::{replication::error::ReplicationV2Error, TASK_QUEUE_TABLE};
+
+use super::StateMachine;
+
+impl StateMachine {
+    /// Check for a snapshot to recover from
+    pub(crate) async fn maybe_recover_snapshot(&mut self) -> Result<(), ReplicationV2Error> {
+        if !self.snapshot_archive_path().exists() {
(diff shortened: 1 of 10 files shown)


state/src/replication/state_machine/recovery.rs
```diff
@@ -0,0 +1,47 @@
...
+
+        self.update_from_snapshot(&dummy_meta, snap_db).await?;
+        self.clear_task_queues()
+    }
+
+    /// Clear all task queues
+    ///
+    /// We do this when recovering to prevent wallets from being blocked on a
+    /// task queue that failed
+    fn clear_task_queues(&self) -> Result<(), ReplicationV2Error> {
```
akirillo, line 36: why is this safe to do? we might recover the state machine into an inconsistent state from the snapshot. will subsequent logs apply on top of it correctly?

akirillo (approved):
confused on why we need to clear task queues and if it's safe to do so

state/src/replication/state_machine/recovery.rs
joeykraut: i'd not say it's safe, but it's a better outcome than blocking a wallet. for one, there's no other action that is easy to do, i.e. continuing the task would be really difficult. and if we leave it blocked the wallet can't make progress.

it's not an assumption i'd like to rely on, but the startup task will look for wallet updates on-chain too.

essentially, i see this as the best of not great options in the case of a crash

PR #632, opened 2024-05-28
node-support: bootloader: Move bootloader to rust binary

joeykraut:
### Purpose
This PR moves the bootloader script from shell to rust. This will allow us to do more complex bootloader operations like fetching the most recent snapshot from s3.

### Testing
- Built a release image and ran it locally, verified that the relayer booted normally


node-support/bootloader/src/main.rs
```diff
@@ -0,0 +1,171 @@
...
+    // Write the modified config back to the original file
+    let new_config_content =
+        toml::to_string(&config).map_err(raw_err_str!("Failed to serialize config: {}"))?;
+    fs::write(CONFIG_PATH, new_config_content)
+        .await
+        .map_err(raw_err_str!("Failed to write config file: {}"))
+}
+
+/// Download the most recent snapshot
+async fn download_snapshot() -> Result<(), String> {
```
joeykraut, line 144: left for a follow up

akirillo (approved):

PR #636, opened 2024-05-30
workers: handshake-manager: Do not use blocking pool for internal engine

joeykraut:
### Purpose
This PR changes the implementation of the internal matching engine to run on the non-blocking Tokio pool. We observed a deadlock in the matching engine when running a relayer under intense load. This deadlock occurs on the blocking Tokio pool when spawning too many internal engine tasks at one; which themselves spawn blocking tasks for DB access. Eventually we exhaust the blocking pool's ability to spawn new tasks or await blocking tasks as these two granularities of blocking tasks get interleaved and exhaust the pool, preventing one another from making progress.

The internal engine is now implemented `async` anyways with the blocking components themselves spawned onto the blocking pool, so it is fine to run the engine `async`. 

I additionally added early exit conditions to the matching engine that we see under heavy load; i.e. a match from elsewhere in the relayer pauses a queue, or an update is interleaved making the wallet invalid. This check catches both cases and ensures we do not continue attempting to match on a stale wallet.

### Todo
- Improve task notifications to be more direct.

### Testing
- Reproduced some of the pool deadlock behavior locally using the DB of the load tested node, verified that it made progress after this fix was implemented.

### DIFF
--- a/workers/handshake-manager/src/manager.rs
+++ b/workers/handshake-manager/src/manager.rs
@@ -122,7 +122,7 @@ pub struct HandshakeExecutor {
     /// The pricer reporter's work queue, used for fetching price reports
     pub(crate) price_reporter_job_queue: PriceReporterQueue,
     /// The global relayer state
-    pub(crate) global_state: State,
+    pub(crate) state: State,
     /// The queue used to send tasks to the task driver
     pub(crate) task_queue: TaskDriverQueue,
     /// The system bus used to publish internal broadcast messages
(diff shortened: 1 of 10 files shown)


workers/handshake-manager/src/manager.rs
```diff
@@ -207,12 +207,7 @@ impl HandshakeExecutor {
             // new order to check for matches
             HandshakeExecutionJob::InternalMatchingEngine { order } => {
                 // Spawn a blocking thread to handle matches
```
akirillo, line 209: nit: change this comment lol, no longer a blocking thread

workers/handshake-manager/src/manager/internal_engine.rs
```diff
@@ -250,4 +256,33 @@ impl HandshakeExecutor {
...
+
+        // Check the queue
+        let queue_len = self.state.get_task_queue_len(&wallet.wallet_id).await?;
+        if queue_len > 0 {
+            return Ok(false);
+        }
+
+        if !self.state.is_queue_paused(&wallet.wallet_id).await? {
+            return Ok(false);
+        }
```
akirillo, line 284: should this be:
```suggestion
        if self.state.is_queue_paused(&wallet.wallet_id).await? {
            return Ok(false);
        }
```

i.e. if the wallet's queue is now paused, it is no longer valid for matching

akirillo (approved):
one smol tweak but lgtm

PR #637, opened 2024-05-30
task-driver: driver: Notify preemptive task senders directly

joeykraut:
### Purpose
This PR changes the way in which we notify waiters about the result of a preemptive task. Instead of going through the open notifications set, we directly send a response channel alongside the `RunImmediate` job. The driver responds directly on this channel rather than looking up channels in the open notifications map.

This fixes an issue we have seen wherein the handshake manager registers a notification with the driver for an immediate task, but beats the task's registration, causing an erroneous `"task not found"` error. The handshake manager will then retry matches unnecessarily and possibly create a duplicate match which will need to be cancelled. The net effect of this is wasted work.

### Testing
- Tested a case in which two matches occur at the same time on the same wallet. Verified that it was handled cleanly

### DIFF
--- a/core/src/setup.rs
+++ b/core/src/setup.rs
@@ -3,11 +3,9 @@
 //! TODO(@joey): This module will eventually become a dedicated task, for now
 //! this is sufficient
 
-use std::time::Duration;
-
-use common::types::tasks::NodeStartupTaskDescriptor;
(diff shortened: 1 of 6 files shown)


akirillo (approved):
nice

PR #638, opened 2024-05-30
replication: snapshot: Use lock file to signal snapshot completion

joeykraut:
### Purpose
This PR changes the snapshot implementation to create and remove a `.lock` file in between snapshot start and end. This allows the `snapshot-sidecar` to watch for a file deletion event. These events are much more reliable in that they only occur once. Write events occur many times when writing a large snapshot, causing the sidecar to save a snapshot too early, when it is incomplete.

### Testing
- Tested the sidecar with the lock file, verified that I could recover a relayer from the snapshot it produced in s3

### DIFF
--- a/state/src/replication/state_machine/snapshot.rs
+++ b/state/src/replication/state_machine/snapshot.rs
@@ -75,13 +75,17 @@ impl SnapshotInfo {
 
 impl RaftSnapshotBuilder<TypeConfig> for StateMachine {
     async fn build_snapshot(&mut self) -> Result<Snapshot<TypeConfig>, RaftStorageError<NodeId>> {
+        // Create the lock file
+        self.create_snapshot_lock().await.map_err(new_snapshot_error)?;
+
         let info = self.take_db_snapshot().await.map_err(new_snapshot_error)?;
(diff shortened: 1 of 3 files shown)


akirillo (approved):

state/src/replication/state_machine/snapshot.rs
```diff
@@ -75,13 +75,17 @@ impl SnapshotInfo {
 
 impl RaftSnapshotBuilder<TypeConfig> for StateMachine {
     async fn build_snapshot(&mut self) -> Result<Snapshot<TypeConfig>, RaftStorageError<NodeId>> {
+        // Create the lock file
+        self.create_snapshot_lock().await.map_err(new_snapshot_error)?;
```
akirillo, line 79: if the snapshot directory doesn't exist, this method will fail - we should have it create the snapshot directory if it doesn't exist like `take_db_snapshot`

akirillo (approved):

state/src/replication/state_machine/snapshot.rs
joeykraut: done, added it to `create_snapshot_lock`

PR #639, opened 2024-06-01
state: replication: misc tweaks

akirillo:
This PR makes some miscellaneous tweaks to the Raft module to aid in stability, particularly around downloading snapshots when joining a cluster. Namely:
1. We increase the snapshot installation RPC timeout to 1 minute, as this must include the time it takes for a node to install/apply a fully downloaded snapshot as its storage layer
2. We increase the Raft heartbeat interval to 3s, as this is also the timeout for any `AppendEntries` RPCs, which can include thousands of entries for a follower to apply
3. We increase the inter-cluster heartbeat timeout to 30s, as empirically we oftentimes see expiries between cluster peers and the bootstrap nodes
4. We include learners when checking if we've already expired a node. Previously, we would always assume we've already expired a learner, preventing us from appropriately expiring them.
5. We apply a `cargo update`, primarily to update from `openraft@0.9.5` to `openraft@0.9.11`, which includes some minor stability improvements.

This is currently being tested in dev.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14,9 +14,9 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
-version = "0.21.0"
+version = "0.22.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


joeykraut (approved):
nice

PR #640, opened 2024-06-03
gossip-api: trace gossip requests across network

joeykraut:
### Purpose
This PR adds basic tracing to the network manager and propagates the traces across nodes sending/receiving a message. I added a layer of encapsulation to the `GossipRequest` and `GossipResponse` to permit adding tracing headers to these types. 

### Testing
- Deployed to dev, verified that I was able to see request/response handlers together in a trace


akirillo (approved):
this is neat, nice

PR #641, opened 2024-06-03
network-manager: Spawn request authentication on blocking thread

joeykraut:
### Purpose
This PR changes the `network-manager` to spawn all authentication tasks on the blocking pool rather than directly calling them. This prevents the async pool from being drained under heavy load with large payloads. 

A follow-up will change the cluster auth mechanism to use a symmetric scheme to further speed up these blocking operations.

### Testing
- Reproduced a bug wherein `AppendEntries` messages continuously timeout and nodes begin expiring one another. Verified that the node was much more stable after this change. Symmetric auth will further reduce this issue.

### DIFF
--- a/workers/network-manager/src/executor/pubsub.rs
+++ b/workers/network-manager/src/executor/pubsub.rs
@@ -43,14 +43,20 @@ impl NetworkManagerExecutor {
     }
 
     /// Handle an incoming network request for a pubsub message
-    pub(super) fn handle_inbound_pubsub_message(
+    pub(super) async fn handle_inbound_pubsub_message(
         &self,
         message: GossipsubMessage,
     ) -> Result<(), NetworkManagerError> {
         // Deserialize into API types and verify auth
+        let pkey = self.cluster_key.public;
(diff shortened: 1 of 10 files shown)


workers/network-manager/src/executor/pubsub.rs
```diff
@@ -43,14 +43,18 @@ impl NetworkManagerExecutor {
...
+    pub(super) async fn handle_inbound_pubsub_message(
         &self,
         message: GossipsubMessage,
     ) -> Result<(), NetworkManagerError> {
         // Deserialize into API types and verify auth
+        let pkey = self.cluster_key.public;
         let event: AuthenticatedPubsubMessage =
             message.data.try_into().map_err(NetworkManagerError::Serialization)?;
-        event.verify_cluster_auth(&self.cluster_key.public)?;
+        let event =
```
akirillo, line 54: tiny nit but can we drop a small comment here regarding _why_ we spawn on the blocking pool? is just helpful for an uncontextualized reader

akirillo (approved):

PR #642, opened 2024-06-03
gossip-api: Use HMAC to authenticate intra-cluster messages

joeykraut:
### Purpose
This PR switches our intra-cluster auth scheme to use a separate, symmetric key for all messages. This has better performance than the asymmetric curve25519 ECDSA scheme used before.

### Testing
- Unit tests pass
- Ran two nodes in the same cluster, they made progress => they authenticated messages correctly


gossip-api/src/pubsub/mod.rs
```diff
@@ -26,27 +26,21 @@ pub struct AuthenticatedPubsubMessage {
...
+        let sig =
+            if body.requires_cluster_auth() { create_hmac(&body, cluster_key) } else { Vec::new() };
 
-        Ok(Self { sig, body })
+        Self { sig, body }
     }
 
     /// Verify the signature on an authenticated request
-    pub fn verify_cluster_auth(&self, cluster_pubkey: &PublicKey) -> Result<(), SignatureError> {
+    pub fn verify_cluster_auth(&self, cluster_pubkey: &ClusterSymmetricKey) -> bool {
```
akirillo, line 38: tiny nit: can we call this argument `cluster_key`, or something other than `cluster_pubkey` which still implies asymmetric crypto is being used

akirillo (approved):

gossip-api/src/pubsub/mod.rs
joeykraut: yes, good catch

PR #644, opened 2024-06-03
bin: Add scripts for manual relayer deployments

joeykraut:
### Purpose
This PR adds two scripts: `build_and_push.sh` and `deploy.sh` to build and push the relayer image and deploy the relayer ECS service on the latest image respectively. These allow us to deploy without going through CI, which can be costly.

### Testing
- Tested both scripts in `dev`

### DIFF
--- a/bin/build_and_push.sh
+++ b/bin/build_and_push.sh
@@ -1,0 +1,13 @@
+#!/bin/sh
+REGION=us-east-2
+ENVIRONMENT=${1:-dev}
+ECR_URL=377928551571.dkr.ecr.us-east-2.amazonaws.com/relayer-$ENVIRONMENT
+
+GIT_HASH=$(git rev-parse HEAD)
+docker build -t relayer:latest  -f ./docker/release/Dockerfile .
+aws ecr get-login-password --region $REGION | docker login --username AWS --password-stdin $ECR_URL
+
+docker tag relayer:latest $ECR_URL:$GIT_HASH
+docker tag relayer:latest $ECR_URL:latest
+docker push $ECR_URL:$GIT_HASH
+docker push $ECR_URL:latest
--- a/bin/deploy.sh
+++ b/bin/deploy.sh
@@ -1,0 +1,35 @@
+#!/bin/sh
+REGION=us-east-2
+ENVIRONMENT=${1:-dev}
+CLUSTER_IDX=${2:-0}
+CLUSTER_NAME=$ENVIRONMENT-cluster$CLUSTER_IDX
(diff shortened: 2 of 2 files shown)


bin/build_and_push.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+REGION=us-east-2
+ENVIRONMENT=${1:-dev}
+ECR_URL=377928551571.dkr.ecr.us-east-2.amazonaws.com/relayer-$ENVIRONMENT
+
+GIT_HASH=$(git rev-parse HEAD)
+docker build -t relayer:latest  -f ./docker/release/Dockerfile .
```
akirillo, line 7: should we have the build target `--platform linux/arm64`? maybe that's why the local build is noticeably faster over the BuildJet runners, which do this?

bin/deploy.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/sh
+REGION=us-east-2
+ENVIRONMENT=${1:-dev}
+CLUSTER_NAME=$ENVIRONMENT-cluster0
```
akirillo, line 4: nit: currently, we only ever use cluster0, but maybe we could have the cluster index be a script arg w/ a default in the same way `ENVIRONMENT` is?

bin/deploy.sh
```diff
@@ -0,0 +1,34 @@
...
+SERVICE_NAME=$ENVIRONMENT-cluster0-service
+TASK_FAMILY=$ENVIRONMENT-cluster0-task-def
+ECR_URL=377928551571.dkr.ecr.us-east-2.amazonaws.com/relayer-$ENVIRONMENT
+
+# Fetch the latest image URI from ECR
+IMAGE_URI=$(aws ecr describe-images --repository-name relayer-dev --region us-east-2 --image-ids imageTag=latest | jq -r '.imageDetails[0].imageTags[0]')
```
akirillo, line 10: when I ran this locally, the value returned was just the `latest` tag. it's quite nice to see the github commit hash in the task definition, but i'm not sure if we can rely on its ordering being before/after `latest`... maybe we could just use the `imageDigest` field then for the full URI? we can still match the image digest to the github hash tag in ECR when debugging

akirillo (approved):
couple of remarks but given this is a util script they're not blocking

bin/build_and_push.sh
joeykraut: dockerfile already defaults to this

bin/deploy.sh
joeykraut: good suggestion, done

PR #646, opened 2024-06-04
state: applicator, storage: Track task executor via `task-assignments` table

joeykraut:
### Purpose
This PR removes the `executor` field on each task in favor of tracking this information in a `task-assignments` table. Doing so will allow us to re-assign tasks when a node goes down in a simpler manner. I.e. by updating this index.

### Testing
- Unit tests pass; more will come in reassignment PR
- Tested the basic task flows

### DIFF
--- a/state/src/applicator/task_queue.rs
+++ b/state/src/applicator/task_queue.rs
@@ -1,6 +1,7 @@
 //! Task queue state transition applicator methods
 
 use common::types::{
+    gossip::WrappedPeerId,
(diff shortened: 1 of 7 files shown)


state/src/applicator/task_queue.rs
```diff
@@ -376,6 +412,35 @@ impl StateApplicator {
         Ok(())
     }
 
+    /// Pop the top task on the queue and add it to the historical state    
+    ///
+    /// Returns the task and the peer that executed it
```
akirillo, line 417: nit:
```suggestion
    /// Returns the task
```

akirillo (approved):

PR #655, opened 2024-06-06
state: replication: raft: Periodically sync gossip and raft membership

joeykraut:
### Purpose
This PR groups the learner promotion, missed expiry, and a new missed learner timer loop into one `raft-membership-sync-loop`. This loop checks for new learners, checks for potential learner promotion, and checks for missed expirations of raft peers. The end result of this loop is that the raft membership should be in sync with the gossip membership.

### Testing
- Tested on a variety of join/leave cases -- in each case the cluster stabilized as expected
- Unit tests pass


state/src/replication/raft.rs
```diff
@@ -414,27 +434,82 @@ impl RaftClient {
...
+        // may be made at a time
+        if self.add_missed_learners(&known_peers).await? > 0 {
+            return Ok(());
+        }
+
+        if self.try_promote_learners().await? > 0 {
+            return Ok(());
+        }
+
+        self.check_expired_nodes(&known_peers).await
```
akirillo, line 465: IMO we should check for expired nodes first - in the case that there are both learners to add and peers to expire, we don't want to try replicating membership changes to nodes that are down. We also get to avoid sending a membership to change to the new learners when these peers are expired

akirillo (approved):
couple small remarks but lgtm, ty for wrapping up & testing

state/src/replication/raft.rs
joeykraut: good point, updated

PR #658, opened 2024-06-07
task-driver: tasks: refresh-wallet: Implement wallet refresh task

joeykraut:
### Purpose
This PR moves the wallet refresh logic to a separate task that better handles edge cases like an already up-to-date wallet. This task will lookup the wallet on-chain much like the `LookupWalletTask` but more directly uses the wallet information already in state.

Clearing the queue via the API causes running tasks to fail. For this reason, I transitioned some errors in the task queue applicator to be warnings or log rejections so as not to panic the raft.

### Testing
- Tested refreshing a wallet in a variety of cases, and verified that errors were appropriately handled:
    - While an `UpdateWallet` is running
    - While a `SettleInternalMatch` is running
    - While no tasks are running
- Unit tests pass

### DIFF
--- a/workers/api-server/src/http/wallet.rs
+++ b/workers/api-server/src/http/wallet.rs
@@ -8,7 +8,7 @@ use circuit_types::{
 use common::types::{
     tasks::{
         LookupWalletTaskDescriptor, NewWalletTaskDescriptor, PayOfflineFeeTaskDescriptor,
-        TaskDescriptor, TaskIdentifier, UpdateWalletTaskDescriptor,
+        RefreshWalletTaskDescriptor, TaskDescriptor, TaskIdentifier, UpdateWalletTaskDescriptor,
     },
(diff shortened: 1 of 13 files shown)


workers/api-server/src/http/wallet.rs
```diff
@@ -282,19 +282,14 @@ impl TypedHandler for RefreshWalletHandler {
         _query_params: QueryParams,
     ) -> Result<Self::Response, ApiServerError> {
         let wallet_id = parse_wallet_id_from_params(&params)?;
-        let wallet = find_wallet_for_update(wallet_id, &self.state).await?;
+        find_wallet_for_update(wallet_id, &self.state).await?;
```
akirillo, line 285: dead code?

akirillo (approved):

workers/api-server/src/http/wallet.rs
joeykraut: no, this now just effectively checks that the wallet is present, it's necessary

PR #659, opened 2024-06-08
state: replication, workers: task-driver: node-startup: remove learners separately, only setup wallet if leader

akirillo:
This PR makes a couple tweaks to how we set up & maintain a Raft cluster, namely:
1. In the node startup task, we only transition to setting up a relayer wallet if the local nodes is the elected leader. It is possible for a node to attempt to initialize a Raft if the true leader takes too long to begin sending it logs, but when it succeeds in doing so, the new node should await voter promotion
2. We correctly remove expired learners via the `sync_membership` loop, and we add batched semantics to the `RemovePeers` state transition. We previously would never correctly remove learners.

All state unit tests pass, and I am testing the effect on thrash in dev.

### DIFF
--- a/state/src/replication/raft.rs
+++ b/state/src/replication/raft.rs
@@ -252,8 +252,8 @@ impl RaftClient {
             .ok_or_else(|| ReplicationV2Error::Proposal(ERR_NO_LEADER.to_string()))?;
 
         // If we're expiring the leader, first change leader then propose an expiry
-        if let &StateTransition::RemoveRaftPeer { peer_id } = update.transition.as_ref()
-            && leader_nid == peer_id
+        if let StateTransition::RemoveRaftPeers { peer_ids } = update.transition.as_ref()
+            && peer_ids.contains(&leader_nid)
         {
             info!("removing raft leader");
             self.change_leader().await?;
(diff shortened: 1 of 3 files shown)


state/src/replication/raft.rs
```diff
@@ -312,22 +314,47 @@ impl RaftClient {
...
+
+        // Remove voters
+        if !voters_to_remove.is_empty() {
+            let voter_removal_change =
+                ChangeMembers::RemoveVoters(BTreeSet::from_iter(voters_to_remove.into_iter()));
+            self.raft()
+                .change_membership(voter_removal_change, false /* retain */)
+                .await
+                .map_err(err_str!(ReplicationV2Error::Proposal))
+                .map(|_| ())?;
```
joeykraut, line 338: do we need this map? we don't use the return value

state/src/replication/raft.rs
```diff
@@ -312,22 +314,47 @@ impl RaftClient {
...
+
+        // Remove learners
+        if !learners_to_remove.is_empty() {
+            let learner_removal_change =
+                ChangeMembers::RemoveNodes(BTreeSet::from_iter(learners_to_remove.into_iter()));
+            self.raft()
+                .change_membership(learner_removal_change, false /* retain */)
+                .await
+                .map_err(err_str!(ReplicationV2Error::Proposal))
+                .map(|_| ())?;
```
joeykraut, line 349: same here

joeykraut (approved):
Looks good, lets make sure to test extensively in `dev` before merging

PR #660, opened 2024-06-08
util: matching-engine: Use fixed point floor div when generating amounts

joeykraut:
### Purpose
This PR changes the implementation of `compute_max_amount` in the matching engine to use `FixedPoint` floor division rather than `f64` floor division. This fixes a bug we saw in a deployment wherein the matching engine generates invalid witnesses, which then fail to prove.

### Testing
- I added a test using an order and balance I saw fail in a load test, verified that `compute_max_amount` returns a valid value now.
- All tests pass

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1818,6 +1818,7 @@ dependencies = [
  "mpc-plonk",
  "mpc-relation",
  "num-bigint",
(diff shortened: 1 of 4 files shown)


akirillo (approved):
good find

PR #662, opened 2024-06-10
workers: api-server: Add CORS header on non-200 responses

sehyunc:
This PR adds CORS headers to non-200 responses so that the client can read the error message on failed requests.

Tested with local relayer and frontend.

### DIFF
--- a/workers/api-server/src/router.rs
+++ b/workers/api-server/src/router.rs
@@ -35,17 +35,17 @@ const ERR_INVALID_QUERY_PARAMS: &str = "invalid query params";
 
 /// Builds an empty HTTP 400 (Bad Request) response
 pub(super) fn build_400_response(err: String) -> Response<Body> {
-    Response::builder().status(StatusCode::BAD_REQUEST).body(Body::from(err)).unwrap()
+    build_response_from_status_code(StatusCode::BAD_REQUEST, err)
 }
 
 /// Builds an empty HTTP 404 (Not Found) response
 pub(super) fn build_404_response(err: String) -> Response<Body> {
-    Response::builder().status(StatusCode::NOT_FOUND).body(Body::from(err)).unwrap()
+    build_response_from_status_code(StatusCode::NOT_FOUND, err)
 }
 
 /// Builds an empty HTTP 500 (Internal Server Error) response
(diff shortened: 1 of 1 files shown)


workers/api-server/src/router.rs
```diff
@@ -35,25 +35,41 @@ const ERR_INVALID_QUERY_PARAMS: &str = "invalid query params";
 
 /// Builds an empty HTTP 400 (Bad Request) response
 pub(super) fn build_400_response(err: String) -> Response<Body> {
-    Response::builder().status(StatusCode::BAD_REQUEST).body(Body::from(err)).unwrap()
+    Response::builder()
+        .status(StatusCode::BAD_REQUEST)
+        .header("Access-Control-Allow-Origin", "*")
+        .body(Body::from(err))
+        .unwrap()
```
joeykraut, line 42: nit: I suggest refactoring this into a helper like `build_err_resp_with_status(status: StatusCode, err: String)` given that all these bodies do the same thing with different status codes

joeykraut (approved):
one nit, o/w lgtm

workers/api-server/src/router.rs
sehyunc: Replaced helper functions with already existing `build_response_from_status_code` fn

workers/api-server/src/router.rs
```diff
@@ -33,27 +33,16 @@ const ERR_INVALID_QUERY_PARAMS: &str = "invalid query params";
...
-
-/// Builds an empty HTTP 404 (Not Found) response
-pub(super) fn build_404_response(err: String) -> Response<Body> {
-    Response::builder().status(StatusCode::NOT_FOUND).body(Body::from(err)).unwrap()
-}
-
-/// Builds an empty HTTP 500 (Internal Server Error) response
-pub(super) fn build_500_response(err: String) -> Response<Body> {
-    Response::builder().status(StatusCode::INTERNAL_SERVER_ERROR).body(Body::from(err)).unwrap()
-}
```
joeykraut, line 49: I would still use these methods to be clear, they might be useful when building other routes that return 404, 400, etc. So I wouldn't get rid of them, maybe just change their implementation to call out to `build_response_from_status_code`

PR #663, opened 2024-06-10
state: storage: tx: relayer-fees: Add storage methods for relayer fees

joeykraut:
### Purpose
This PR adds storage helpers for relayer fees. This allows us to set the `match-fee` field (representing the relayer's take rate) on a per-wallet basis. We will use this to implement a `fee-whitelist.json` file in the config, which allows us to set per-wallet fees.

### Testing
- Unit tests pass

### DIFF
--- a/state/src/storage/tx/relayer_fees.rs
+++ b/state/src/storage/tx/relayer_fees.rs
@@ -1,0 +1,98 @@
+//! Storage helpers for relayer fees
+
+use circuit_types::fixed_point::FixedPoint;
+use common::types::wallet::WalletIdentifier;
+use libmdbx::{TransactionKind, RW};
+
+use crate::{storage::error::StorageError, RELAYER_FEES_TABLE};
+
+use super::StateTxn;
+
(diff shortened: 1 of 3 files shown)


state/src/storage/tx/relayer_fees.rs
```diff
@@ -0,0 +1,98 @@
...
+        // Setup the default fee
+        let fee = FixedPoint::from_f64_round_down(0.01);
+        let tx = db.new_write_tx().unwrap();
+        tx.set_relayer_take_rate(&fee).unwrap();
+        tx.commit().unwrap();
+
+        // Ensure the fee is set correctly
+        let tx = db.new_read_tx().unwrap();
+        let fee = tx.get_relayer_fee(&wallet_id).unwrap();
+        assert_eq!(fee, fee);
```
akirillo, line 77: this assertion will trivially pass, let's rename the variable for the fee fetched via `get_relayer_fee`

akirillo (approved):

PR #664, opened 2024-06-10
config: Add `relayer-fee-whitelist` option and use whitelisted fees in `new-wallet` creation

joeykraut:
### Purpose
This PR adds a `--relayer-fee-whitelist` config option which allows the parent process to specify a `.json` file containing whitelist entries like the following:
```json
{
    "wallet_id": "id",
    "fee": 0.0005,
    ...
}
```

These values are parsed and stored in state. Upon wallet creation, a fee is looked up (defaulting to the relayer's base fee) and the wallet's `match_fee` field is set using this value.

### Testing
- Added a wallet to the whitelist and created a new wallet under that ID, verified that it was given the whitelisted fee
- Added a new wallet not in the whitelist, verified that it was given the base fee

### DIFF
--- a/config/src/cli.rs
+++ b/config/src/cli.rs
@@ -19,7 +19,7 @@ use std::{
 };
 use url::Url;
 
-use crate::parsing::parse_config_from_args;
+use crate::parsing::{parse_config_from_args, RelayerFeeWhitelistEntry};
 
 // -------
 // | CLI |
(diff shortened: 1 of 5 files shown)


state/src/interface/relayer_fees.rs
akirillo, line 1: is this file meant to be kept around?

akirillo (approved):

state/src/interface/relayer_fees.rs
joeykraut: oops, no, thank you

PR #666, opened 2024-06-11
external-api, common: tasks: Record fee payment mint + amount in history

joeykraut:
### Purpose
This PR adds the `amount`, `mint`, and `is_protocol` fields to the historical fee payment task description. This will allow for these fields to be consumed via the API and displayed on the FE.

### Testing
- Tested via the API locally

### DIFF
--- a/workers/task-driver/src/tasks/pay_offline_fee.rs
+++ b/workers/task-driver/src/tasks/pay_offline_fee.rs
@@ -39,6 +39,10 @@ use super::{ERR_BALANCE_MISSING, ERR_NO_MERKLE_PROOF, ERR_WALLET_MISSING};
 /// The name of the task
 const TASK_NAME: &str = "pay-offline-fee";
 
+/// Error message emitted when the fee amount in the descriptor is more than the
+/// fees owed
+const ERR_INVALID_FEE_AMOUNT: &str = "Fee amount in descriptor does not equal paid amount";
+
 // --------------
 // | Task State |
 // --------------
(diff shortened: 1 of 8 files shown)


workers/task-driver/src/tasks/pay_offline_fee.rs
```diff
@@ -175,15 +179,14 @@ impl Task for PayOfflineFeeTask {
             .ok_or_else(|| PayOfflineFeeTaskError::State(ERR_WALLET_MISSING.to_string()))?;
 
         // Construct the new wallet
-        let (note, new_wallet) = Self::get_wallet_and_note(
-            descriptor.is_protocol_fee,
-            &descriptor.balance_mint,
-            &old_wallet,
-        )?;
+        let (note, new_wallet) = Self::get_wallet_and_note(&descriptor, &old_wallet)?;
+        if descriptor.amount != note.amount {
```
akirillo, line 183: hmm, let's say we have fee payment tasks enqueued, and a match (for the fee assets) preempts the queue. then this error will be tripped. this isn't a huge deal since the FE can retry the fee payment task, but it doesn't seem like it should cause an error. is it too overbearing to have historical task descriptors be calculated from tasks as opposed to task descriptors?

akirillo (approved):

workers/task-driver/src/tasks/pay_offline_fee.rs
joeykraut: well the case you described is quite the edge case and i think it's okay if that fails; note that a settle match won't be started if this task exists. 

It would end up being quite the refactor, getting all this information where it needs to be. And I don't love the idea of expanding the scope of this PR to cover a rare edge case

PR #667, opened 2024-06-11
state: storage: order-history: Record partial fills in order history

joeykraut:
### Purpose
This PR adds partial fill information to historical orders. Each partial fill contains the amount, price, and timestamp of the partial fill.

### Testing
- Unit tests pass
- Tested via the CLI with unmatched, incrementally matched, and fully matched orders

### DIFF
--- a/state/src/interface/wallet_index.rs
+++ b/state/src/interface/wallet_index.rs
@@ -107,10 +107,10 @@ mod test {
     // Wallet update handlers in the applicator handle order history changes to
     // hide the wallet index/order history denormalization
 
-    use circuit_types::balance::Balance;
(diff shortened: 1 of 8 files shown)


state/src/interface/wallet_index.rs
```diff
@@ -126,7 +126,7 @@ mod test {
             .map(|_| OrderMetadata {
                 id: OrderIdentifier::new_v4(),
                 state: OrderState::Filled,
-                filled: 1,
+                fills: vec![],
```
akirillo, line 129: Should this have a partial fill w amount 1? If tests pass tho ofc doesn't matter

akirillo (approved):
Clean

PR #668, opened 2024-06-11
state: replication, workers: gossip-server: gossip/raft stability tweaks

akirillo:
This PR makes a number of small tweaks to raft & gossip functionality to improve stability, particularly in times of concurrent join/leave, such as an upgrade. Namely, these changes are:
- Raft changes
  - Add a lock around the application of cluster membership changes. The `openraft` implementation awaits the commitment of membership changes to the cluster before others can be added, so we have to serialize membership changes to avoid them being dropped. A lock is the simplest way to achieve this.
  - Decrease the raft heartbeat interval to 1s. In awaiting the commitment of a cluster membership change, `openraft` awaits for a response _or timeout_ from each of the known peers. Thus, we previously would wait 6s for a cluster membership change including a removal to be processed (because `openraft` commits the same membership twice to move out of joint config)
  - Remove redundant membership changes at the time of application. We get proposals to add/remove peers from all other peers in the cluster, and `openraft` will _still_ commit these membership logs even if they have no effect on the membership
- Gossip changes
  - Exclude expiry candidates from the `known_peers` included in a gossip heartbeat. This only adds to the potential thrash of the cluster by having new nodes add peers that will likely be promptly removed. If the peers are not removed, then new nodes will soon discover them via the heartbeat protocol.
  - Don't mark fresh heartbeats when adding a peer. This was a subtle but insidious feature that caused expiry candidates to be rejected when new nodes join the cluster, leading to a lot of failed replication and thrash
  - Increase the expiry invisibility window to 1 minute. We would see dead peers get re-added right outside of the invisibility window, in part due to the fact that they were re-added due to the fresh heartbeats on addition by new nodes that was also removed in this PR. Setting a 1 minute invisibility window gives some more reasonable breathing room, and in our clusters we don't reuse peer IDs, anyway
  - Remove expiry candidates when we receive heartbeats from them. This would sometimes lead to incorrect expiries of peers when new nodes come online, discover peers - but _don't_ mark a fresh heartbeat - and propose an expiry before they exchange heartbeats with the peers because of this. The expiry proposal is seen by no one but still counts down

The sum total of these changes has been tested in dev, by upgrading a cluster w/ snapshots and a fair amount of load on it. The upgrade went through very smoothly with no thrash. Many other upgrade tests were run without snapshots or load, which were all also very smooth.


state/src/replication/raft.rs
```diff
@@ -293,39 +301,67 @@ impl RaftClient {
...
-            .change_membership(change, true /* retain */)
-            .await
-            .map_err(err_str!(ReplicationV2Error::Proposal))
-            .map(|_| ())
+        let _lock = self.membership_change_lock.lock().await;
+
+        // Only add learners that haven't already been added
+        let membership = self.membership();
+        let mut to_add = Vec::new();
+        for learner in learners {
```
joeykraut, line 309: nit: `for (node_id, _info) in learners.iter()` is more readable

workers/gossip-server/src/peer_discovery/expiry_window.rs
```diff
@@ -12,10 +12,10 @@ use tokio::sync::{RwLockReadGuard, RwLockWriteGuard};
 
 /// The amount of time other cluster peers are allowed to give liveness
 /// attestations for an expiry candidate
-pub(crate) const EXPIRY_CANDIDATE_WINDOW_MS: u64 = 10_000; // 10 seconds
+pub(crate) const EXPIRY_CANDIDATE_WINDOW_MS: u64 = 15_000; // 15 seconds
```
joeykraut, line 15: why? this means we expire a peer 1.5x slower, is there a good reason for this change? 10s should be more than enough time for peers to contest an expiry. If anything we should be lowering this

workers/gossip-server/src/peer_discovery/peers.rs
```diff
@@ -159,6 +150,11 @@ impl GossipProtocolExecutor {
                 continue;
             }
 
+            // Check that the peer is not already in the peer index
+            if self.state.get_peer_info(&peer.peer_id).await?.is_some() {
+                continue;
+            }
```
joeykraut, line 156: why do we want to do this? and if we do, I'd suggest moving this to `state`, if this solves a bug it should be solved in the `state` interface

joeykraut (approved):
few suggestions, o/w lgtm

workers/gossip-server/src/peer_discovery/peers.rs
akirillo: we do this because the re-addition of a peer overwrites our local view of the peer's last heartbeat with whatever was sent in the payload here. this can lead to us writing in stale heartbeat timestamps, as opposed to the local node's actual view.

I did this here as opposed to `state.add_peers_batch` since nodes filtered according to this clause shouldn't have their addresses written in via `add_new_addrs`, but now that I mention it, we should probably decouple these two notions. i.e. avoid overwriting heartbeats for a node we've already indexed to maintain an accurate view of liveness, but allow the other metadata to be updated.

Will do this in `state`

workers/gossip-server/src/peer_discovery/expiry_window.rs
akirillo: now that we don't mark fresh heartbeats when adding a peer, new nodes that are still warming up into gossip are quick to expire discovered peers since they are using heartbeat timestamps from received peer info payloads. this change is to allow for a little more time for new peers to exchange heartbeats w/ peers they discover so they don't preemptively expire them

akirillo: as per internal discussion, will decrease back to 10s and run another test. this should be enough time for heartbeat exchange, and if not, a more appropriate approach would be to disallow expiry proposals during gossip warmup (not implemented in this PR due to complexity)

PR #671, opened 2024-06-12
state: interface: add timer for raft metrics

akirillo:
This PR adds a timer loop that records metrics about Raft cluster composition, namely:
- The local node's view of the cluster size
- The local node's view of whether or not it's a leader

We'll use both of these metrics to keep track of split brains and general cluster health. For the latter metric, we already have an `is_leader` endpoint that we could use to check for the existence of a leader, but recording a 0/1 for if the local node is a leader or not lets us sum this metric across the cluster to see if there are multiple leaders, a clear split brain.

I tested these metrics by running a local cluster and asserting that they are reported expectedly.


joeykraut (approved):
nice, i like the approach

PR #672, opened 2024-06-12
common: wallet: order-metadata: Report partial fill price as f64

joeykraut:
### Purpose
This PR changes the type used to store prices at which partial fills executed from `FixedPoint` to `f64`. This makes for a more interpretable, readable API response; and the partial fills are only used for historical reference.

### Testing
- Unit tests pass
- Tested the endpoint locally


sehyunc (approved):
LGTM

PR #679, opened 2024-06-18
util: hex, circuit-types: ensure biguint -> address conversion has proper padding

akirillo:
This PR replaces the `biguint_to_hex_str` conversion method w/ a specialized `biguint_to_hex_addr` one that preserves padding to 20 bytes for addresses where appropriate. The native `BigUint::to_str_radix` would trim leading zeroes, leading to issues across the stack when token addresses had leading zeroes, e.g. inability to find tokens by address in the token mapping, and returning trimmed addresses via the API which the frontend would fail to match to its own token mapping.

All unit tests pass, and this is being tested in the mini-env.

### DIFF
--- a/util/src/hex.rs
+++ b/util/src/hex.rs
@@ -5,7 +5,7 @@ use circuit_types::{
     elgamal::BabyJubJubPoint,
     keychain::{NonNativeScalar, PublicSigningKey},
 };
-use constants::{EmbeddedCurveConfig, Scalar};
+use constants::{EmbeddedCurveConfig, Scalar, ADDRESS_BYTE_LENGTH};
 use num_bigint::BigUint;
 use num_traits::Num;
 use renegade_crypto::fields::{biguint_to_scalar, scalar_to_biguint};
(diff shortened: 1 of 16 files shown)


util/src/hex.rs
```diff
@@ -17,6 +17,20 @@ pub fn biguint_to_hex_string(val: &BigUint) -> String {
...
 }
 
+/// From a BigUint, get a lowercase hex string with a 0x prefix, padded to the
+/// Ethereum address length
+pub fn biguint_to_hex_addr(val: &BigUint) -> String {
+    let mut bytes = [0_u8; ADDRESS_BYTE_LENGTH];
+    let val_bytes = val.to_bytes_be();
+
+    let len = val_bytes.len();
+    assert!(len <= ADDRESS_BYTE_LENGTH, "BigUint too large for an address");
```
joeykraut, line 27: Nit: can this be a debug assert?

joeykraut (approved):

PR #680, opened 2024-06-22
common: types, workers: handshake-manager: use timestamped price in order metadata

akirillo:
This PR introduces the `TimestampedPrice` struct and uses it in the `SettleMatch` and `SettleMatchInternal` tasks with the purpose of it being used in the `PartialOrderFill` struct.

The goal here is to communicate to the frontend the time at which the price used in each of an order's partial fills was sampled. This allows us to more accurately estimate the savings relative to hypothetically executing the fill on Binance.

We use the timestamp of the price sampling, as opposed to the timestamp of the fill execution, as we deem that to be the point at which the execution process for a fill begins. In calculating the savings, we want to communicate to the user: "if you began the execution process at this time in Renegade, you'd get this price, if you began it at this time in Binance, you'd get this price".

All unit tests pass, and this is currently being tested in dev.

Note: because this changes a type that is communicated both in gossip and is stored directly in the DB, upgrading testnet to use this change will require an intermediary version to prevent serde issues

### DIFF
--- a/common/src/types.rs
+++ b/common/src/types.rs
@@ -1,4 +1,7 @@
 //! Defines common types that many crates can depend on
+
+use serde::{Deserialize, Serialize};
+
 pub mod exchange;
 pub mod gossip;
 pub mod handshake;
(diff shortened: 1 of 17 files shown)


joeykraut (approved):
lgtm

akirillo:
Going to leave this unmerged until the frontend is ready to ingest these changes so we don't risk deploying it to testnet

PR #683, opened 2024-06-25
state: applicator, workers: task-driver: pop tasks that cannot be constructed

akirillo:
This PR ensures that we clean up tasks that failed to be constructed from their descriptors. Concretely we propose a `PopTask` state transition for them, which handles removing them from the queue, cleaning up their assignment, clearing the queue if they failed (in this case, they did), etc.

We only do this for non-preemptive tasks, as this logic is executed in task queue resumption for preemptive tasks. On that note, we do this unilaterally for tasks that do and do not bypass the task queue. We ensure that this is safe by having the applicator gracefully reject the state transition if the task to be popped cannot be found at the top of its queue (e.g. if it is a task that bypasses the queue).

All unit tests pass (except for `test_raft_scale_down`, though that should be unaffected by these changes), and I added an extra unit test asserting that the applicator rejects when double-popping the same task

### DIFF
--- a/state/src/applicator/task_queue.rs
+++ b/state/src/applicator/task_queue.rs
@@ -45,6 +45,11 @@ fn task_not_running(task_id: TaskIdentifier) -> String {
     format!("task {task_id} is not running")
 }
 
+/// Construct a task not in queue error
+fn task_not_in_queue(task_id: TaskIdentifier, queue_key: TaskQueueKey) -> String {
+    format!("task {task_id} is not in queue {queue_key}")
+}
+
 // -----------
 // | Helpers |
 // -----------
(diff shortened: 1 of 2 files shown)


state/src/applicator/task_queue.rs
```diff
@@ -128,6 +133,18 @@ impl StateApplicator {
...
+        // transition
+        let queued_tasks = tx.get_queued_tasks(&key)?;
+        let top_of_queue = queued_tasks.first();
+        match top_of_queue {
+            None => return Err(StateApplicatorError::Rejected(task_not_in_queue(task_id, key))),
+            Some(task) if task.id != task_id => {
+                return Err(StateApplicatorError::Rejected(task_not_in_queue(task_id, key)));
+            },
+            _ => {},
+        }
```
joeykraut, line 146: nit: up to preference but could rewrite this as

```rust
if top_of_queue.is_none() || top_of_queue.unwap().id != task_id {
    return ...;
}
```

slightly more readable IMO

joeykraut (approved):

PR #685, opened 2024-06-26
state: replication: clear tables instead of dropping & remove db when applying snapshots

akirillo:
This PR makes the following 2 changes to how snapshots are handled:
1. When applying an installed snapshot, we use MDBX's `clear_table` method instead of dropping & re-adding a table. We are seeing `BadDbi` errors occasionally after installing snapshots that are generally linked to accessing tables that are deleted ([ex](https://github.com/erthink/libmdbx/issues/146))
2. When updating from a snapshot, either when recovering from one on startup or after installing one, we delete the snapshot data afterwards to save on disk space in the relayer

Things are working smoothly in dev, though it has proven difficult to reproduce the issue that prompted these changes even before deploying them to dev.


joeykraut (approved):
nice one

PR #687, opened 2024-06-27
workers: network-manager: composed_protocol: bump request-response timeout

akirillo:
This PR bumps the request/response timeout on the libp2p behavior to 1 minute from the default of 10s. In the testnet cluster, we see failures to send `InstallSnapshot` RPC responses when the application of the snapshot takes over 10s - this is intended to alleviate that.

This has been tested in dev w/ orderbot load and we saw no service degradation.

### DIFF
--- a/workers/network-manager/src/composed_protocol.rs
+++ b/workers/network-manager/src/composed_protocol.rs
@@ -20,8 +20,9 @@ use libp2p::{
     identity::Keypair,
     kad::{record::store::MemoryStore, Kademlia, KademliaEvent},
     request_response::{
-        Behaviour as RequestResponse, Codec as RequestResponseCodec, Event as RequestResponseEvent,
-        ProtocolName, ProtocolSupport,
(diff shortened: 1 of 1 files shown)


workers/network-manager/src/composed_protocol.rs
```diff
@@ -67,10 +69,12 @@ impl ComposedNetworkBehavior {
         keypair: &Keypair,
     ) -> Result<Self, NetworkManagerError> {
         // Construct the point-to-point request response protocol
+        let mut request_response_config: RequestResponseConfig = Default::default();
+        request_response_config.set_request_timeout(Duration::from_secs(60));
```
joeykraut, line 73: nit: can we make this a const?

joeykraut (approved):
looks great

workers/network-manager/src/composed_protocol.rs
akirillo: ack

PR #691, opened 2024-06-29
api-server: compliance: Check wallet screening compliance if configured

joeykraut:
### Purpose
This PR adds a `--compliance-service-url` config option to the the relayer which, when set, enables compliance checks on deposits. The only requirement on the config is that the server on the other end must implement the [compliance-server-api](https://github.com/renegade-fi/relayer-extensions/tree/master/compliance/compliance-api).

If not set, compliance screens are not enabled.

### Testing
- Tested a deposit with no `--compliance-service-url` specified
- Tested a deposit on a low risk wallet with `--compliance-service-url` pointing to a locally running compliance service
- Tested a deposit on a high risk wallet with `--compliance-service-url` pointing to a locally running service. Received http 400

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -95,9 +95,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-rlp"
-version = "0.3.5"
+version = "0.3.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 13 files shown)


akirillo (approved):
lgtm but need to update mock node

PR #692, opened 2024-06-29
state: storage: matching_pools: create pool table & barebones access methods

akirillo:
This PR initializes the storage layer for the matching pools feature, establishing the DB layout for the `POOL_TABLE` and the minimum necessary access methods for the expected usage.

This includes some basic unit tests, all of which pass.

Next up is the interface layer for interacting with matching pools state.

### DIFF
--- a/state/src/storage/tx/matching_pools.rs
+++ b/state/src/storage/tx/matching_pools.rs
@@ -1,0 +1,209 @@
+//! Helpers for accessing information about matching pools in the database
+
+use common::types::wallet::OrderIdentifier;
+use libmdbx::{TransactionKind, RW};
+
+use crate::{storage::error::StorageError, POOL_TABLE};
+
+use super::StateTxn;
+
+// -------------
+// | Constants |
+// -------------
+
+/// The name of the global matching pool
+pub const GLOBAL_MATCHING_POOL: &str = "global";
+
+/// The prefix of all matching pool keys
+pub const POOL_KEY_PREFIX: &str = "matching-pool/";
+
+/// The error message used when trying to create a matching pool that already
+/// exists
+const MATCHING_POOL_EXISTS_ERR: &str = "matching pool already exists";
+/// The error message used when assigning an order to a nonexistent matching
+/// pool
+const MATCHING_POOL_DOES_NOT_EXIST_ERR: &str = "matching pool does not exist";
+/// The error message used when a non-empty matching pool is attempted to be
+/// destroyed
+const MATCHING_POOL_NOT_EMPTY_ERR: &str = "matching pool not empty";
+
+// ---------------
+// | Key Helpers |
+// ---------------
+
+/// The key for the set of all matching pools
+pub fn all_matching_pools_key() -> String {
+    "all-matching-pools".to_string()
+}
+
+/// The key for the given order's matching pool
+pub fn matching_pool_key(order_id: &OrderIdentifier) -> String {
+    format!("{}{}", POOL_KEY_PREFIX, order_id)
+}
+
+// -----------
+// | Getters |
+// -----------
+
+impl<'db, T: TransactionKind> StateTxn<'db, T> {
+    /// Get the name of the matching pool the given order is in, if it's been
+    /// assigned to one
+    pub fn get_matching_pool_for_order(
+        &self,
+        order_id: &OrderIdentifier,
+    ) -> Result<Option<String>, StorageError> {
+        let pool_key = matching_pool_key(order_id);
+        self.inner().read(POOL_TABLE, &pool_key)
+    }
+
+    /// Whether or not a pool with the given name exists
+    pub fn matching_pool_exists(&self, pool_name: &str) -> Result<bool, StorageError> {
+        let all_pools_key = all_matching_pools_key();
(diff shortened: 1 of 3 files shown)


state/src/storage/tx/matching_pools.rs
```diff
@@ -0,0 +1,206 @@
...
+use crate::{storage::error::StorageError, POOL_TABLE};
+
+use super::StateTxn;
+
+// -------------
+// | Constants |
+// -------------
+
+/// The name of the global matching pool
+pub const GLOBAL_MATCHING_POOL: &str = "global";
```
joeykraut, line 15: do we need to define this at the storage layer, it seems only used in tests. if we intend to use this at higher layers, it probably makes sense to define it there.

state/src/storage/tx/matching_pools.rs
```diff
@@ -0,0 +1,206 @@
...
+
+    /// Whether or not a pool with the given name exists
+    pub fn pool_exists(&self, pool_name: &String) -> Result<bool, StorageError> {
+        let all_pools_key = all_pools_key();
+        let all_pools = self.read_set(POOL_TABLE, &all_pools_key)?;
+        Ok(all_pools.contains(pool_name))
+    }
+
+    /// Whether or not the matching pool is empty
+    pub fn pool_is_empty(&self, pool_name: &String) -> Result<bool, StorageError> {
```
joeykraut, line 64: any reason this is `&String` and not `&str`? I know we require a non-reference type for DB access but you can see how other methods handle this with string keys at the storage layer; i.e. `&s.to_string()`. I feel this gives a cleaner interface as it accepts a wider range of types, and removes the need to do `to_string` in the caller. Plus would be nice to be consistent across the storage layer

state/src/storage/tx/matching_pools.rs
```diff
@@ -0,0 +1,206 @@
...
+
+        let all_pools_key = all_pools_key();
+        self.remove_from_set(POOL_TABLE, &all_pools_key, pool_name)
+    }
+
+    /// Allocate an order to a matching pool.
+    ///
+    /// Note that we allow overwriting an order's pool, this signifies moving
+    /// the order from one pool to another.
+    pub fn allocate_order_to_pool(
```
joeykraut, line 115: nit: i'd prefer `assign`; i don't think of "allocating" and object "to" another object, assigning is more accurate. Plus no order is "allocated"

state/src/storage/tx/matching_pools.rs
```diff
@@ -0,0 +1,206 @@
...
+}
+
+// -----------
+// | Getters |
+// -----------
+
+impl<'db, T: TransactionKind> StateTxn<'db, T> {
+    /// Get the name of the matching pool the given order is in, if it's been
+    /// allocated to one
+    pub fn get_pool_for_order(
```
joeykraut, line 48: can we do `:s/pool/matching_pool`? The naming is fair in the context of this file, but when I look at the documentation of the `StateTxn` object, i see "pool" thrown around in signatures, and it's unclear what the "pool" is. It's clearer if these are labelled `matching_pool`. Things like `pool_exists` are underspecified for example, so i propose referring to pools as matching pools across the interface.

state/src/storage/tx/matching_pools.rs
```diff
@@ -0,0 +1,206 @@
...
+/// The key for the given order's matching pool
+pub fn matching_pool_key(order_id: &OrderIdentifier) -> String {
+    format!("{}{}", POOL_KEY_PREFIX, order_id)
+}
+
+// -----------
+// | Getters |
+// -----------
+
+impl<'db, T: TransactionKind> StateTxn<'db, T> {
```
joeykraut, line 45: perhaps deferred to a follow-up, but I assume we want a getter for all the orders in a pool?

joeykraut (commented):
very nice, suggestions are all small

joeykraut (approved):

state/src/storage/tx/matching_pools.rs
akirillo: yes ack, as I massage this into the relayer I'll find a better home for it (perhaps the `interface` layer)

state/src/storage/tx/matching_pools.rs
akirillo: yeah, honestly was just nice to leverage type inference on the tx/cursor methods, but can use `&str` here

state/src/storage/tx/matching_pools.rs
akirillo: sure sgtm

state/src/storage/tx/matching_pools.rs
akirillo: actually, i'm not sure that we do. the intended usage is in the matching engine, where we get all of the local orders, and in the filtering applied to them [here](https://github.com/renegade-fi/renegade/blob/main/state/src/interface/order_book.rs#L181), we simply lookup the candidate order's matching pool, and see if it's equal to that of the order being matched.

the fact that that is the primary access pattern is the reason I went with a `order -> matching pool` lookup instead of the other direction

PR #695, opened 2024-07-01
workers: task-driver: node_startup: create global matching pool

akirillo:
This PR adds a step to the node startup task which creates the global matching pool. Some points of note:
1. We create the global matching pool iff 1) we initialized a raft && 2) we did not recover from a snapshot. In all other cases the global matching pool should already exist in the node's state by the time it gets to this task step.
2. We create the global matching pool _before_ setting up the relayer wallet. In the off chance that a looked up relayer wallet has orders in it, the global pool should already exist so those orders can be assigned to it when they're indexed.

I tested this by running a local node, it got through this task step with no issues.


workers/task-driver/src/tasks/node_startup.rs
```diff
@@ -333,12 +340,27 @@ impl NodeStartupTask {
...
+    /// Create the global matching pool
+    async fn create_global_matching_pool(&self) -> Result<(), NodeStartupTaskError> {
+        // If the node recovered from a snapshot, we do not need to re-create the global
+        // matching pool
+        if self.state.was_recovered_from_snapshot() {
+            return Ok(());
+        }
+
+        // Otherwise, create the global matching pool
+        let waiter = self.state.create_global_matching_pool().await?;
```
joeykraut, line 358: This should probably just be done in [setup_node_metadata](https://github.com/renegade-fi/renegade/blob/caa46562c7ae1e636d033feb6a4ffb524ca03b73/state/src/interface/node_metadata.rs#L104) (maybe we can rename it to `setup_node_state` or something. It doesn't really make sense to send this through consensus in my opinion. Given that it's static information, the best way to achieve consensus is to put it in code!

akirillo: hmm, fair point, yeah. will make this change.

state/src/interface/matching_pools.rs
```diff
@@ -68,55 +68,4 @@ impl State {
...
-        self.create_matching_pool(GLOBAL_MATCHING_POOL.to_string()).await
-    }
-
-    /// Assign an order to the global matching pool
-    pub async fn assign_order_to_global_matching_pool(
-        &self,
-        order_id: OrderIdentifier,
-    ) -> Result<ProposalWaiter, StateError> {
-        self.assign_order_to_matching_pool(order_id, GLOBAL_MATCHING_POOL.to_string()).await
-    }
```
akirillo, line 83: I've removed these two methods from the state interface because we effectively never use them. We create the global matching pool & assign orders to it directly via the `tx` (the latter is in the next PR)

joeykraut (approved):

PR #696, opened 2024-07-01
workers: task-driver: update_wallet: order matching pool assignment

akirillo:
This PR implements the logic for assigning orders to the appropriate matching pool. The logic is as follows: by default, indexed new orders are added to the global matching pool. However, we add the desired matching pool for an order to be added to the `UpdateWalletTaskDescriptor`, and in the `UpdateWalletTask` we subsequently assign a new order to the desired matching pool after it is indexed.

This gets us 2 things:
1. In the "happy path," orders that are created in the global matching pool do not need to invoke another state transition for being assigned to the global matching pool.
2. Orders that are intended for non-global matching pools are assigned to those pools during the `UpdateWalletTask`, i.e. there is no danger of a race in which we begin matching those orders across the (default) global matching pool before they are assigned to the correct one.

All unit tests pass, and I updated the `test_update_wallet` test to also check that a new order is assigned to the global matching pool by default. Because integration tests are blocked at the moment, testing of creating orders into non-global matching pools is deferred until the admin API endpoint for doing this is implemented.

### DIFF
--- a/common/src/types/tasks/descriptors/mod.rs
+++ b/common/src/types/tasks/descriptors/mod.rs
@@ -215,9 +215,6 @@ pub mod mocks {
         NewWalletTaskDescriptor, QueuedTask, QueuedTaskState, TaskDescriptor, TaskQueueKey,
     };
 
-    /// The name of a mock matching pool used for testing task history
-    pub const MOCK_MATCHING_POOL: &str = "mock-matching-pool";
-
     /// Generate the wallet update signature for a new wallet
     pub fn gen_wallet_update_sig(wallet: &Wallet, key: &SecretSigningKey) -> Vec<u8> {
         let new_wallet_comm = wallet.get_wallet_share_commitment();
@@ -271,11 +268,13 @@ pub mod mocks {
 mod test {
     use constants::Scalar;
 
-    use crate::types::wallet_mocks::{mock_empty_wallet, mock_order};
+    use crate::types::{
+        wallet::OrderIdentifier,
+        wallet_mocks::{mock_empty_wallet, mock_order},
+    };
 
     use super::{
-        mocks::{gen_wallet_update_sig, MOCK_MATCHING_POOL},
-        NewWalletTaskDescriptor, UpdateWalletTaskDescriptor,
+        mocks::gen_wallet_update_sig, NewWalletTaskDescriptor, UpdateWalletTaskDescriptor,
     };
 
     /// Tests creating a new wallet task with an invalid secret sharing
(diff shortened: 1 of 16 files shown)


state/src/interface/matching_pools.rs
```diff
@@ -89,27 +89,10 @@ mod test {
 
     use crate::{matching_pools::GLOBAL_MATCHING_POOL, test_helpers::mock_state};
 
-    /// Test creating the global matching pool
-    #[tokio::test]
-    async fn test_create_global_matching_pool() {
```
akirillo, line 94: doesn't make sense to have this test anymore since the global matching pool is created in `mock_state()`

common/src/types/tasks/descriptors/mod.rs
```diff
@@ -215,6 +215,9 @@ pub mod mocks {
         NewWalletTaskDescriptor, QueuedTask, QueuedTaskState, TaskDescriptor, TaskQueueKey,
     };
 
+    /// The name of a mock matching pool used for testing task history
+    pub const MOCK_MATCHING_POOL: &str = "mock-matching-pool";
```
joeykraut, line 219: nit: let's move this into the `mocks` module or otherwise somewhere gated behind the "mocks" feature

common/src/types/tasks/descriptors/update_wallet.rs
```diff
@@ -39,6 +40,10 @@ pub enum WalletUpdateType {
     PlaceOrder {
         /// The order to place
         order: Order,
+        /// The ID of the order
+        id: OrderIdentifier,
+        /// The matching pool to assign the order to
+        matching_pool: String,
```
joeykraut, line 46: nit: maybe define `MatchingPoolId` or similar type alias. It fits in better with the precedent in our codebase and makes clearer interfaces that may otherwise return `String` down the line

workers/api-server/src/http/wallet.rs
```diff
@@ -423,9 +423,11 @@ impl TypedHandler for CreateOrderHandler {
 
         let task = UpdateWalletTaskDescriptor::new_order(
```
joeykraut, line 424: nit: could be nice to have `new_order` default to the global and add a separate `new_order_with_matching_pool`

joeykraut (approved):

common/src/types/tasks/descriptors/mod.rs
akirillo: this is already gated behind the mocks feature, and honestly i'm not sure what a better place for it would be, so i'm partial to keeping it here

PR #697, opened 2024-07-01
workers: handshake-manager: internal_engine: check that orders are in the same matching pool

akirillo:
This PR updates the internal matching engine to check that local orders which are candidates for a match are in the same matching pool. For now, we only add this check the internal matching engine.

I tested this ad-hoc by running a local relayer, putting up 2 crossing orders via the CLI, and ensuring that they matched & settled successfully.

Testing that orders _not_ in the same matching pool do _not_ match is deferred for when the admin API endpoint for creating orders in different matching pools is implemented.

### DIFF
--- a/workers/handshake-manager/src/manager/internal_engine.rs
+++ b/workers/handshake-manager/src/manager/internal_engine.rs
@@ -22,6 +22,9 @@ use super::HandshakeExecutor;
 
 /// Error emitted when proofs of validity cannot be found for an order
 const ERR_MISSING_PROOFS: &str = "validity proofs not found in global state";
+/// Error emitted when an order being matched on is not assigned to any matching
+/// pool
+const ERR_NO_MATCHING_POOL: &str = "order not assigned to matching pool";
 
 // ------------------------
 // | Matching Engine Impl |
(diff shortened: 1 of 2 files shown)


workers/handshake-manager/src/manager/internal_engine.rs
```diff
@@ -62,22 +62,33 @@ impl HandshakeExecutor {
...
                 .await?
                 .ok_or_else(|| HandshakeManagerError::State(ERR_NO_WALLET.to_string()))?;
+
             if !self.wallets_can_match(wallet.wallet_id, other_wallet_id) {
                 continue;
             }
 
-            // Lookup the witness used for this order
+            // Orders must be in the same matching pool
+            let my_pool = self.state.get_matching_pool_for_order(&network_order.id).await?;
```
joeykraut, line 82: we should move this out of the `for` loop

workers/handshake-manager/src/manager/internal_engine.rs
```diff
@@ -62,22 +62,33 @@ impl HandshakeExecutor {
...
+            let other_pool = self.state.get_matching_pool_for_order(&order_id).await?;
```
joeykraut, line 83: why not have `get_matching_pool_for_order` return `Result<String, ...>` instead of `Result<Option<...>>`? We already default to the global matching pool semantically and it just complicates interfaces to express this as an option when there really isn't one.

workers/handshake-manager/src/manager/internal_engine.rs
```diff
@@ -62,22 +62,33 @@ impl HandshakeExecutor {
```
joeykraut, line 62: might make more sense to move this whole check into a pre-filter implemented in `get_locally_matchable_orders`. We could change that method into `get_locally_matchable_orders_in_pool` or something of the like. This scales a bit better as we don't need to open and close a tx for each new order and all the filtering can be done up front. Plus putting the check where it is now, we incur a bunch of other lookups ahead of even getting to the order (e.g. looking up the wallets for both orders).

This is why, for example, the implementation of `get_locally_matchable_orders` does all these checks up front; ensuring there's a wallet, ensuring proofs exist, making sure the queue isn't paused, etc. I feel that's a more natural home for this sort of filtering.

joeykraut (commented):
want to discuss the location of this change before moving forward, open to pushback though

workers/handshake-manager/src/manager/internal_engine.rs
akirillo: hmm, I'd push back on this. we set this as a default behavior at a higher level of abstraction than the storage layer getter method, and I'd prefer to keep it there. I'd like the storage layer to be consistent, i.e. if `get_matching_pool_for_order` says that order `o` is in pool `p`, it's iff there is a table entry mapping `o -> p`.

workers/handshake-manager/src/manager/internal_engine.rs
akirillo: yeah i totally see where you're coming from. my initial approach was to do this in `get_locally_matchable_orders` but then I saw it's used in a number of other places, e.g. the open orders admin endpoint & the handshake manager, and I feel in those places filtering by matching pool does not make sense. but I like the approach of making a separate `get_locally_matchable_orders_in_pool`, I'll go ahead with that.

joeykraut (approved):
thanks for making the changes

PR #698, opened 2024-07-02
workers: api-server: router, admin: add admin routes for matching pool mgmt

akirillo:
This PR adds the following Admin API routes:
- `/v0/admin/:matching_pool/create`: create a matching pool
- `/v0/admin/:matching_pool/destroy`: destroy a matching pool
- `/v0/admin/wallet/:wallet_id/orders/:matching_pool`: create an order in a matching pool

This required a small update to how we handle route authentication. Namely, because the last endpoint requires both wallet and admin auth, the router now accepts a `Vec<AuthType>` so that routes can be configured with multiple authentication methods.

This is currently being tested via a local relayer w/ the CLI


external-api/src/http/admin.rs
```diff
@@ -11,6 +11,12 @@ use serde::{Deserialize, Serialize};
 pub const IS_LEADER_ROUTE: &str = "/v0/admin/is-leader";
 /// Get the open orders managed by the node
 pub const ADMIN_OPEN_ORDERS_ROUTE: &str = "/v0/admin/open-orders";
+/// Route to create a matching pool
+pub const ADMIN_MATCHING_POOL_CREATE_ROUTE: &str = "/v0/admin/:matching_pool/create";
```
joeykraut, line 15: i'd suggest we do `/v0/admin/matching-pools/:name/create`. Think about what this endpoint ends up being when we substitute in a concrete value, maybe we call a pool `order123`: `/v0/admin/order123/create`. It's unclear _what_ object I'm creating. & we shouldn't rely on pools being named with "pool" in them or something, that'd be an odd coupling.

Same for the other endpoints too to be clear.

joeykraut: One nit here too: I'd remove `create` from the path. We typically think about "posting" to a path to be editing or creating the object at that path. This is in line with precedent elsewhere in the relayer's API (e.g. wallet creation) so it'd be good to be consistent. `destroy` is fine though, since it specifies a non-default post operation

external-api/src/http/admin.rs
```diff
@@ -11,6 +11,12 @@ use serde::{Deserialize, Serialize};
...
+/// Route to destroy a matching pool
+pub const ADMIN_MATCHING_POOL_DESTROY_ROUTE: &str = "/v0/admin/:matching_pool/destroy";
+/// Route to create an order in a matching pool
+pub const ADMIN_CREATE_ORDER_ROUTE: &str = "/v0/admin/wallet/:wallet_id/orders/:matching_pool";
```
joeykraut, line 19: nit: maybe just `/v0/admin/wallet/:wallet_id/order-in-pool` and take the `matching-pool` as a parameter of the request body. Again, the thinking here is that we're modifying a scoped object of the wallet, which a matching pool is not, but an order is

workers/api-server/src/http.rs
```diff
@@ -175,6 +185,17 @@ pub(super) fn parse_task_id_from_params(
...
 
+/// A helper to parse out a matching pool name from a URL param
+pub(super) fn parse_matching_pool_from_params(
+    params: &UrlParams,
+) -> Result<String, ApiServerError> {
+    params
+        .get(MATCHING_POOL_URL_PARAM)
+        .ok_or_else(|| bad_request(ERR_MATCHING_POOL_PARSE.to_string()))?
+        .parse()
+        .map_err(|_| bad_request(ERR_MATCHING_POOL_PARSE.to_string()))
```
joeykraut, line 196: nit: you don't need to go `to_string` on this, `bad_request` will call `to_string`

workers/api-server/src/http/wallet.rs
```diff
@@ -79,6 +80,50 @@ async fn append_task_and_await(
...
+    // Lookup the wallet in the global state
+    let old_wallet = find_wallet_for_update(wallet_id, state).await?;
+    let mut new_wallet = old_wallet.clone();
+    let new_order: Order = order.into();
+
+    // Check that the timestamp is not too old, then add to the wallet
+    new_wallet.add_order(id, new_order.clone()).map_err(bad_request)?;
+    new_wallet.reblind_wallet();
+
+    let task = match matching_pool {
```
joeykraut, line 103: feel like it'd be cleaner to just create a more direct constructor that accepts an option; like `new_with_with_maybe_pool`

workers/api-server/src/http/wallet.rs
```diff
@@ -409,30 +454,9 @@ impl TypedHandler for CreateOrderHandler {
...
-            old_wallet,
-            new_wallet,
-            req.statement_sig,
-        )
-        .map_err(bad_request)?;
-
-        // Propose the task and await for it to be enqueued
-        let task_id = append_task_and_await(task.into(), &self.state).await?;
-        Ok(CreateOrderResponse { id, task_id })
+        create_order(req.order, req.statement_sig, wallet_id, &self.state, None).await
```
joeykraut, line 459: nit: add `/* pool_name */` after the `None`; good to avoid literals without documentation

joeykraut (approved):
some things that i think should change but approving optimistically

external-api/src/http/admin.rs
akirillo: ack, i already make this change in https://github.com/renegade-fi/renegade/pull/700 but I'll remove the `/create` path segment and have it be a POST method

PR #699, opened 2024-07-02
state: storage: matching_pools: default global pool at storage layer

akirillo:
This PR lowers the assumption of the default global matching pool into the storage layer. We now no longer need to explicitly create the global pool or assign orders into it. The global pool is not tracked as a separate matching pool in the `POOL_TABLE`, and as such any orders w/o an explicit pool assignment are considered to be in the global pool.

Additionally, we now only remove orders from their matching pool when they are cancelled (this implicitly moves them back to the global pool, but they are not matchable regardless). This is the only situation in which we want to remove orders from their matching pools.

All unit tests pass.

### DIFF
--- a/state/src/storage/tx/matching_pools.rs
+++ b/state/src/storage/tx/matching_pools.rs
@@ -7,6 +7,9 @@ use crate::{storage::error::StorageError, POOL_TABLE};
 
 use super::StateTxn;
 
+/// The name of the global matching pool
+pub const GLOBAL_MATCHING_POOL: &str = "global";
+
 /// The prefix of all matching pool keys
 pub const POOL_KEY_PREFIX: &str = "matching-pool/";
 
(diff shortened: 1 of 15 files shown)


state/src/storage/tx/matching_pools.rs
```diff
@@ -95,11 +82,6 @@ impl<'db> StateTxn<'db, RW> {
 
     /// Destroy a matching pool
     pub fn destroy_matching_pool(&self, pool_name: &str) -> Result<(), StorageError> {
-        // Check that the pool is empty
-        if !self.matching_pool_is_empty(pool_name)? {
-            return Err(StorageError::Other(MATCHING_POOL_NOT_EMPTY_ERR.to_string()));
-        }
-
         let all_pools_key = all_matching_pools_key();
```
joeykraut, line 85: should we remove the pool mapping of any orders left in the pool when it is destroyed? That way, an order is still matchable even if its pool is destroyed.

joeykraut (approved):

state/src/storage/tx/matching_pools.rs
akirillo: Oh yikes yea good catch, totally blanked on this. That way orders actually get "moved" to the global pool

PR #701, opened 2024-07-03
workers: api-server: admin: order metadata endpoint

akirillo:
This PR introduces a convenience endpoint for getting an order's metadata via the admin API. This will be used by the quoters to poll for the successful fill of an order that has been targeted for matching. The information returned here is already available vis the open orders API endpoint, but this makes it simpler to search for a specific order.

This has been tested locally via the CLI.


external-api/src/http/admin.rs
```diff
@@ -35,3 +37,10 @@ pub struct OpenOrdersResponse {
     /// The open orders
     pub orders: Vec<OrderMetadata>,
 }
+
+/// The response to an "order metadata" request
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct OrderMetadataResponse {
```
joeykraut, line 43: nit: `AdminOrderMetadataResponse`; this makes it more clear where the type will be used

joeykraut (approved):

PR #703, opened 2024-07-06
workers: api-server: Fix IDOR vulnerabilities

joeykraut:
### Purpose
This PR fixes a few IDOR vulnerabilities discovered in the process of auditing the api server. These are largely around creating orders or wallets -- objects that are indexed globally and may overwrite the orders/wallets of other users.

### Testing
- Unit tests pass
- Tested the IDOR vulnerabilities after after update

### DIFF
--- a/workers/api-server/src/http/wallet.rs
+++ b/workers/api-server/src/http/wallet.rs
@@ -55,8 +55,13 @@ const DEFAULT_ORDER_HISTORY_LEN: usize = 100;
 /// The name of the query parameter specifying the length of the order history
 /// to return
 const ORDER_HISTORY_LEN_PARAM: &str = "order_history_len";
+
 /// The error message returned when a deposit address is screened out
 const ERR_SCREENED_OUT_ADDRESS: &str = "deposit address was screened as high risk";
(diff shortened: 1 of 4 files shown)


workers/api-server/src/http/wallet.rs
```diff
@@ -453,6 +471,12 @@ impl TypedHandler for CreateOrderHandler {
     ) -> Result<Self::Response, ApiServerError> {
         let wallet_id = parse_wallet_id_from_params(&params)?;
 
+        // Check that the order does not already exist
+        let oid = req.order.id;
+        if self.state.get_order_metadata(&oid).await?.is_some() {
```
akirillo, line 476: nit: should this just be `self.state.contains_order(&oid)`? semantically both of these queries should return `true` in the same cases, but just for readability & consistency w/ the check in the `AdminCreateOrderInMatchingPoolHandler`

akirillo (approved):

workers/api-server/src/http/wallet.rs
joeykraut: Good call

PR #704, opened 2024-07-11
workers: api-server: admin: add admin route for fetching a wallet's balances

akirillo:
This PR adds an admin route for fetching a wallet's balances. This will be used in the quoter dashboard to determine the maximum size of a matching order to construct.

This has been tested via the CLI & local relayer, the balances returned match those included in fetching a wallet using wallet auth


akirillo:
closing on account of this feature not being strictly necessary and in the interest of minimizing the scope of the admin API

PR #705, opened 2024-07-12
state: applicator: wallet_index: update local orders set & network order state when cancelling

akirillo:
This PR smooths out some bugs / inconsistencies in order cancellation. In doing so, we slightly change the semantics of the local order set from "any orders managed by the local cluster" to "any _open_ orders managed by the local cluster," where "open" signifies that they are not cancelled and not completely filled.

Namely, the following changes are made:
1. When an order is cancelled, remove it from the local orders set, and mark the underlying `NetworkOrder` as cancelled
    - This also drops the validity proofs/witnesses from the order, which previously was not happening, resulting in the order still being considered `ready_for_match`
2. When recording an order fill, if the order is completely filled, remove it from the local orders set

All unit tests pass.

### DIFF
--- a/common/src/types/network_order.rs
+++ b/common/src/types/network_order.rs
@@ -119,20 +119,13 @@ impl NetworkOrder {
         self.validity_proofs = Some(validity_proofs)
     }
 
-    /// The following state transition methods are made module private because
-    /// we prefer that access flow through the parent (`OrderBook`) object.
-    /// This object has a reference to the system bus for internal events to
-    /// be published
-
     /// Transitions the state of an order back to the received state, this drops
     /// the existing proof of `VALID COMMITMENTS`
-    #[allow(unused)]
     pub fn transition_received(&mut self) {
         self.state = NetworkOrderState::Received;
     }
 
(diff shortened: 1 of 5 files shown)


workers/task-driver/src/utils/order_states.rs
```diff
@@ -35,6 +35,9 @@ pub async fn record_order_fill(
     metadata.record_partial_fill(match_res.base_amount, price);
     if metadata.data.amount == metadata.total_filled() {
         metadata.state = OrderState::Filled;
+        // If the order was completely filled, remove it from the local orders set
+        // as it is no longer matchable
+        state.remove_local_order(&order_id).await?;
```
joeykraut, line 40: this will only happen on the executor, perhaps we should make this part of a state application. Otherwise, it's not entirely important this happens, so maybe we just delete this one

joeykraut (approved):

workers/task-driver/src/utils/order_states.rs
akirillo: good point, i'll try to do this in the applicator

PR #706, opened 2024-07-13
workers: api-server: websocket: add path fragment for ws api routes

akirillo:
This PR prepends a `/ws` to all of the websocket API routes. Primarily, this is to enable Vercel Edge functions to establish connections with the websocket endpoints. This is currently impossible because outgoing WSS/HTTPS connections from an Edge function replace the port being dialed to 443. In our internal infrastructure, we've configured the load balancer to forward all traffic on port 443 to the relayer's HTTP API port - this includes websocket traffic, unfortunately. By adding the `/ws` path fragment, we can set up a single rule to route traffic on port 443 targeting a path with that fragment to the WS port on the relayer, and route all other traffic to the HTTP port. As such, this change represents a slight coupling with our current infra stack (Vercel, AWS ALB), but given it is so low-overhead and generally sensible I think it's fine.

TODO:
- Make the appropriate changes to our infra config
- Make the appropriate changes to the JS SDK

### DIFF
--- a/workers/api-server/src/websocket.rs
+++ b/workers/api-server/src/websocket.rs
@@ -56,21 +56,21 @@ const ERR_HEADER_PARSE: &str = "error parsing headers";
 // ----------
 
 /// The handshake topic, events include when a handshake beings and ends
-const HANDSHAKE_ROUTE: &str = "/v0/handshake";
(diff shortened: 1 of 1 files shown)


joeykraut (approved):

akirillo:
closing due to us no longer needing websockets in edge functions

PR #707, opened 2024-07-15
config: cli: Allow public-key-only configuration for fee key

joeykraut:
### Purpose
This PR adds a `fee-encryption-key` config option and makes this option mutually exclusive with `fee-decryption-key`. For relayers that run fee decryption and settlement externally, the decryption key need not be placed in the relayer config; only the encryption key.

### Testing
- Ran a relayer with only an encryption key, went through common flows of creating a new wallet, depositing, placing an order, matching, then paying fees. Verified that all flows, including fee payment, worked as expected

### DIFF
--- a/config/src/cli.rs
+++ b/config/src/cli.rs
@@ -1,7 +1,11 @@
 //! The relayer CLI and config definitions
 
 use arbitrum_client::constants::Chain;
-use circuit_types::{elgamal::DecryptionKey, fixed_point::FixedPoint, Amount};
+use circuit_types::{
+    elgamal::{DecryptionKey, EncryptionKey},
+    fixed_point::FixedPoint,
+    Amount,
+};
 use clap::Parser;
 use common::types::{
     exchange::Exchange,
@@ -194,8 +198,14 @@ pub struct Cli {
         value_delimiter = ' ',
     )]
     pub arbitrum_private_keys: Vec<String>,
+    /// The key used to encrypt fee payments
+    /// 
+    /// May be specified _instead of_ `fee_decryption_key` in the case that separate infrastructure is used for 
+    /// fee collection and settlement
+    #[clap(long = "fee-encryption-key", value_parser, conflicts_with = "fee_decryption_key")]
+    pub fee_encryption_key: Option<String>,
     /// The key used to decrypt fee payments
-    #[clap(long = "fee-decryption-key", value_parser)]
+    #[clap(long = "fee-decryption-key", value_parser, conflicts_with = "fee_encryption_key")]
     pub fee_decryption_key: Option<String>,
 
     // -------------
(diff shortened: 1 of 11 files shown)


state/src/storage/tx/node_metadata.rs
```diff
@@ -94,7 +95,7 @@ impl<'db, T: TransactionKind> StateTxn<'db, T> {
     }
 
     /// Get the local relayer's fee decryption key
-    pub fn get_fee_decryption_key(&self) -> Result<DecryptionKey, StorageError> {
+    pub fn get_fee_key(&self) -> Result<RelayerFeeKey, StorageError> {
         self.inner()
             .read(NODE_METADATA_TABLE, &LOCAL_RELAYER_DECRYPTION_KEY.to_string())?
```
akirillo, line 100: nit: can we rename `LOCAL_RELAYER_DECRYPTION_KEY` to `LOCAL_RELAYER_FEE_KEY`?

config/src/cli.rs
```diff
@@ -430,6 +440,47 @@ impl Clone for RelayerConfig {
     }
 }
 
+/// Wraps an encryption key (public or private) to allow for the relayer to be
+/// configured with either of the encryption key or the decryption key
+///
+/// Configuring the relayer with the decryption key allows the relayer to
+/// automate tasks like fee redemption
+#[derive(Copy, Clone, Debug, Serialize, Deserialize)]
+pub enum RelayerFeeKey {
```
akirillo, line 449: nit: we use this type throughout the codebase, would be more discoverable to put it in `common::types` than here in `config::cli`

workers/task-driver/src/tasks/pay_relayer_fee.rs
```diff
@@ -354,7 +364,8 @@ impl PayRelayerFeeTask {
             .clone()
             .ok_or_else(|| PayRelayerFeeTaskError::State(ERR_NO_MERKLE_PROOF.to_string()))?;
 
-        let recipient_decryption_key = self.state.get_fee_decryption_key().await?;
+        let recipient_decryption_key =
+            self.state.get_fee_key().await?.secret_key().expect("decryption key missing");
```
akirillo, line 368: nit: can we:
```suggestion
            self.state.get_fee_key().await?.secret_key().ok_or(PayRelayerFeeTaskError::State(ERR_FEE_KEY_MISSING.to_string()))?;
```

I see no need for an `expect` here

akirillo (approved):

workers/task-driver/src/tasks/pay_relayer_fee.rs
joeykraut: it's validated above

config/src/cli.rs
joeykraut: i thought about this, but don't want `types` to become a dumping ground for anything used in multiple places. so we can move it if it needs a more general location.

state/src/storage/tx/node_metadata.rs
joeykraut: yes, good catch

PR #708, opened 2024-07-16
task-driver: driver: Decrease baseline task backoff

joeykraut:
### Purpose
This PR decreases the task driver's initial backoff time from 2 seconds to 100ms. The initial value was set on Starknet to accommodate long pending tx times. On Arbitrum, however, our failures are much more instantly resolved, e.g. a nonce was too low.


akirillo (approved):
yuge

PR #711, opened 2024-07-22
workers: api-server: admin: open-orders: matching pool filter + fillable amt

akirillo:
This PR tweaks the admin `/open-orders` endpoint to support a `matching_pool` query param to filter open orders by, and also to calculate the fillable amount of each order & include that in the response. This may require using the current price of the order's base asset, which is naturally liable to drift by the time the client receives this API response. This is assumed to be understood by the client.

I tested this by running a local relayer & probing the `/open-orders` API while I moved an order between matching pools, including the global pool. I asserted that the responses were properly filtered and included the expected fillable amount

### DIFF
--- a/external-api/src/http/admin.rs
+++ b/external-api/src/http/admin.rs
@@ -5,7 +5,7 @@
 // ---------------
 
 use circuit_types::Amount;
-use common::types::{wallet::order_metadata::OrderMetadata, MatchingPoolName};
+use common::types::{wallet::order_metadata::OrderMetadata, MatchingPoolName, Price};
 use serde::{Deserialize, Serialize};
 
 use crate::types::ApiOrder;
@@ -51,6 +51,8 @@ pub struct OpenOrder {
     pub order: OrderMetadata,
     /// The fillable amount of the order
     pub fillable: Amount,
+    /// The price used to calculate the fillable amount
+    pub price: Price,
 }
 
 /// The request type to add a new order to a given wallet, within a non-global
--- a/external-api/src/http/admin.rs
+++ b/external-api/src/http/admin.rs
@@ -4,6 +4,7 @@
 // | HTTP Routes |
 // ---------------
 
+use circuit_types::Amount;
 use common::types::{wallet::order_metadata::OrderMetadata, MatchingPoolName};
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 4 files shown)


external-api/src/http/admin.rs
```diff
@@ -37,7 +38,21 @@ pub struct IsLeaderResponse {
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct OpenOrdersResponse {
     /// The open orders
-    pub orders: Vec<OrderMetadata>,
+    pub orders: Vec<OpenOrder>,
+}
+
+/// An open order, containing the order's metadata
+/// as well as the fillable amount of the order
+/// given the underlying wallet's balances and potentially
```
joeykraut, line 46: nit: why potentially? `price` is not optional

workers/api-server/src/http.rs
```diff
@@ -122,6 +123,8 @@ const PEER_ID_URL_PARAM: &str = "peer_id";
 const TASK_ID_URL_PARAM: &str = "task_id";
 /// The :matching_pool param in a URL
 const MATCHING_POOL_URL_PARAM: &str = "matching_pool";
+/// The ?matching_pool param in a query string
+const MATCHING_POOL_QUERY_PARAM: &str = "matching_pool";
```
joeykraut, line 127: nit: just use the constant above it? rename if you think the name isn't general enough

workers/api-server/src/http/admin.rs
```diff
@@ -105,21 +121,82 @@ impl TypedHandler for AdminOpenOrdersHandler {
...
             }
         }
 
         Ok(OpenOrdersResponse { orders })
     }
 }
 
+/// Get the fillable amount of an order using the underlying wallet's balances,
+/// & potentially the price of the base asset
+async fn get_fillable_amount_and_price(
```
joeykraut, line 158: nit: can we move this to a helpers section? it's nice to have all the route handlers next to one another

joeykraut (approved):

external-api/src/http/admin.rs
akirillo: ah poor wording on my behalf, the implication here is that only for buy orders does the price factor into the fillable amount. clearer to just remove "potentially".

PR #713, opened 2024-08-02
external-api: network: Add `local_cluster_id` to network topology resp

joeykraut:
### Purpose
This PR adds the local peer's cluster ID to the network topology response in `GET /v0/network`. This will be used by clients to filter the network response by exactly the cluster responding to the request.

I also cleaned up a bunch of clippy lints that were left over from upgrading the rust toolchain.

### Testing
- Tested the endpoint


akirillo (approved):
ty for cleaning up clippy

PR #716, opened 2024-08-05
node-support: bootloader: Register gas wallet on startup

joeykraut:
### Purpose
This PR updates the bootloader to:
1. Generate the peer id for the local node
2. Register a gas wallet with the funds manager under that peer id on startup

### Testing
- Tested against a locally deployed funds manager

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -153,9 +153,9 @@ checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
 
 [[package]]
 name = "anstream"
-version = "0.6.14"
+version = "0.6.15"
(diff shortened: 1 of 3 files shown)


akirillo (approved):
lgtm, i'm sure you're on top of it but this means we should be injecting the `FUNDS_MANAGER_URL` & `FUNDS_MANAGER_KEY` as container env var/secret accordingly in the infra config

joeykraut:
> lgtm, i'm sure you're on top of it but this means we should be injecting the `FUNDS_MANAGER_URL` & `FUNDS_MANAGER_KEY` as container env var/secret accordingly in the infra config

yep, going to follow up with an infra pr

PR #719, opened 2024-08-13
state: benches: applicator: Add state transition benchmarks

joeykraut:
### Purpose
This PR adds benchmarks for the state applicator for a few important state transitions. Importantly a few of these are on different wallets each iteration to prevent queue buildup affecting the results.

### Testing
- Ran all benchmarks

### DIFF
--- a/state/benches/applicator.rs
+++ b/state/benches/applicator.rs
@@ -1,0 +1,132 @@
+//! Benchmarks the state applicator
+#![allow(missing_docs, clippy::missing_docs_in_private_items)]
+use std::time::{Duration, Instant};
+
+use common::types::{
+    gossip::WrappedPeerId,
+    tasks::{mocks::mock_queued_task, QueuedTaskState},
+    wallet_mocks::mock_empty_wallet,
+};
+use criterion::{criterion_group, criterion_main, Criterion, Throughput};
+use state::{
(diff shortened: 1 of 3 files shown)


state/benches/applicator.rs
```diff
@@ -0,0 +1,132 @@
...
+            let mut total_time = Duration::default();
+            for _ in 0..n_iters {
+                let start = Instant::now();
+                applicator.handle_state_transition(Box::new(transition.clone())).unwrap();
+                total_time += start.elapsed();
+            }
+            total_time
+        });
+    });
+}
```
akirillo, line 47: how useful is this benchmark to us if it tests transitioning an empty wallet to another empty wallet? there's lots of interesting conditional logic e.g. indexing new & cancelled orders which will never be invoked here but is common in practice

akirillo (approved):
lgtm but we may want fuzzier wallet update benchmarks as well / later

state/benches/applicator.rs
joeykraut: yeah agreed we should test those things, though at the moment these benchmarks tell us that serialization of large arrays is the bottleneck so currently it's sufficient for highlighting the bottleneck apart from any conditional logic

PR #731, opened 2024-08-22
circuits, circuit-types: Add keychain nonce and constrain increments

joeykraut:
### Purpose
This PR adds a `nonce` field to the `PublicKeychain` type and constrains its use in the circuits. The nonce is allowed to change exactly when the keychain rotates, and in no other location. 

I added constraints where necessary, but most circuits already captured this constraint in a more general keychain equality check. For those circuits, I made sure to test the fallibility of our nonce tests when these constraints were not present.

### Testing
- For each circuit, I added a unit test for the nonce behavior. Then I commented out constraints to make the test fail, and then added them back to ensure the test passes.
- All other tests pass

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1944,6 +1944,7 @@ dependencies = [
  "inventory",
  "itertools 0.10.5",
  "jf-primitives",
+ "k256",
  "lazy_static",
  "mpc-plonk",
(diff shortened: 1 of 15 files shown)


sehyunc (approved):
Nice

PR #732, opened 2024-08-23
external-api: Make API changes for key rotation

joeykraut:
### Purpose
This PR adds API support for key rotation by making two key changes:
1. Only accept `sk_match` when looking up or creating a wallet. This is necessary as a client may not know their `sk_root` value if it has rotated an unknown number of times. They will find out their `sk_root` value via the `nonce` field in the looked up wallet.
2. Accept a new public root key in each wallet update request. We validate that if the root key is rotated, the nonce must be incremented.

### Testing
- Deferred until a follow up with further API changes

### DIFF
--- a/common/src/types/wallet/keychain.rs
+++ b/common/src/types/wallet/keychain.rs
@@ -1,15 +1,79 @@
 //! Keychain helpers for the wallet
 
+use circuit_types::keychain::{
+    PublicIdentificationKey, PublicKeyChain, PublicSigningKey, SecretIdentificationKey,
+    SecretSigningKey,
+};
 use constants::Scalar;
 use contracts_common::custom_serde::BytesSerializable;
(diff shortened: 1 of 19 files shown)


common/src/types/wallet/keychain.rs
```diff
@@ -1,16 +1,80 @@
...
     core::k256::ecdsa::SigningKey as EthersSigningKey,
     types::{Signature, U256},
     utils::keccak256,
 };
+use serde::{Deserialize, Serialize};
 use util::raw_err_str;
 
 use super::Wallet;
 
+/// Represents the private keys a relayer has access to for a given wallet
```
joeykraut, line 20: moved from `wallet.rs`

sehyunc (approved):
LGTM

PR #734, opened 2024-08-23
api-server: Use HMAC auth for wallet requests

joeykraut:
### Purpose
This PR changes the `api-server` to use symmetric auth for all wallet requests based on the wallet's symmetric key. As well, I went and changed the gossip key types to centralize HMAC logic on the `HmacKey` type.

### Testing
- Unit tests pass
- API testing deferred until CLI changes go in


sehyunc (approved):
lgtm

PR #737, opened 2024-08-26
common: token: Add new tokens and exchange support entries

joeykraut:
### Purpose
This PR adds new tokens as per the new token map as well as their support entries for the exchanges we pull prices from.

### Testing
- Tested these tokens with a price reporter running locally

### DIFF
--- a/workers/price-reporter/Cargo.toml
+++ b/workers/price-reporter/Cargo.toml
@@ -18,7 +18,7 @@ tokio = { workspace = true }
 
 # === Networking === #
 tokio-stream = "0.1"
-tokio-tungstenite = "0.18"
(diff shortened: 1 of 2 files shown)


workers/price-reporter/Cargo.toml
```diff
@@ -18,7 +18,7 @@ tokio = { workspace = true }
 
 # === Networking === #
 tokio-stream = "0.1"
-tokio-tungstenite = "0.18"
+tokio-tungstenite = { version = "0.18", features = ["native-tls"] }
```
joeykraut, line 21: this change makes sure that tls is compiled in, even when using the `price-reporter` only as a standalone library

akirillo (approved):

PR #742, opened 2024-08-29
config: cli: Allow some config entries to come from env vars

joeykraut:
### Purpose
This PR makes certain relayer configs available via environment variables. This allows us to specify them as container environment variables in a more configurable way.

### Testing
- Tested each environment variable


akirillo (approved):
nice, i like this

PR #746, opened 2024-08-31
handshake-manager: price-agreement: Initialize streams from token map

joeykraut:
### Purpose
This PR removes the `DEFAULT_PAIRS` static and instead replaces it with the token mapped tickers for all named tokens we support. This more directly represents the prices we want in an environment and works better across environments with different token lists.

### Testing
- Ran a relayer after this change without panicking on the mainnet token list
- All tests pass

### DIFF
--- a/workers/handshake-manager/src/manager/price_agreement.rs
+++ b/workers/handshake-manager/src/manager/price_agreement.rs
@@ -4,16 +4,12 @@ use std::collections::HashMap;
 
 use common::types::{
     exchange::PriceReporterState,
-    token::{
-        Token, AAVE_TICKER, BNB_TICKER, CRV_TICKER, DYDX_TICKER, LDO_TICKER, LINK_TICKER,
-        MATIC_TICKER, UNI_TICKER, USDC_TICKER, WBTC_TICKER, WETH_TICKER,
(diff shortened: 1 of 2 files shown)


workers/handshake-manager/src/manager/price_agreement.rs
```diff
@@ -86,7 +33,10 @@ lazy_static! {
 pub fn init_price_streams(
     price_reporter_job_queue: PriceReporterQueue,
 ) -> Result<(), HandshakeManagerError> {
-    for (base, quote) in DEFAULT_PAIRS.iter() {
+    let quote = Token::from_ticker(USDC_TICKER);
+
+    for (addr, _) in TOKEN_REMAPS.get().unwrap().iter() {
```
akirillo, line 38: we should probably filter out USDC from this list, here and below (not sure the behavior when fetching USDC-USDC price)

akirillo (approved):
god bless ur soul

workers/handshake-manager/src/manager/price_agreement.rs
joeykraut: good point

PR #758, opened 2024-09-06
metrics-sampler: create metrics-sampler module & port raft metrics 

akirillo:
This PR introduces the `metrics-sampler` module, which defines a framework for sampling various system metrics on fixed intervals using the `system-clock`.

As an example, we port over the raft metrics to `metrics-sampler` in the form of the `RaftMetricsSampler`.

This pattern will subsequently be used to sample product insight metrics.

I tested this with a local relayer & on testnet, where the raft metrics are being recorded as expected.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5702,6 +5702,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "metrics-sampler"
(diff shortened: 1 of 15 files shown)


joeykraut (approved):
very nice

PR #759, opened 2024-09-07
metrics-sampler: add aggregate onboarding metrics samples

akirillo:
This PR utilizes the new `AsyncMetricsSampler` primitive to sample metrics regarding aggregate user onboarding progress states.

This has been tested in testnet by creating a new user, executing the full suite of onboarding tasks, and confirming that the sampled metrics reflected this progress.

### DIFF
--- a/core/src/main.rs
+++ b/core/src/main.rs
@@ -42,7 +42,7 @@ use state::State;
 use system_bus::SystemBus;
 
 use error::CoordinatorError;
-use metrics_sampler::{raft_metrics_sampler::RaftMetricsSampler, sampler::MetricSampler};
+use metrics_sampler::setup_metrics_samplers;
 use system_clock::SystemClock;
 use task_driver::worker::{TaskDriver, TaskDriverConfig};
 use tokio::{select, sync::watch};
@@ -134,7 +134,7 @@ async fn main() -> Result<(), CoordinatorError> {
     .await?;
 
     // Register metrics samplers
-    RaftMetricsSampler::new(global_state.clone()).register(&system_clock).await?;
+    setup_metrics_samplers(global_state.clone(), &system_clock).await?;
 
     if args.debug {
         // Build the TUI
--- a/metrics-sampler/src/onboarding_metrics_sampler.rs
+++ b/metrics-sampler/src/onboarding_metrics_sampler.rs
@@ -1,0 +1,157 @@
+//! Samples onboarding-related metrics at a fixed interval
+
+use std::time::Duration;
+
+use common::types::{
+    tasks::{HistoricalTaskDescription, WalletUpdateType},
+    wallet::{Wallet, WalletIdentifier},
+};
+use state::State;
+
+use crate::sampler::AsyncMetricSampler;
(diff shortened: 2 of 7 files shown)


core/src/main.rs
```diff
@@ -135,6 +139,7 @@ async fn main() -> Result<(), CoordinatorError> {
 
     // Register metrics samplers
     RaftMetricsSampler::new(global_state.clone()).register(&system_clock).await?;
+    OnboardingMetricsSampler::new(global_state.clone()).register(&system_clock).await?;
```
joeykraut, line 142: maybe break these out into a helper to colocate for future? maybe `setup_metrics_samplers`

metrics-sampler/src/onboarding_metrics_sampler.rs
```diff
@@ -0,0 +1,131 @@
...
+
+use common::types::tasks::{HistoricalTaskDescription, WalletUpdateType};
+use state::State;
+
+use crate::sampler::AsyncMetricSampler;
+
+/// The name of the sampler for onboarding metrics
+const ONBOARDING_METRICS_SAMPLER_NAME: &str = "onboarding-metrics-sampler";
+/// The interval at which to sample onboarding metrics
+const ONBOARDING_METRICS_SAMPLE_INTERVAL_MS: u64 = 10_000;
```
joeykraut, line 13: can we increase this to 10m-1h or something higher? doing a full scan of the db every 10s will not scale.

similarly this whole approach will become a performance issue as the number of wallets grow. I'm a fan of getting _something_ out the door, but i do think we should seek to replace this quite quickly.

metrics-sampler/src/onboarding_metrics_sampler.rs
```diff
@@ -0,0 +1,131 @@
...
+        // for all wallets, and bucket into sets from there
+
+        let wallets = self.state.get_all_wallets().await?;
+        let num_wallets = wallets.len();
+
+        let mut num_wallets_with_deposits = 0;
+        let mut num_wallets_with_orders = 0;
+        let mut num_wallets_with_matches = 0;
+
+        for wallet in wallets {
```
joeykraut, line 62: best to break this out into a helper, `process_wallet_onboarding_metrics` or similar

metrics-sampler/src/onboarding_metrics_sampler.rs
```diff
@@ -0,0 +1,131 @@
...
+
+                // We fetch only the last 50 tasks for the wallet. This should be
+                // a sufficient sample to determine the wallet's onboarding state.
+                let tasks = self.state.get_task_history(50, &wallet.wallet_id).await?;
+
+                let mut has_deposited = false;
+                let mut has_placed_order = false;
+                let mut has_matched = false;
+
+                for task in tasks {
```
joeykraut, line 84: similarly, we should break out this into a helper, `process_task_metrics` or similar. not a fan of long methods with lots of nested control flow

joeykraut (approved):

PR #760, opened 2024-09-07
metrics-sampler: churn_metrics_sampler: create churn metrics sampler

akirillo:
This PR creates a metrics sampler for churn metrics. Two notable omissions from this sampler are metrics for the time since order creation and the time since last fill. These are non-trivial to compute using a sample-based approach, we'd need to scan through task history and even then we wouldn't be able to match on the correct cancellation event since order cancellation historical task types have the entire cancelled order struct in them, which likely has a different `amount` than the order metadata.

This is running in testnet & is reporting metrics accurately.

### DIFF
--- a/metrics-sampler/src/churn_metrics_sampler.rs
+++ b/metrics-sampler/src/churn_metrics_sampler.rs
@@ -1,0 +1,183 @@
+//! Samples order-cancellation-related metrics at a fixed interval
+
+use std::time::Duration;
+
+use common::types::{
+    exchange::PriceReporterState, network_order::NetworkOrderState, token::Token,
+    wallet::order_metadata::OrderMetadata, Price,
+};
+use futures::future::join_all;
+use job_types::price_reporter::{PriceReporterJob, PriceReporterQueue};
+use num_bigint::BigUint;
+use state::State;
+use util::hex::biguint_to_hex_addr;
+
+use crate::sampler::AsyncMetricSampler;
+
+/// The name of the sampler for cancellation metrics
+const CANCELLATION_METRICS_SAMPLER_NAME: &str = "cancellation-metrics-sampler";
+/// The interval at which to sample cancellation metrics
+const CANCELLATION_METRICS_SAMPLE_INTERVAL_MS: u64 = 3_600_000; // 1 hour
+
+/// Metric describing the number of orders cancelled
+const NUM_ORDERS_CANCELLED_METRIC: &str = "num_orders_cancelled";
+/// Metric describing the value of a cancelled order
+const CANCELLED_ORDER_VALUE_METRIC: &str = "cancelled_order_value";
+/// Metric describing the % filled of a cancelled order
+const CANCELLED_ORDER_FILL_PERCENT_METRIC: &str = "cancelled_order_fill_percent";
+
+/// Metric tag denoting the order ID
+const ORDER_ID_METRIC_TAG: &str = "order_id";
+/// Metric tag denoting the base ticker
+const BASE_TICKER_METRIC_TAG: &str = "base_ticker";
+/// Metric tag denoting the quote ticker
+const QUOTE_TICKER_METRIC_TAG: &str = "quote_ticker";
+
+/// Samples cancellation metrics at a fixed interval
+#[derive(Clone)]
+pub struct CancellationMetricsSampler {
+    /// A handle to the global state
+    state: State,
+    /// A handle to the price reporter's job queue
+    price_reporter_job_queue: PriceReporterQueue,
+}
+
+impl CancellationMetricsSampler {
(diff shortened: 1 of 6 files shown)


metrics-sampler/src/onboarding_metrics_sampler.rs
```diff
@@ -118,39 +118,35 @@ impl AsyncMetricSampler for OnboardingMetricsSampler {
...
-        let mut num_wallets_with_matches = 0;
-
-        for wallet in wallets {
-            let (has_deposited, has_placed_order, has_matched) =
-                self.process_wallet_onboarding_progress(wallet).await?;
-
-            if has_deposited {
-                num_wallets_with_deposits += 1;
+        // Only sample on the leader to avoid duplicate metrics
+        if self.state.is_leader() {
```
joeykraut, line 122: nit: cleaner to do
```rust
if !self.state.is_leader() {
    return Ok(())
}
```
less indentation and long control flow

state/src/applicator/wallet_index.rs
```diff
@@ -147,6 +147,15 @@ impl StateApplicator {
                     .get_order_metadata(wallet_id, id)?
                     .ok_or(StateApplicatorError::MissingEntry(ERR_NO_METADATA))?;
 
+                // TODO: Here is where we should emit cancellation metrics.
```
joeykraut, line 150: this comment is relevant to a design that is not settled decided on, we should remove it because it informs nothing about the code.

In particular, i disagree that we'd want to fetch prices here. that would make sense for a metrics collector or a service external to the relayer to do. the metrics export interface should be dead simple and almost always one liners.

but aside, let's remove this comment til we decide what we'll do

metrics-sampler/src/churn_metrics_sampler.rs
```diff
@@ -0,0 +1,202 @@
...
+        .into_iter()
+        .filter_map(Result::ok)
+        .map(|o| o.unwrap())
+        .collect();
+
+        Ok(cancelled_orders_meta)
+    }
+
+    /// Get the prices for all unique pairs across the given orders
+    async fn get_order_prices(
```
joeykraut, line 80: if we're sampling this once an hour is this even a useful thing to do? Prices can change so wildly in an hour that this seems very noisy.

metrics-sampler/src/churn_metrics_sampler.rs
```diff
@@ -0,0 +1,202 @@
...
+            let state = recv.await.map_err(|e| e.to_string())?;
+
+            // Only insert the price if it is nominal
+            let entry = match state {
+                PriceReporterState::Nominal(report) => Some(report.price),
+                _ => None,
+            }
+            .map(|p| ((base, quote), p));
+
+            Ok(entry)
```
joeykraut, line 103: call out to a helper here, `self.get_price`

metrics-sampler/src/churn_metrics_sampler.rs
```diff
@@ -0,0 +1,202 @@
...
+/// Samples churn metrics at a fixed interval
+#[derive(Clone)]
+pub struct ChurnMetricsSampler {
+    /// A handle to the global state
+    state: State,
+    /// A handle to the price reporter's job queue
+    price_reporter_job_queue: PriceReporterQueue,
+}
+
+impl ChurnMetricsSampler {
```
joeykraut, line 51: `OrderCancellationMetricsSampler` is a better, more specific name considering this only tracks cancelations

metrics-sampler/src/churn_metrics_sampler.rs
joeykraut, line 80: Actually, I'd just get rid of this method and re-fetch a price for each order. it takes no time to fetch a price (on the order of 2-digit nanoseconds), prefetching and aggregating them for an already expensive operation we do will make little impact on the execution time here. It just makes things harder to read to pass around a hashmap

joeykraut (commented):
overall looks good, approving to unblock a deploy, but think we can restructure just a bit to make it simpler

state/src/applicator/wallet_index.rs
akirillo: whoops, this is a completely outdated comment (before I thought of the worker idea), will remove

metrics-sampler/src/churn_metrics_sampler.rs
akirillo: fair enough, happy to make this change

akirillo: this is a valid point, but without prices at the moment we have no way of discerning the value of cancelled orders, which is very useful even if we keep in mind that the prices are sampled when the metrics are.

PR #772, opened 2024-10-03
task-driver: refresh-wallet: Check for public blinder on-chain

joeykraut:
### Purpose
This PR adds logic to the `refresh-wallet` task which checks whether the wallet's public blinder has been seen on-chain. If it has _not_, the refresh task can assume that the shares held locally by the relayer are up-to-date.

### Testing
- Unit tests pass
- Tested with a locally running relayer for both cases:
    - Tested the case in which a wallet was up to date
    - Tested the case in which a wallet was out of date (updated it on a different relayer)

### DIFF
--- a/common/src/types/wallet/shares.rs
+++ b/common/src/types/wallet/shares.rs
@@ -49,6 +49,17 @@ impl Wallet {
         self.private_shares.blinder
     }
 
+    /// Get the public blinder of the wallet
+    pub fn public_blinder(&self) -> Scalar {
+        self.blinded_public_shares.blinder
+    }
+
+    /// Get the next public blinder of the wallet
+    pub fn next_public_blinder(&self) -> Scalar {
+        let (new_blinder, new_blinder_private) = self.new_blinder_and_private_share();
+        new_blinder - new_blinder_private
+    }
+
     /// Get the last non-blinder wallet share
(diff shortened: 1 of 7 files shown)


workers/task-driver/src/tasks/refresh_wallet.rs
```diff
@@ -173,8 +174,12 @@ impl Task for RefreshWalletTask {
             },
 
             RefreshWalletTaskState::FindingWallet => {
-                self.find_wallet().await?;
-                self.task_state = RefreshWalletTaskState::CreatingValidityProofs;
+                // If the wallet is up to date, skip to creating validity proofs
```
akirillo, line 177:
```suggestion
                // If the wallet is up to date, skip the creation of validity proofs
```

common/src/types/wallet/shares.rs
```diff
@@ -49,6 +49,17 @@ impl Wallet {
         self.private_shares.blinder
     }
 
+    /// Get the public blinder of the wallet
+    pub fn public_blinder(&self) -> Scalar {
```
akirillo, line 53: looks like we don't use this, but still seems like a convenient getter

akirillo (approved):

PR #777, opened 2024-10-08
circuits: valid-match-settle-atomic: Add external party fees

joeykraut:
### Purpose
This PR adds constraints for the external party's settlement; which is effectively just their fees. Specifically:
- Constrain both of the protocol fee and relayer fee to be valid amounts
- Constrain the protocol fee to be computed correctly

We do not validate the relayer fee _value_; instead allowing the relayer to set its fee on a per-trade basis. The actor submitting the tx should validate this fee.

### Todo
- Proof linking
- Testing

### Testing
- Unit tests pass

### DIFF
--- a/circuits/src/zk_circuits/valid_match_settle_atomic.rs
+++ b/circuits/src/zk_circuits/valid_match_settle_atomic.rs
@@ -9,6 +9,7 @@
 
 use crate::{
     zk_gadgets::{
+        arithmetic::NoopGadget,
         comparators::GreaterThanEqGadget,
         fixed_point::FixedPointGadget,
         select::{CondSelectGadget, CondSelectVectorGadget},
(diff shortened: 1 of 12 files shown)


circuits/src/zk_circuits/valid_match_settle_atomic.rs
```diff
@@ -53,13 +54,25 @@ where
...
 
+    /// Constrain the fields that appear in no constraints to prevent them from
+    /// being changed
+    fn constrain_malleable_fields(
+        statement: &ValidMatchSettleAtomicStatementVar<MAX_BALANCES, MAX_ORDERS>,
+        _witness: &ValidMatchSettleAtomicWitnessVar<MAX_BALANCES, MAX_ORDERS>,
+        cs: &mut PlonkCircuit,
+    ) -> Result<(), CircuitError> {
+        // Constrain the relayer fee address
+        NoopGadget::constrain_noop(&statement.relayer_fee_address, cs)
```
akirillo, line 73: don't we also have to constrain the external party's relayer fee take to be non-malleable?

akirillo (approved):
question about the malleability of relayer fee take but o/w lgtm, approving to unblock

circuits/src/zk_circuits/valid_match_settle_atomic.rs
joeykraut: we do not, because it separately appears in a constraint. Specifically, `statement.external_party_fees.relayer_fee` is constrained to be a valid amount in `validate_external_fees`.

PR #778, opened 2024-10-09
circuit: valid-match-settle-atomic: Add circuit tests

joeykraut:
### Purpose
This PR adds a full test suite for all constraints in the `VALID MATCH SETTLE ATOMIC` circuit. For each test case, I first falsified it by removing the target constraint and ensuring the test fails. Then I put back the constraint and made sure it passed.

The settlement tests are sparser given that we just call out to the `VALID MATCH SETTLE` settlement gadget, which is extensively tested.

### Todo
- Proof linking
- Benchmarks

### Testing
- All unit tests pass


circuits/src/zk_circuits/valid_match_settle_atomic.rs
```diff
@@ -348,9 +348,524 @@ where
...
+    }
+
+    /// Get a random address
+    fn random_address() -> Address {
+        let random_u128 = rand::random::<u128>();
+        Address::from_bytes_be(random_u128.to_be_bytes().as_slice())
+    }
+
+    /// Create an external match result from a match result
+    /// Create an external match result from a match result
```
akirillo, line 520: nit: duplicate doc comment

circuits/src/zk_circuits/valid_match_settle_atomic.rs
```diff
@@ -348,9 +348,524 @@ where
...
+        // Quote amount too high
+        let (mut witness, mut statement) = create_witness_statement_sell_side();
+        let balance = &mut witness.internal_party_balance;
+        let order = &mut witness.internal_party_order;
+
+        // Setup the balance and order to be valid
+        let price_int = 2;
+        let price = FixedPoint::from_integer(price_int);
+        balance.amount = max_amount();
+        order.amount = max_amount() + 1;
```
akirillo, line 688: for the confused reader, can we add a quick comment that `order.amount` is constrained to be a valid amount in another circuit, so it being set to `max_amount() + 1` is not what causes constraint insatisfaction here?

circuits/src/zk_circuits/valid_match_settle_atomic.rs
```diff
@@ -348,9 +348,524 @@ where
...
+        }
+
+        assert!(!check_matching_engine_constraints(&witness, &statement));
+    }
+
+    // --------------------
+    // | Settlement Tests |
+    // --------------------
+
+    /// Test the case in which the price is out of the user-defined limits
```
akirillo, line 771: nit: update doc comment to reflect test

akirillo (approved):

PR #783, opened 2024-10-14
api-server: external-match: Define initial endpoint

joeykraut:
### Purpose
This PR defines initial request/response types for the external (atomic) matching engine. The request type specifies an external order which will match against our consenting internal orders. The response type is to be defined.

The endpoint will send a job to the handshake manager for matching. If a match is found, the task driver will drive the proof and bundling of a response before sending the full payload back to the api layer.

### Testing
- Hit the endpoint
- Unit tests pass

### DIFF
--- a/external-api/src/http/external_match.rs
+++ b/external-api/src/http/external_match.rs
@@ -1,0 +1,87 @@
+//! API types for external matches
+//!
+//! External matches are those brokered by the darkpool between an "internal"
+//! party (one with state committed into the protocol), and an external party,
+//! one whose trade obligations are fulfilled directly through erc20 transfers;
+//! and importantly do not commit state into the protocol
+//!
+//! Endpoints here allow permissioned solvers, searchers, etc to "ping the pool"
+//! for consenting liquidity on a given token pair.
+
+use circuit_types::{fixed_point::FixedPoint, order::OrderSide, Amount};
+use common::types::wallet::Order;
+use num_bigint::BigUint;
+use serde::{Deserialize, Serialize};
+
+use crate::{deserialize_biguint_from_hex_string, serialize_biguint_to_hex_addr};
+
+// ---------------
+// | HTTP Routes |
+// ---------------
+
+/// The route for requesting an atomic match
+pub const REQUEST_EXTERNAL_MATCH_ROUTE: &str = "/v0/matching-engine/request-external-match";
+
+// -------------------------------
+// | HTTP Requests and Responses |
+// -------------------------------
+
+/// The request type for requesting an external match
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct ExternalMatchRequest {
+    /// The external order
+    pub external_order: ExternalOrder,
+}
+
+/// The response type for requesting an external match
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct ExternalMatchResponse {
+    // TODO: Define this type
+}
+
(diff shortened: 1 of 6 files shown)


workers/task-driver/src/driver.rs
```diff
@@ -57,7 +57,7 @@ const TASK_DRIVER_THREAD_NAME: &str = "renegade-task-driver";
 /// The number of times to retry a step in a task before propagating the error
 const TASK_DRIVER_N_RETRIES: usize = 5;
 /// The stack size to allocate for task driver threads
-const DRIVER_THREAD_STACK_SIZE: usize = 5_000_000; // 5MB
+const DRIVER_THREAD_STACK_SIZE: usize = 50_000_000; // 5MB
```
joeykraut, line 60: just going to change this now so that we can run in debug mode

external-api/src/http/external_match.rs
```diff
@@ -0,0 +1,87 @@
...
+    /// The amount of the order
+    pub amount: Amount,
+    /// The minimum fill size for the order
+    #[serde(default)]
+    pub min_fill_size: Amount,
+}
+
+impl From<ExternalOrder> for Order {
+    fn from(order: ExternalOrder) -> Self {
+        let worst_case_price = if order.side == OrderSide::Buy {
```
akirillo, line 72: I'm in favor of exposing this as an optional field on the `ExternalOrder` type, I could see users of this endpoint wanting to set it

workers/api-server/src/http.rs
```diff
@@ -390,6 +394,15 @@ impl HttpServer {
             GetOrderHistoryHandler::new(state.clone()),
         );
 
+        // --- External Match Routes --- //
+
+        // The "/external-match/request" route
+        router.add_unauthenticated_route(
```
akirillo, line 400: Certainly out of scope for this PR, but I am curious the auth method we'll use for this endpoint. We probably don't want to use a single symmetric secret like the other endpoints so that we can have better control over per-user access granting / revocation, right?

workers/task-driver/src/driver.rs
```diff
@@ -57,7 +57,7 @@ const TASK_DRIVER_THREAD_NAME: &str = "renegade-task-driver";
 /// The number of times to retry a step in a task before propagating the error
 const TASK_DRIVER_N_RETRIES: usize = 5;
 /// The stack size to allocate for task driver threads
-const DRIVER_THREAD_STACK_SIZE: usize = 5_000_000; // 5MB
+const DRIVER_THREAD_STACK_SIZE: usize = 50_000_000; // 50MB
```
akirillo, line 60: do we want to change this back here, or in a later PR?

akirillo (approved):

workers/task-driver/src/driver.rs
joeykraut: no, see my comment

workers/api-server/src/http.rs
joeykraut: yeah we'll allow relayers to issue api keys

external-api/src/http/external_match.rs
joeykraut: this is more of a "quote" endpoint than a commitment to continuously match the order, so regular user protections don't quite make sense here. the recipient can see the price their order will execute at, and has full discretion to reject on price. we can add it if requested, but i'd rather start lean

PR #784, opened 2024-10-15
handshake-manager: matching: Generalize local matching engine logic

joeykraut:
### Purpose
This PR generalizes the matching engine logic for locally inferred matches. Specifically, we break out much of the core matching logic into a set of helpers in the `HandshakeExecutor` that will allow for easy re-use between the internal engine and external (atomic) engine.

This has the benefit of also simplifying and optimizing the match critical path.

I also removed the match mutual exclusion list as this argument is no longer needed.

### Testing
- All tests pass
- Live testing deferred until the feature branch is ready for merge

### DIFF
--- a/state/src/interface/wallet_index.rs
+++ b/state/src/interface/wallet_index.rs
@@ -3,6 +3,7 @@
 //! Wallet index updates must go through raft consensus so that the leader may
 //! order them
 
+use circuit_types::balance::Balance;
 use common::types::{
     tasks::QueuedTask,
     wallet::{Order, OrderIdentifier, Wallet, WalletIdentifier},
@@ -57,6 +58,29 @@ impl State {
             let wallet_id = res_some!(tx.get_wallet_for_order(&id)?);
             let wallet = res_some!(tx.get_wallet(&wallet_id)?);
             Ok(wallet.orders.get(&id).cloned())
+        })
+        .await
+    }
+
+    /// Get the order for a given order ID and the balance that capitalizes it
+    pub async fn get_managed_order_and_balance(
+        &self,
+        id: &OrderIdentifier,
+    ) -> Result<Option<(Order, Balance)>, StateError> {
+        let id = *id;
+        self.with_read_tx(move |tx| {
+            let wallet_id = res_some!(tx.get_wallet_for_order(&id)?);
+            let wallet = res_some!(tx.get_wallet(&wallet_id)?);
+            let order = res_some!(wallet.orders.get(&id)).clone();
+
+            // Get the balance that capitalizes the order
+            let sell_mint = order.send_mint().clone();
+            let balance = wallet
+                .balances
+                .get(&sell_mint)
+                .cloned()
+                .unwrap_or_else(|| Balance::new_from_mint(sell_mint));
+            Ok(Some((order, balance)))
         })
         .await
     }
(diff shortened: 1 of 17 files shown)


state/src/interface/wallet_index.rs
```diff
@@ -61,6 +62,29 @@ impl State {
...
+            let wallet = res_some!(tx.get_wallet(&wallet_id)?);
+            let order = res_some!(wallet.orders.get(&id)).clone();
+
+            // Get the balance that capitalizes the order
+            let sell_mint = order.send_mint().clone();
+            let balance = wallet
+                .balances
+                .get(&sell_mint)
+                .cloned()
+                .unwrap_or_else(|| Balance::new_from_mint(sell_mint));
```
akirillo, line 82: Can we document in the doc comment that we will return a "phantom" empty balance (that doesn't actually exist in the wallet) if there is no capitalizing balance?

workers/handshake-manager/src/manager/matching/internal_engine.rs
```diff
@@ -39,86 +41,43 @@ impl HandshakeExecutor {
...
             .orders
             .get(&network_order.id)
+            .cloned()
             .ok_or_else(|| HandshakeManagerError::State(ERR_NO_ORDER.to_string()))?;
-        let (my_proof, my_witness) = self
-            .get_validity_proof_and_witness(&network_order.id)
-            .await?
-            .ok_or_else(|| HandshakeManagerError::State(ERR_MISSING_PROOFS.to_string()))?;
+        let sell_mint = my_order.send_mint();
+        let my_balance = wallet.get_balance_or_default(sell_mint);
```
akirillo, line 55: 1. Looks like there's no `wallet.get_balance_or_default` method, so I guess you've added it in another PR, but at any rate
2. This block of code looks logically equivalent to your new `get_managed_order_and_balance` method, can use it here

workers/handshake-manager/src/manager/matching/internal_engine.rs
```diff
@@ -133,47 +92,47 @@ impl HandshakeExecutor {
...
+            match_candidates.remove(&other_order_id);
         }
 
-        info!("No internal matches found for {order:?}");
         Ok(())
     }
 
+    /// Get the set of match candidates for an order
+    ///
+    /// Shuffles the ordering of the other orders for fairness
```
akirillo, line 105: Have been thinking about this - do we want to shuffle? I think it's worth considering ordering either by earliest -> latest creation date, or by size, to ensure we match up the biggest fills

akirillo: Also, there's no explicit shuffle going on here, we just insert the orders into a hash set

workers/handshake-manager/src/manager/matching/internal_engine.rs
```diff
@@ -197,30 +156,12 @@ impl HandshakeExecutor {
...
     }
 
     // -----------
     // | Helpers |
     // -----------
 
-    /// Check if two wallets may match
-    fn wallets_can_match(&self, w1: WalletIdentifier, w2: WalletIdentifier) -> bool {
-        // Same wallet
-        if w1 == w2 {
```
akirillo, line 210: We still need to do this check somewhere, and we no longer do. I think the best place would be in `get_match_candidates`, though it would mean two extra state lookups for wallet IDs there. conversely, it could be done in `try_settle_match`, where we already look up the wallet IDs

akirillo (approved):
have some comments but lgtm - just want to make sure we add back the "different wallets" check but that's a small fix so approving to unblock

workers/handshake-manager/src/manager/matching/internal_engine.rs
joeykraut: yeah the shuffle is unnecessary, we can order them however we like down the line

PR #785, opened 2024-10-15
handshake-manager: matching: external-engine: Add external engine logic

joeykraut:
### Purpose
This PR sets up the initial logic for the external matching engine and connects it via the `HandshakeManager`'s job queue to the api server's endpoint.

### Todo
- External match settlement; i.e. return the bundle to the caller and await on-chain settlement

### Testing
- Unit tests pass
- Tested the endpoint, verified that it correctly called the external engine

### DIFF
--- a/workers/handshake-manager/src/manager/matching/external_engine.rs
+++ b/workers/handshake-manager/src/manager/matching/external_engine.rs
@@ -7,16 +7,120 @@
 //! The external matching engine is responsible for matching an external order
 //! against all known internal order
 
-use common::types::wallet::Order;
+use std::collections::HashSet;
+
+use circuit_types::{balance::Balance, fixed_point::FixedPoint, r#match::MatchResult};
+use common::types::{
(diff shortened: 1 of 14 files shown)


workers/handshake-manager/src/manager/matching/external_engine.rs
```diff
@@ -7,16 +7,121 @@
...
+        // TODO: Implement this method
         todo!()
     }
+
+    /// Mock a balance for an external order
+    ///
+    /// We cannot know the external party's balance here so we mock it for the
+    /// matching engine. We assume the external order is fully capitalized and
+    /// so we mock a full balance
+    fn mock_balance_for_external_order(&self, order: &Order, price: f64) -> Balance {
```
akirillo, line 113: nit: we can just accept a `FixedPoint` price argument here, we already compute it in the caller

akirillo (approved):

PR #787, opened 2024-10-16
task-driver: tasks: settle-match-external: Create atomic settlement bundles

joeykraut:
### Purpose
This PR implements the proof generation and client response steps of the settle external match task. This involves creating the witness and statement for the proof of `VALID MATCH SETTLE ATOMIC`, proving the circuit w/ proof-links, and responding via the system bus to the client.

### Testing
- Unit tests pass
- Hit the API endpoint and got back a settlement bundle

### DIFF
--- a/workers/api-server/src/error.rs
+++ b/workers/api-server/src/error.rs
@@ -56,6 +56,12 @@ impl From<ApiServerError> for Response<Body> {
     }
 }
 
+/// Create an `ApiServerError` with a 204 no content code
+#[allow(clippy::needless_pass_by_value)]
+pub(crate) fn no_content<E: ToString>(e: E) -> ApiServerError {
+    ApiServerError::HttpStatusCode(StatusCode::NO_CONTENT, e.to_string())
+}
+
 /// Create an `ApiServerError` with a 400 bad request code
 #[allow(clippy::needless_pass_by_value)]
 pub(crate) fn bad_request<E: ToString>(e: E) -> ApiServerError {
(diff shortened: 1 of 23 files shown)


workers/api-server/src/error.rs
```diff
@@ -56,6 +56,12 @@ impl From<ApiServerError> for Response<Body> {
     }
 }
 
+/// Create an `ApiServerError` with a 204 no content code
```
akirillo, line 59: TIL about 204

workers/api-server/src/http/external_match.rs
```diff
@@ -26,6 +30,27 @@ use crate::{
...
+/// The message returned when no atomic match is found
+const NO_ATOMIC_MATCH_FOUND: &str = "no atomic match found";
+
+// -----------
+// | Helpers |
+// -----------
+
+/// Await a response from the external matching engine
+///
+/// TODO: Add a timeout
```
akirillo, line 42: do we actually need a timeout here? the external matching engine definitely resolves with either `AtomicMatchFound` or `NoAtomicMatchFound`, so it's not as though we use a timeout to infer an error or absence of a match.

if the matching process needs to fit within some time constraint, API clients can set their own timeouts

akirillo: I guess, I can understand wanting a timeout to obviate issues with the system bus, rather than the matching engine... in which case let's be sure to set it relatively high, as we have no reason to expect any issues here

akirillo (approved):

PR #789, opened 2024-10-16
external-api, workers: api-server: http: admin: add augmented order endpoint

akirillo:
This PR tweaks the existing admin open orders & order metadata endpoints, shifting the responsibility of potentially calculating an order's fillable amount from the former to the latter.

As a result, the open orders endpoint now simply returns the IDs of open orders, and the order metadata endpoint (now the "augmented order" endpoint) returns an "augmented order" (fka `OpenOrder`) that contains the order metadata, managing wallet ID, and optionally the fillable amount at the current price.

This makes the open orders endpoint significantly lighter, and shifts the expectation of relayer admins to "follow up" with augmented order requests for open orders.

This was tested on a local relayer using the CLI

### DIFF
--- a/external-api/src/http/admin.rs
+++ b/external-api/src/http/admin.rs
@@ -6,7 +6,7 @@
 
 use circuit_types::Amount;
 use common::types::{
-    wallet::{order_metadata::OrderMetadata, WalletIdentifier},
+    wallet::{order_metadata::OrderMetadata, OrderIdentifier, WalletIdentifier},
     MatchingPoolName, Price,
 };
 use serde::{Deserialize, Serialize};
@@ -46,23 +46,22 @@ pub struct IsLeaderResponse {
 /// The response to an "open orders" request
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct OpenOrdersResponse {
-    /// The open orders
-    pub orders: Vec<OpenOrder>,
+    /// The open order IDs
+    pub orders: Vec<OrderIdentifier>,
 }
 
-/// An open order, containing the order's metadata
-/// and potentially the fillable amount of the order
-/// given the underlying wallet's balances and
-/// the current price of the base asset
+/// An order's metadata, augmented with the containing
+/// wallet's ID, and optionally the fillable amount
+/// of the order and the price used to calculate it
 #[derive(Clone, Debug, Serialize, Deserialize)]
-pub struct OpenOrder {
+pub struct AdminOrderMetadata {
     /// The order metadata
     pub order: OrderMetadata,
     /// The ID of the wallet containing the order
     pub wallet_id: WalletIdentifier,
     /// The fillable amount of the order, if calculated
     pub fillable: Option<Amount>,
-    /// The price used to calculate the fillable amount, if calculated
+    /// The price used to calculate the fillable amount
     pub price: Option<Price>,
 }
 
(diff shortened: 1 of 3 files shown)


external-api/src/http/admin.rs
```diff
@@ -46,23 +46,22 @@ pub struct IsLeaderResponse {
...
+/// An order's metadata, augmented with the managing
+/// wallet's ID, and optionally the fillable amount
+/// of the order and the price used to calculate it
 #[derive(Clone, Debug, Serialize, Deserialize)]
-pub struct OpenOrder {
+pub struct AdminOrderMetadata {
     /// The order metadata
     pub order: OrderMetadata,
-    /// The ID of the wallet containing the order
+    /// The ID of the wallet managing the order
```
joeykraut, line 60: containing is the correct word here. a relayer "manages" a wallet "contains"

workers/api-server/src/http/admin.rs
```diff
@@ -225,15 +198,37 @@ impl TypedHandler for AdminOrderMetadataHandler {
...
+                &self.state,
+                &self.price_reporter_job_queue,
+            )
+            .await
+            .map(|(fillable, price)| (Some(fillable), Some(price)))?
+        } else {
+            (None, None)
+        };
+
+        let order = AdminOrderMetadata { order: order_metadata, wallet_id, fillable, price };
```
joeykraut, line 230: nit: it's cleaner to just build the order metadata and modify it if `include_fillable` is set; the `else { (None, None) }` looks a bit odd, maybe

```rust
        let mut order = AdminOrderMetadata { order: order_metadata, wallet_id, ..Default::default() };
        let include_fillable = parse_include_fillable_from_query_params(&query_params)?;
        if include_fillable {
            let (fillable, price) = get_fillable_amount_and_price(
                &order_metadata,
                &wallet_id,
                &self.state,
                &self.price_reporter_job_queue,
            )
            .await;
            
            order.fillable = fillable;
            order.price = price;
        }
```

workers/api-server/src/http/admin.rs
```diff
@@ -488,19 +483,18 @@ fn parse_include_fillable_from_query_params(params: &QueryParams) -> Result<bool
...
-    let wallet = state.get_wallet(&wallet_id).await?.ok_or(not_found(ERR_WALLET_NOT_FOUND))?;
+    let mut wallet = state.get_wallet(wallet_id).await?.ok_or(not_found(ERR_WALLET_NOT_FOUND))?;
 
     // Get an up-to-date order & balance.
     // The order stored on the `OrderMetadata` does not have the order amount
     // updated to account for fills, which will lead to an incorrect calculation
     // of the fillable amount.
-    let order = wallet.get_order(&meta.id).ok_or(not_found(ERR_ORDER_NOT_FOUND))?;
-    let balance = wallet.get_balance_for_order(order).ok_or(not_found(ERR_BALANCE_NOT_FOUND))?;
+    let order = wallet.remove_order(&meta.id).ok_or(not_found(ERR_ORDER_NOT_FOUND))?;
```
joeykraut, line 496: `remove_order` doesn't make much sense here, why are we doing this?

joeykraut (approved):
approved but would like these resolved before merging

workers/api-server/src/http/admin.rs
akirillo: just to get rid of an unnecessary clone

joeykraut: Get order is appropriate here, this clone is not going to make a dent in performance and as a reader it's confusing why we're removing the order

PR #792, opened 2024-10-16
task-driver: settle-match-external: Await atomic match settlement

joeykraut:
### Purpose
This PR implements the last step of the `settle-external-match` task; waiting for the atomic settlement to land on-chain. We do so by listening for a nullifier spent event with a timeout. 

If the timeout is exceeded, the task completes immediately under assumption that the bundle was not sent to the contract. If the nullifier is seen within the timeout, we refresh the wallet to pick up the newly modified shares.

### Todo
- Add refresh logic in the `chain-events` worker to pick up any missed settlements

### Testing
- Unit tests pass
- Tested both cases:
    - Had the solver client drop the match (no submission on-chain). I verified that the task timed out and did not refresh the wallet
    - Had the solver client submit the match, I verified that the nullifier was seen by the relayer and the wallet correctly refreshed.

### DIFF
--- a/workers/task-driver/src/tasks/settle_match_external.rs
+++ b/workers/task-driver/src/tasks/settle_match_external.rs
@@ -2,6 +2,7 @@
 
 use std::error::Error;
 use std::fmt::{Display, Formatter, Result as FmtResult};
+use std::time::Duration;
 
 use crate::task_state::StateWrapper;
 use crate::tasks::ERR_AWAITING_PROOF;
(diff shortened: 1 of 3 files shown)


workers/task-driver/src/tasks/settle_match_external.rs
```diff
@@ -348,14 +348,36 @@ impl SettleMatchExternalTask {
...
     }
 
     /// Refresh the wallet after the match settlement is witnessed on-chain
+    ///
+    /// Instead of replicating the logic of the refresh task we simply enqueue a
+    /// refresh task for the wallet.
+    ///
+    /// The wallet should have no other tasks enqueued by virtue of it being
+    /// considered for atomic matches, but if it does they will fail through
+    /// and be discarded
```
akirillo, line 372: this part of the doc comment is confusing to me, why is it the case that no other tasks should be enqueued? iirc there is no task queue clearing ahead of an external match settlement task (or any immediate task)

akirillo (approved):

workers/task-driver/src/tasks/settle_match_external.rs
joeykraut: because an order will not be selected as a match candidate if its task queue is not empty

PR #794, opened 2024-10-17
state: storage: wallet-index: Track nullifier <-> wallet mapping

joeykraut:
### Purpose
This PR adds a `wallet_id` <-> `nullifier` double sided mapping. This mapping is updated when a wallet is written. It will allow us to determine which wallets need refreshing when a nullifier is seen on-chain.

We will allow this mapping to backfill naturally through wallet writes in deployment; its use in the events listener will require no backfill.

### Testing
- Unit tests pass
- Added a test for the new mapping

### DIFF
--- a/state/src/lib.rs
+++ b/state/src/lib.rs
@@ -45,7 +45,7 @@ use uuid::Uuid;
 // -------------
 
 /// The number of tables to open in the database
-const NUM_TABLES: usize = 17;
+const NUM_TABLES: usize = 18;
 
 /// The name of the db table that stores node metadata
 pub(crate) const NODE_METADATA_TABLE: &str = "node-metadata";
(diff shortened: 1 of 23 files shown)


state/src/lib.rs
```diff
@@ -69,6 +69,8 @@ pub(crate) const POOL_TABLE: &str = "matching-pools";
 
 /// The name of the db table that maps order to their encapsulating wallet
 pub(crate) const ORDER_TO_WALLET_TABLE: &str = "order-to-wallet";
+/// The name of the db table that maps nullifiers to wallets
+pub(crate) const NULLIFIER_TO_WALLET_TABLE: &str = "nullifier-to-wallet";
```
akirillo, line 73: We also have entries in the `ORDERS_TABLE` mapping from nullifier -> order. Just probing, what are your thoughts on removing those entries, and having the associated queries (e.g. `get_orders_by_nullifier`) look up nullifier -> wallet -> orders? It's an extra "join" but it reduces the amount of state that we have to keep in sync when nullifiers are spent

akirillo (approved):

state/src/lib.rs
joeykraut: I considered this, but want the nullifying logic to work _even if the wallet has no orders_. In the framing you've suggested, we'd want a "full outer join", which isn't possible in this setup.

akirillo: We could keep your implementation of this table as-is, so the logic for enqueuing a wallet refresh still works when the wallet has no orders. We would basically just change the `nullify_orders` code path.
But now that I think about it, I think we rely on that for nullifying network orders as well, so nevermind...

PR #796, opened 2024-10-17
external-api, workers: api-server: websocket: add admin websocket endpoint for all wallet updates

akirillo:
This PR introduces an admin websocket endpoint for subscribing to wallet updates without filtering by wallet ID. This endpoint doesn't stream out any substantive information about the updates themselves, rather just the wallet / order IDs.

This required adding an admin authentication method for the websocket server, as well as a new system bus topic / message for these opaque updates.

This will be tested downstream of implementing the consumption of this API in the JS SDK

### DIFF
--- a/external-api/src/bus_message.rs
+++ b/external-api/src/bus_message.rs
@@ -15,7 +15,7 @@ use uuid::Uuid;
 
 use crate::{
     http::{external_match::AtomicMatchApiBundle, task::ApiTaskStatus},
-    types::{ApiHistoricalTask, ApiWallet},
+    types::{AdminWalletUpdate, ApiHistoricalTask, ApiWallet},
 };
 
 // ----------------------------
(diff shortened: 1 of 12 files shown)


external-api/src/bus_message.rs
```diff
@@ -159,6 +162,10 @@ pub enum SystemBusMessage {
     },
     /// A message indicating that no atomic match was found for a request
     NoAtomicMatchFound,
+
+    // --- Admin -- //
+    /// An opaque message indicating that a wallet has been updated
```
joeykraut, line 167: nit: opaque doesn't mean anything here

joeykraut (approved):
if you have not already, please make sure to test that these changes have no affect on existing websocket clients (specifically the FE)

akirillo:
> if you have not already, please make sure to test that these changes have no affect on existing websocket clients (specifically the FE)

Tested on testnet, frontend seems to be working just fine (getting task updates, etc.)

PR #800, opened 2024-10-18
task-driver: settle-match-external: Record historical fill and metrics

joeykraut:
### Purpose
This PR updates the `settle-external-match` task to record order fill metrics when an atomic match is witnessed on-chain. Note that if a client fails to submit the atomic match bundle on-chain the fill will be missed. For now, we choose not to solve this problem under the assumption that this is a sufficiently unlikely edge case. We can reevaluate if this makes sense once the atomic match system is deployed.

### Testing
- Unit tests pass
- Submitted an atomic match and printed the fill history after the match

### DIFF
--- a/workers/handshake-manager/src/manager/matching/external_engine.rs
+++ b/workers/handshake-manager/src/manager/matching/external_engine.rs
@@ -14,6 +14,7 @@ use common::types::{
     tasks::SettleExternalMatchTaskDescriptor,
     token::Token,
     wallet::{Order, OrderIdentifier},
+    TimestampedPrice,
 };
 use constants::Scalar;
 use external_api::bus_message::SystemBusMessage;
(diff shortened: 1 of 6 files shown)


workers/handshake-manager/src/manager/matching/external_engine.rs
```diff
@@ -65,7 +66,12 @@ impl HandshakeExecutor {
             // may match the external order as the first party
             match_res.direction = order.side.opposite().match_direction();
             let settle_res = self
-                .try_settle_external_match(other_order_id, price, match_res, response_topic.clone())
```
akirillo, line 68: tiny nit: we can now use `ts_price.as_fixed_point()` when defining `price` above

akirillo (approved):

PR #801, opened 2024-10-22
external-api: auth: Add auth v2 helpers

joeykraut:
### Purpose
This PR adds auth v2 helpers to the `external-api` crate available behind the `auth` feature flag. These headers change the MAC payload to be `[path_bytes, header_bytes, body_bytes]` rather than just the body, and prefix renegade headers with `x-renegade`.

I'll upgrade the relayer to check first for a v2 auth then for a v1 auth in the next pr. We will leave the relayer backwards compatible until clients can upgrade.

### Testing
- Tested both the wallet and admin APIs

### DIFF
--- a/external-api/src/auth/error.rs
+++ b/external-api/src/auth/error.rs
@@ -1,0 +1,26 @@
+//! Error types for authentication helpers
+
+use thiserror::Error;
+
+/// Error type for authentication helpers
+#[derive(Error, Debug)]
+pub enum AuthError {
+    /// Error displayed when the signature is invalid
+    #[error("invalid signature")]
+    InvalidSignature,
+    /// Error displayed when the expiration format is invalid
+    #[error("could not parse signature expiration timestamp")]
+    ExpirationFormatInvalid,
+    /// Error displayed when the HMAC is missing from the request
+    #[error("HMAC is missing from the request")]
+    HmacMissing,
+    /// Error displayed when the HMAC format is invalid
+    #[error("HMAC format invalid")]
+    HmacFormatInvalid,
+    /// Error displayed when the signature expiration header is missing
+    #[error("signature expiration missing from headers")]
+    SignatureExpirationMissing,
+    /// Error displayed when a signature has expired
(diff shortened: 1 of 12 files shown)


workers/api-server/src/auth/helpers.rs
```diff
@@ -47,3 +31,73 @@ pub(crate) fn compute_expiring_hmac(key: &HmacKey, payload: &[u8], expiration: u
     let msg_bytes = [payload, &expiration.to_le_bytes()].concat();
     key.compute_mac(&msg_bytes)
 }
+
```
joeykraut, line 34: added this all back in one place so we can delete it when we migrate

external-api/src/auth/error.rs
```diff
@@ -0,0 +1,26 @@
+//! Error types for authentication helpers
+
+use thiserror::Error;
+
+/// Error type for authentication helpers
+#[derive(Error, Debug)]
+pub enum AuthError {
+    /// Error displayed when the signature is invalid
+    #[error("invalid signature")]
```
akirillo, line 9: this is neat

external-api/src/lib.rs
```diff
@@ -13,18 +13,20 @@ use serde::{
...
 pub mod http;
 pub mod types;
 pub mod websocket;
 
-/// Header name for the HTTP auth signature
-pub const RENEGADE_AUTH_HEADER_NAME: &str = "renegade-auth";
-/// Header name for the expiration timestamp of a signature
-pub const RENEGADE_SIG_EXPIRATION_HEADER_NAME: &str = "renegade-auth-expiration";
+/// Header name for the HTTP auth signature; lower cased
+pub const RENEGADE_AUTH_HEADER_NAME: &str = "x-renegade-auth";
```
akirillo, line 24: will we still be using this header, or will we be using the same header (`RENEGADE_AUTH_HMAC_HEADER_NAME`) for both admin and wallet auth?

akirillo (approved):

PR #802, opened 2024-10-22
external-api: http, bus_message, workers: api-server: admin: replace order placement events w/ wallet order IDs endpoint

akirillo:
This PR adds an admin API endpoint that returns the order IDs for a given wallet. This allows us to simplify the admin wallet updates system bus topic to only be concerned with wallet updates, agnostic of whether they were order placements, cancellations, or otherwise.


external-api/src/http/admin.rs
```diff
@@ -31,6 +31,8 @@ pub const ADMIN_CREATE_ORDER_IN_MATCHING_POOL_ROUTE: &str =
 pub const ADMIN_ASSIGN_ORDER_ROUTE: &str = "/v0/admin/orders/:order_id/assign-pool/:matching_pool";
 /// Route to get the matching pool for an order
 pub const ADMIN_GET_ORDER_MATCHING_POOL_ROUTE: &str = "/v0/admin/orders/:order_id/matching-pool";
+/// Route to get all the order IDs for a given wallet
+pub const ADMIN_WALLET_ORDERS_ROUTE: &str = "/v0/admin/wallet/:wallet_id/orders";
```
joeykraut, line 35: nit: i'd group these up above with the other getter endpoints. Can we also make the last path segment `order-ids` to be explicit about the type we return here? We have other endpoints with path `../orders` that return an order type

joeykraut (approved):

PR #803, opened 2024-10-22
api-server: auth: Add authv2 shims while clients migrate

joeykraut:
### Purpose
This PR migrates the relayer authentication to use v2 authentication. Specifically, I introduce shim code to first try v2 auth then fall back to v1 if v2 is not available. This preserves backwards compatibility while the clients are being updated.

### Testing
- Tested with an old client using v1 auth
- Tested with a new client using v2 auth
- All unit tests pass

### DIFF
--- a/external-api/src/lib.rs
+++ b/external-api/src/lib.rs
@@ -25,9 +25,6 @@ pub const RENEGADE_AUTH_HEADER_NAME: &str = "x-renegade-auth";
 /// Header name for the expiration timestamp of a signature; lower cased
 pub const RENEGADE_SIG_EXPIRATION_HEADER_NAME: &str = "x-renegade-auth-expiration";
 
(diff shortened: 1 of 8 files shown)


external-api/src/lib.rs
```diff
@@ -25,9 +25,6 @@ pub const RENEGADE_AUTH_HEADER_NAME: &str = "x-renegade-auth";
 /// Header name for the expiration timestamp of a signature; lower cased
 pub const RENEGADE_SIG_EXPIRATION_HEADER_NAME: &str = "x-renegade-auth-expiration";
 
-/// Header name for the HTTP auth HMAC
```
joeykraut, line 28: admin now uses the same auth headers and path

akirillo (approved):

PR #806, opened 2024-10-25
workers: api-server: http: admin: return matchable order IDs for wallet

akirillo:
This PR tweaks the `/admin/wallet/:id/order-ids` endpoint to only return the IDs of _matchable_ orders in the wallet. This is preferable to returning all order IDs, including those of default / empty orders.

This has been tested in testnet.

### DIFF
--- a/workers/api-server/src/http/admin.rs
+++ b/workers/api-server/src/http/admin.rs
@@ -17,8 +17,8 @@ use external_api::{
     http::{
         admin::{
             AdminGetOrderMatchingPoolResponse, AdminOrderMetadataResponse,
-            AdminWalletOrderIdsResponse, CreateOrderInMatchingPoolRequest, IsLeaderResponse,
-            OpenOrdersResponse,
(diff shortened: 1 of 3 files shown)


workers/api-server/src/http/admin.rs
```diff
@@ -272,9 +272,10 @@ impl TypedHandler for AdminWalletOrderIdsHandler {
         let wallet_id = parse_wallet_id_from_params(&params)?;
         let wallet =
             self.state.get_wallet(&wallet_id).await?.ok_or(not_found(ERR_WALLET_NOT_FOUND))?;
-        let order_ids = wallet.orders.keys().cloned().collect();
 
-        Ok(AdminWalletOrderIdsResponse { order_ids })
+        let order_ids = wallet.get_matchable_orders().into_iter().map(|o| o.0).collect();
```
joeykraut, line 276: nit: for readability can we change the map closure to `|id, _order| id`? It's unclear just by reading it what types we're working with

joeykraut (approved):

PR #808, opened 2024-10-26
wokers: api-server: admin: decimal-adjust fillable price

akirillo:
This PR fixes the calculation of the fillable amount in the admin order metadata endpoint. We need to decimal-correct the price when computing the fillable amount, but we return the "original" price (which effectively assumes the base/quote tokens use the same number of decimals) to the client, as this is easier to interpret. This basically defers the responsibility of decimal-adjusting the fillable amount to the client if it wishes to compute the value by multiplying it with the price.

### DIFF
--- a/common/src/types.rs
+++ b/common/src/types.rs
@@ -15,6 +15,7 @@ pub mod token;
 pub mod transfer_auth;
 pub mod wallet;
 
+use token::Token;
 use util::get_current_time_millis;
 // Re-export the mock types
 #[cfg(feature = "mocks")]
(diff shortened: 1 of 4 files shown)


common/src/types.rs
```diff
@@ -57,6 +58,26 @@ impl TimestampedPrice {
     pub fn as_fixed_point(&self) -> FixedPoint {
         FixedPoint::from_f64_round_down(self.price)
     }
+
+    /// Decimal correct a price for a given token pair
+    pub fn decimal_correct(
```
joeykraut, line 63: nit: `get_decimal_corrected_price` would be a better name here. The current name seems to imply that we mutate the receiver

workers/api-server/src/http/admin.rs
```diff
@@ -551,19 +551,29 @@ async fn get_fillable_amount_and_price(
...
+            base_token: base_token.clone(),
+            quote_token: quote_token.clone(),
+            channel: price_tx,
+        })
         .map_err(internal_error)?;
 
-    let price = match price_rx.await.map_err(internal_error)? {
-        PriceReporterState::Nominal(report) => report.price,
+    let price: TimestampedPrice = match price_rx.await.map_err(internal_error)? {
+        PriceReporterState::Nominal(report) => (&report).into(),
```
joeykraut, line 562: why this change? An opaque `into` is less readable than field access

joeykraut (approved):

workers/api-server/src/http/admin.rs
akirillo: so that we can call `TimestampedPrice::decimal_correct` on the result

PR #809, opened 2024-10-26
circuits: zk-gadgets: fixed-point: Add floor constraint gadget

joeykraut:
### Purpose
This PR adds a `constrain_equal_floor` method to the `FixedPointGadget` that constrains a given fixed point to equal a given integral value when floored. This will replace the `constrain_equal_integer_ignore_fraction` that is used throughout the matching engine.

### Testing
- Unit tests pass
- Tested all edge cases for the floor gadget

### DIFF
--- a/circuits/src/zk_gadgets/fixed_point.rs
+++ b/circuits/src/zk_gadgets/fixed_point.rs
@@ -10,6 +10,7 @@ use mpc_relation::{errors::CircuitError, traits::Circuit, BoolVar, Variable};
 
 use super::{
     arithmetic::DivRemGadget,
+    bits::{BitRangeGadget, MultiproverBitRangeGadget},
     comparators::{EqGadget, GreaterThanEqZeroGadget, MultiproverGreaterThanEqZeroGadget},
 };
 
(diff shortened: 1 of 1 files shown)


circuits/src/zk_gadgets/fixed_point.rs
```diff
@@ -129,6 +147,22 @@ impl MultiproverFixedPointGadget {
...
+        integer: Variable,
+        fabric: &Fabric,
+        cs: &mut MpcPlonkCircuit,
+    ) -> Result<(), CircuitError> {
+        let zero_var = cs.zero();
+        let one = ScalarField::one();
+
+        let lhs_minus_rhs =
+            cs.lc(&[fp.repr, integer, zero_var, zero_var], &[one, -*TWO_TO_M_SCALAR, one, one])?;
+        MultiproverGreaterThanEqZeroGadget::<{DEFAULT_FP_PRECISION}>::constrain_greater_than_eq_zero(lhs_minus_rhs, fabric, cs)
```
akirillo, line 164: tiny nit: would it be possible to use `MultiproverBitRangeGadget` here? `constrain_greater_than_eq_zero` just calls out to that, but reading this constraint as "the diff must fit in `DEFAULT_FP_PRECISION` bits" makes a lot more sense than "the diff must be >= 0"

akirillo (approved):

circuits/src/zk_gadgets/fixed_point.rs
joeykraut: yeah good call, forgot this gadget existed

PR #820, opened 2024-10-31
state: storage: order-book: Delete orders when cancelled

joeykraut:
### Purpose
This PR deletes orders when they are cancelled. We have the order history now so keeping around cancelled orders merely pollutes the table.

### Todo
- Prune phantom orders from state with migration

### Testing
- Unit tests pass
- Tested the case in which an order is deleted
- [ ] Currently testing deleting an order with a locally running relayer

### DIFF
--- a/state/src/storage/tx/order_book.rs
+++ b/state/src/storage/tx/order_book.rs
@@ -210,7 +210,19 @@ impl<'db> StateTxn<'db, RW> {
 
     /// Delete an order from the order book
     pub fn delete_order(&self, order_id: &OrderIdentifier) -> Result<(), StorageError> {
-        self.inner().delete(ORDERS_TABLE, &order_key(order_id)).map(|_| ())
+        let order_key = order_key(order_id);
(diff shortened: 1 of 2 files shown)


state/src/storage/tx/order_book.rs
```diff
@@ -210,15 +210,28 @@ impl<'db> StateTxn<'db, RW> {
...
     }
 
     /// Nullify the orders indexed by the given nullifier
     pub fn nullify_orders(&self, nullifier: Nullifier) -> Result<(), StorageError> {
         // Delete all orders
         let orders = self.get_orders_by_nullifier(nullifier)?;
         for order_id in orders {
-            self.delete_order(&order_id)?;
+            let order_key = order_key(&order_id);
+            self.inner().delete(ORDERS_TABLE, &order_key).unwrap();
```
sehyunc, line 234: nit: may want to use `?` instead of unwrapping to propagate errors (all other `.delete()` do this)

sehyunc (approved):
small nit regarding unwrapping but otherwise lgtm

PR #821, opened 2024-11-01
task-driver: state-migration: remove-phantom-orders: Define migration

joeykraut:
### Purpose
This PR introduces the notion of a state migration to the relayer's startup task, and defines an initial one to remove phantom orders from the order book. These migrations are run in the last step of the node startup task and can be used to define arbitrary cleanup, backfill, etc logic. 

The migration defined here removes all orders which are not contained within a wallet.

### Testing
- Unit tests pass
- Relayer starts up correctly
- [x] Currently testing against a snapshot containing phantom orders
    - Tested against the mainnet snapshot, order book phantom orders went from 918 -> 0 and local order phantom orders went from 409 -> 0

### DIFF
--- a/workers/task-driver/src/state_migration/remove_phantom_orders.rs
+++ b/workers/task-driver/src/state_migration/remove_phantom_orders.rs
@@ -1,0 +1,121 @@
+//! Removes orders that are no longer part of any wallet
+//!
+//! This checks both the order book table and the local orders list
+
+use common::types::wallet::OrderIdentifier;
+use state::State;
+use tracing::{error, info};
+
+/// Remove all orders that are no longer part of a wallet
+pub async fn remove_phantom_orders(state: &State) -> Result<(), String> {
+    // Remove all phantom orders in the orders table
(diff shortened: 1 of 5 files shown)


workers/task-driver/src/state_migration/remove_phantom_orders.rs
```diff
@@ -0,0 +1,121 @@
...
+    for order_id in orders {
+        let oid = *order_id;
+        let res = state
+            .with_write_tx(move |tx| {
+                tx.delete_order(&oid)?;
+                Ok(())
+            })
+            .await;
+
+        // Do not about on error, just log them
```
sehyunc, line 94: typo: "abort"?

sehyunc (approved):
lgtm other than small typo. out of curiosity, how long did the state migration take in your test and how big was the snapshot?

joeykraut:
> lgtm other than small typo. out of curiosity, how long did the state migration take in your test and how big was the snapshot?

used the mainnet snapshot, it took a handful of seconds on a release build

PR #823, opened 2024-11-02
state: applicator: Update order book cache on state updates

joeykraut:
### Purpose
This PR adds logic in the state transition handlers to keep the order book cache up to date. Specifically:
- When a new order is added to a wallet, we add it to the `externally_enabled` cache if it has `allow_external_matches=true`
- When an order is cancelled, we remove it from the cache entirely
- When an order's validity proof is added to the state, we add it to the `matchable_orders` cache

### Testing
- Added unit tests
- Will test full functionality when I integrate into the matching engine queries

### DIFF
--- a/state/src/applicator/mod.rs
+++ b/state/src/applicator/mod.rs
@@ -8,7 +8,7 @@ use external_api::bus_message::SystemBusMessage;
 use job_types::{handshake_manager::HandshakeManagerQueue, task_driver::TaskDriverQueue};
 use system_bus::SystemBus;
 
-use crate::{storage::db::DB, StateTransition};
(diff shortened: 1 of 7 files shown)


sehyunc (approved):
Looks good, would this cache be populated on startup as wallets are recovered / refreshed?

Nice catch regarding external matches that are awaiting validity proofs.

joeykraut:
> Looks good, would this cache be populated on startup as wallets are recovered / refreshed?
> 
> Nice catch regarding external matches that are awaiting validity proofs.

yes, still need to figure out the specifics, but it will have to be populated on startup

PR #824, opened 2024-11-03
state: interface: order-book: Use order book cache

joeykraut:
### Purpose
This PR updates the order book methods in the `state` to use the order book cache for expensive match queries. Specifically for fetching the matchable and _externally_ matchable orders.

### Todo
- Cache backfill logic from snapshots
- Rigorous testing in testnet

### Testing
- All unit tests pass
- Full testing deferred until feature complete

### DIFF
--- a/state/src/interface/order_book.rs
+++ b/state/src/interface/order_book.rs
@@ -184,17 +184,13 @@ impl StateInner {
     }
 
     /// Get a list of order IDs that are locally managed and ready for match
-    ///
-    /// TODO: Optimize this if necessary, possibly by caching this list
     #[instrument(name = "get_locally_matchable_orders", skip_all)]
     pub async fn get_locally_matchable_orders(&self) -> Result<Vec<OrderIdentifier>, StateError> {
-        self.with_read_tx(|tx| {
-            // Get all the local orders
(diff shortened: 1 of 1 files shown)


state/src/interface/order_book.rs
```diff
@@ -371,50 +357,22 @@ impl StateInner {
 // -----------
 
 impl StateInner {
-    /// Checks whether or not a locally managed order is matchable
-    fn is_order_matchable<T: TransactionKind>(
```
joeykraut, line 375: now that we're caching the bulk of this check, we only need to check whether the task queue is paused.

sehyunc (approved):
Changes LGTM.

 I'm sure you're aware but wanted to remind you of the following TODO for `get_external_match_candidates`: `TODO: Replace this with a correct implementation that filters out orders which do not consent to external matching` which seems related.

joeykraut:
> Changes LGTM.
> 
> I'm sure you're aware but wanted to remind you of the following TODO for `get_external_match_candidates`: `TODO: Replace this with a correct implementation that filters out orders which do not consent to external matching` which seems related.

thanks, removed in the follow-up PR

PR #825, opened 2024-11-03
state: replication: snapshot: Backfill order book cache from snapshot

joeykraut:
### Purpose
This PR adds logic to backfill an orderbook cache from a received snapshot. This will bring cache consistency between nodes when a new node comes online.

### Todo
- Test extensively in testnet

### Testing
- Unit tests pass
- Added a test applying a snapshot to the cache

### DIFF
--- a/state/src/applicator/order_book.rs
+++ b/state/src/applicator/order_book.rs
@@ -87,7 +87,7 @@ impl StateApplicator {
 
         // Update the order metadata into the `Matching` state
         let wallet = tx
-            .get_wallet_for_order(&order_id)?
+            .get_wallet_id_for_order(&order_id)?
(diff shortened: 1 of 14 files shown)


state/src/applicator/order_book.rs
```diff
@@ -87,7 +87,7 @@ impl StateApplicator {
 
         // Update the order metadata into the `Matching` state
         let wallet = tx
-            .get_wallet_for_order(&order_id)?
+            .get_wallet_id_for_order(&order_id)?
```
joeykraut, line 90: i changed this method signature and added a separate `get_wallet_for_order`, so these one-liners appear throughout the PR.

sehyunc (approved):

PR #829, opened 2024-11-06
external-api: external-match: Allow quote denominated external orders

joeykraut:
### Purpose
This PR allows for quote-denominated external orders at the api layer. We do this by sampling a price and converting a quote-denominated order to a base-denominated one before passing the order to the matching engine.

I also refactored some logic in the price reporter's queue to give a cleaner interface.

### Testing
- Unit tests pass
- Tested with both quote denominated and base denominated orders

### DIFF
--- a/external-api/src/http/external_match.rs
+++ b/external-api/src/http/external_match.rs
@@ -9,11 +9,12 @@
 //! for consenting liquidity on a given token pair.
 
 use circuit_types::{
-    fixed_point::FixedPoint, max_price, order::OrderSide, r#match::ExternalMatchResult, Amount,
+    fixed_point::FixedPoint, order::OrderSide, r#match::ExternalMatchResult, Amount,
 };
 use common::types::wallet::Order;
 use ethers::types::transaction::eip2718::TypedTransaction;
 use num_bigint::BigUint;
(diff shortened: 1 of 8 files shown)


external-api/src/http/external_match.rs
```diff
@@ -65,28 +66,48 @@ pub struct ExternalOrder {
...
     /// The side of the market this order is on
     pub side: OrderSide,
-    /// The amount of the order
-    pub amount: Amount,
+    /// The base amount of the order
+    #[serde(default, alias = "amount")]
+    pub base_amount: Amount,
+    /// The quote amount of the order
+    #[serde(default)]
+    pub quote_amount: Amount,
```
sehyunc, line 74: We should explicitly check for one of these fields to exist and return an error if not.

sehyunc: Should also error if both exist, since then it's ambiguous what to set as the `base_amount`.

sehyunc (approved):
LGTM, would be nice to have helpful error messages when creating an external order if none or both amounts are provided

external-api/src/http/external_match.rs
joeykraut: done

PR #830, opened 2024-11-06
arbitrum-client: event-indexing: Use debug traces for unknown selectors

joeykraut:
### Purpose
This PR changes the unknown selector wallet share parsing logic to use the `debug_traceTransaction` RPC method. Previously we were scanning the calldata for a known selector, which is not robust.

The new approach walks the call tree, finding all transactions to the darkpool, then attempts to parse wallet shares from each one. We also check that the parsed wallet shares' public blinder matches the expected blinder, in case multiple darkpool transactions were submitted in one parent frame.

### Testing
- [x] Unit tests
- [x] Test basic lookup functionality
- [x]  Test with a wrapper contract in testnet

### DIFF
--- a/arbitrum-client/src/client/event_indexing.rs
+++ b/arbitrum-client/src/client/event_indexing.rs
@@ -2,6 +2,7 @@
 //! emitted by the darkpool contract
 
 use std::cmp::Reverse;
+use std::collections::VecDeque;
 use std::time::Duration;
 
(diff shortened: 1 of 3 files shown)


akirillo (approved):
lgtm, can i request that when testing w/ a wrapper contract, we also test parsing public shares from a tx where multiple darkpool calls are made?

PR #837, opened 2024-11-18
workers: event-manager: scaffold event types

akirillo:
This PR stubs out the new `event-manager` worker crate, to be used for more robust historical state management, with the event types to be ingested / exported.

TODO: Worker implementation


Cargo.toml
```diff
@@ -31,6 +31,7 @@ members = [
 	"renegade-metrics",
 	"system-clock",
 	"metrics-sampler",
+	"workers/event-manager",
```
joeykraut, line 34: place this with the other "worker" crates

workers/event-manager/src/event_types.rs
```diff
@@ -0,0 +1,139 @@
...
+    /// The event ID
+    pub event_id: Uuid,
+    /// The time at which the event occurred
+    pub event_timestamp: SystemTime,
+    /// The event payload itself
+    pub event: RelayerEventPayload,
+}
+
+/// The payload of a system event generated by the relayer.
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
joeykraut, line 30: I'd externally tag this enum to make it more human-readable

workers/event-manager/src/event_types.rs
```diff
@@ -0,0 +1,139 @@
...
+pub struct OrderCancellationEvent {
+    /// The ID of the wallet that cancelled the order
+    pub wallet_id: WalletIdentifier,
+    /// The ID of the order that was cancelled
+    pub order_id: OrderIdentifier,
+    /// The cancelled order
+    pub order: Order,
+    /// The raw (decimal-unadjusted) remaining amount of the base asset in the
+    /// order
```
joeykraut, line 100: don't know that we need to call out that it's decimal adjusted considering that's the convention everywhere in the relayer except when fetching prices. it's slightly more confusing this way because it supposes that convention is different

workers/event-manager/src/event_types.rs
```diff
@@ -0,0 +1,139 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct MatchEvent {
+    /// The ID of the first wallet in the match
+    pub wallet_id_0: WalletIdentifier,
+    /// The ID of the second wallet in the match
+    pub wallet_id_1: WalletIdentifier,
+    /// The ID of the first order in the match
+    pub order_id_0: OrderIdentifier,
+    /// The ID of the second order in the match
+    pub order_id_1: OrderIdentifier,
```
joeykraut, line 117: naming convention in the relayer would be `wallet_id0`, let's stick with that

joeykraut (approved):
looks good, just some comments on maintaining convention

PR #839, opened 2024-11-20
task-driver: settle-match-external: Set external relayer fee to zero

joeykraut:
### Purpose
This PR updates the external party's relayer fee to be zero for external matches

### Testing
- [ ] Testing in testnet


sehyunc (approved):
lgtm

PR #840, opened 2024-11-22
common: token: get supported exchanges from token mapping

akirillo:
This PR removes the relayer's mapping of supported exchanges for each token ticker in favor of parsing this information from the new token mapping format.

This was tested w/ a couple quick unit tests where we fetch the testnet token mapping, and try instantiating token objects / reading their supported exchanges & tickers, and by running a price reporter that ingested these changes and asserting that all price streams were initialized correctly.

I also tested a number of trades in testnet using this version of the relayer & the associated price reporter version, both buys & sells, high-cap & low-cap tokens. Everything worked smoothly.

### DIFF
--- a/common/src/types/token.rs
+++ b/common/src/types/token.rs
@@ -16,18 +16,26 @@
 //! In general, Named Tokens use all exchanges where they are listed, whereas
 //! Unnamed Tokens only use Uniswap V3 for the price feed.
 use bimap::BiMap;
-use lazy_static::lazy_static;
 use num_bigint::BigUint;
 use serde::{Deserialize, Serialize};
 use std::{
     collections::{HashMap, HashSet},
     fmt::{self, Display},
-    iter,
     sync::OnceLock,
 };
 use util::hex::biguint_to_hex_addr;
 
-use super::exchange::{Exchange, ALL_EXCHANGES};
+use super::exchange::Exchange;
+
+// ---------
+// | Types |
+// ---------
+
+/// A type alias representing the set of supported exchanges for a
+/// given token.
+/// The type is a mapping from exchanges to the ticker used to fetch the
+/// token's price from that exchange
+type ExchangeSupport = HashMap<Exchange, String>;
 
 // ----------------
 // | Quote Tokens |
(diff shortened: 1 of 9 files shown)


common/src/types/token.rs
```diff
@@ -383,51 +55,10 @@ pub static TOKEN_REMAPS: OnceLock<BiMap<String, String>> = OnceLock::new();
...
-                    .unwrap()
-                    .insert(String::from(*erc20_ticker), String::from(okx_ticker));
-            }
-        }
-        exchange_tickers
-    };
-}
+/// The mapping from ERC-20 ticker to the set of exchanges that list the token,
+/// along with the the ticker used to fetch the token's price from the exchange
+pub static EXCHANGE_SUPPORT_MAP: OnceLock<HashMap<String, HashMap<Exchange, String>>> =
```
joeykraut, line 60: can we turn `HashMap<Exchange, String>` here into a more readable type? Like `ExchangeSupport`?

common/src/types/token.rs
```diff
@@ -502,10 +133,12 @@ impl Token {
...
-            .filter(|&exchange| EXCHANGE_TICKERS.get(exchange).unwrap().contains_key(ticker))
+        EXCHANGE_SUPPORT_MAP
+            .get()
+            .unwrap()
+            .get(ticker)
+            .map_or(vec![], |supported_exchanges| supported_exchanges.keys().collect())
+            .into_iter()
             .for_each(|&exchange| {
                 exchanges.insert(exchange);
             });
```
joeykraut, line 144: this is a sort of odd way to mutate the `exchanges` value above. I'd suggest we clean up the method as:
```rust
pub fn supported_exchanges(&self) -> HashSet<Exchange> {
    if !self.is_named() {
        return HashSet::from([Exchange::UniswapV3])
    }

    let ticker = self.get_ticker().unwrap();
    get_exchange_support().get(ticker).map(|exchanges| exchanges.keys().collect()).unwrap_or_default()
}
```

Note also that I added a helper `get_exchange_support()`; probably a useful thing to export that unwraps the `OnceLock`

config/src/token_remaps.rs
```diff
@@ -84,11 +102,31 @@ pub fn setup_token_remaps(remap_file: Option<String>, chain: Chain) -> Result<()
...
+        [(Exchange::Coinbase, usd_ticker.clone()), (Exchange::Kraken, usd_ticker.clone())]
+            .into_iter()
+            .collect(),
+    );
+
+    match EXCHANGE_SUPPORT_MAP.get() {
+        Some(_) => {
+            warn!("Exchange support map already set, cannot override");
+            Ok(())
+        },
```
joeykraut, line 126: this should probably just panic like the others do, a warning would be too easy to look over, so no need to match here.

workers/price-reporter/src/exchange/binance.rs
```diff
@@ -65,15 +65,34 @@ pub struct BinanceConnection {
...
+        base_token: &Token,
+        quote_token: &Token,
+    ) -> Result<Url, ExchangeConnectionError> {
+        let base_ticker = base_token.get_exchange_ticker(Exchange::Binance).ok_or(
+            ExchangeConnectionError::UnsupportedPair(
+                base_token.clone(),
+                quote_token.clone(),
+                Exchange::Binance,
+            ),
+        )?;
```
joeykraut, line 78: we do this exact logic in a few places, probably best to just write a helper for it

joeykraut (approved):

config/src/token_remaps.rs
akirillo: Mm, we also handle double-setting of the `TOKEN_REMAP` and `ADDR_DECIMALS_MAP` in this way, but I think panicking is appropriate in this case. So I'll update those to panic if the `OnceLock` is already set as well

akirillo: for posterity, calling `set` on the `OnceLock` only fails if it is already set, so I remove all the matches and just make a call to `set`, propagating the error instead of panicking

PR #843, opened 2024-11-23
api-server: external-match: Sign external match quotes

joeykraut:
### Purpose
This PR signs the external quote handed back to the external party. This allows the external party to submit the quote to be assembled with the signature as proof that the quote is valid. This allows the relayer to keep the quote's price firm between quote and assembly.

### Testing
- [x] Unit tests pass
- [x] Tested locally

### DIFF
--- a/external-api/src/http/external_match.rs
+++ b/external-api/src/http/external_match.rs
@@ -17,7 +17,10 @@ use ethers::types::transaction::eip2718::TypedTransaction;
 use num_bigint::BigUint;
 use renegade_crypto::fields::scalar_to_u128;
 use serde::{Deserialize, Serialize};
-use util::hex::{biguint_from_hex_string, biguint_to_hex_string};
+use util::{
+    get_current_time_millis,
+    hex::{biguint_from_hex_string, biguint_to_hex_string},
+};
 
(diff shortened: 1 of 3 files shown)


external-api/src/http/external_match.rs
```diff
@@ -177,3 +180,55 @@ impl From<ExternalMatchResult> for ApiExternalMatchResult {
...
+    /// The amount of the quote token exchanged by the match
+    pub quote_amount: Amount,
+    /// The amount of the base token exchanged by the match
+    pub base_amount: Amount,
+    /// The direction of the match
+    pub direction: OrderSide,
+    /// The price of the match
+    pub price: f64,
+    /// The timestamp of the quote
+    pub timestamp: u64,
```
akirillo, line 209: not sure what the intended consumption of this field is, but just flagging that we could make it the timestamp from the timestamped price used by the matching engine, if we pass that back up

akirillo (approved):

external-api/src/http/external_match.rs
joeykraut: to make sure the quote is not stale, I'd prefer to set it more directly as the time at which the quote was issued.

PR #845, opened 2024-11-24
external-api: admin: add token mapping refresh endpoint

akirillo:
This PR adds a new admin endpoint that refreshes the token mapping from the remote source (Github repo). Implementing this required changing the way in which we store the static token mapping objects (address -> ticker mapping, address -> decimals mapping, ticker -> supported exchanges mapping). Namely, instead of using a `OnceLock` that is initialized at startup, we wrap a `RwLock` with a `LazyLock` so we can have a mutable statics that can be overwritten by the refresh endpoint.
Mutable static structures may raise eyebrows, but this approach felt simpler than lifting the mappings into heap-allocated objects that we pass around references to, particularly when one considers that consumers of the relayer library crates (i.e., the price reporter) also rely on these statics being set.

This was tested by running a local relayer w/ a test endpoint that returns the ticker, decimals, and supported exchanges for a token. I used a testing branch for the token mapping repo where I removed the ETHFI token, confirming that the endpoint could not return any info for it. I then re-added the token to the mapping, hit the refresh endpoint, and confirmed that the testing endpoint returned all the correct info for the token.

### DIFF
--- a/common/src/types/token.rs
+++ b/common/src/types/token.rs
@@ -21,7 +21,7 @@ use serde::{Deserialize, Serialize};
 use std::{
     collections::{HashMap, HashSet},
     fmt::{self, Display},
-    sync::OnceLock,
+    sync::{LazyLock, RwLock, RwLockReadGuard, RwLockWriteGuard},
 };
 use util::hex::biguint_to_hex_addr;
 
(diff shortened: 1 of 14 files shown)


common/src/types/token.rs
```diff
@@ -180,10 +188,34 @@ impl Token {
 // | HELPERS |
 // -----------
 
-/// Returns a reference to the exchange support map,
-/// unwrapping the `OnceLock`.
-pub fn get_exchange_support<'a>() -> &'a HashMap<String, ExchangeSupport> {
-    EXCHANGE_SUPPORT_MAP.get().unwrap()
+/// Returns a read lock quard to the token remap
+pub fn read_token_remap<'a>() -> RwLockReadGuard<'a, BiMap<String, String>> {
+    TOKEN_REMAPS.read().unwrap()
```
joeykraut, line 193: can we change these all to `expect`s to get a debug message?

common/src/types/token.rs
```diff
@@ -180,10 +188,34 @@ impl Token {
...
+}
+
+/// Returns a read lock quard to the decimal map
+pub fn read_addr_decimals_map<'a>() -> RwLockReadGuard<'a, HashMap<String, u8>> {
```
joeykraut, line 197: nit: `read_token_decimals_map` is a better name in the context of a caller

joeykraut (approved):

PR #848, opened 2024-11-25
api-server: external-match: Add endpoint to assemble quotes to bundles

joeykraut:
### Purpose
This PR adds an endpoint for external match clients to assemble quotes that they receive from the relayer.

### Testing
- [x] Unit tests pass
- [x] Tested locally

### DIFF
--- a/workers/api-server/src/http/external_match.rs
+++ b/workers/api-server/src/http/external_match.rs
@@ -12,13 +12,14 @@ use std::{sync::Arc, time::Duration};
 
 use arbitrum_client::client::ArbitrumClient;
 use async_trait::async_trait;
-use circuit_types::r#match::ExternalMatchResult;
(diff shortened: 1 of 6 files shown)


workers/api-server/src/http/external_match.rs
```diff
@@ -220,22 +231,59 @@ impl ExternalMatchProcessor {
...
+                order.clone(),
+                false, // quote_only
+                Some(price),
+            )
+            .await?;
+
+        match resp {
+            SystemBusMessage::NoAtomicMatchFound => Err(no_content(NO_ATOMIC_MATCH_FOUND)),
+            SystemBusMessage::AtomicMatchFound { match_bundle, validity_proofs } => {
+                self.build_api_bundle(false, &order, match_bundle, validity_proofs).await
```
akirillo, line 267: tiny nit, can we document the `false` here?

akirillo (approved):

PR #849, opened 2024-11-26
api-server: external-match: Allow `min_fill` in quote units

joeykraut:
### Purpose
This PR allows the `min_fill_size` to be specified in quote units in the external matching engine. Specifically, if the size of the order is specified by `quote_amount`, we expect `min_fill_size` in units of the quote asset. Whereas if the order size is specified by `base_amount`, we expect it in units of base.

### Testing
- [x] Unit tests pass
- [x] Tested all cases locally

### DIFF
--- a/external-api/src/http/external_match.rs
+++ b/external-api/src/http/external_match.rs
@@ -16,7 +16,7 @@ use circuit_types::{
     Amount,
 };
 use common::types::{proof_bundles::AtomicMatchSettleBundle, wallet::Order, TimestampedPrice};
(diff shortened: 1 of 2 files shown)


sehyunc (approved):
LGTM. Will make the corresponding changes to the docs, but IIUC this doesn't change any API request shapes so nothing to be done there.

joeykraut:
> LGTM. Will make the corresponding changes to the docs, but IIUC this doesn't change any API request shapes so nothing to be done there.

that's correct, thanks for taking care of the docs

PR #851, opened 2024-12-02
api-server: Add `/v0/supported-tokens` route

joeykraut:
### Purpose
This PR adds a `/v0/supported-tokens` endpoint to the relayer, returning a list of tokens by address and ticker.

### Testing
- [x] Unit tests pass
- [x] Tested locally

### DIFF
--- a/external-api/src/types/api_order_book.rs
+++ b/external-api/src/types/api_order_book.rs
@@ -44,3 +44,19 @@ impl From<NetworkOrder> for ApiNetworkOrder {
         }
     }
 }
+
+/// A token in the the supported token list
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct ApiToken {
+    /// The token address
+    pub address: String,
+    /// The token symbol
+    pub symbol: String,
+}
+
+impl ApiToken {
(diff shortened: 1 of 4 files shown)


external-api/src/types/api_order_book.rs
```diff
@@ -44,3 +44,19 @@ impl From<NetworkOrder> for ApiNetworkOrder {
...
     }
 }
+
+/// A token in the the supported token list
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct ApiToken {
+    /// The token address
+    pub address: String,
+    /// The token symbol
+    pub symbol: String,
```
sehyunc, line 54: Should this be `ticker` for consistency with the token map?

sehyunc (approved):
LGTM, I'll defer the decision of `symbol` vs. `ticker` to you

external-api/src/types/api_order_book.rs
joeykraut: symbol is better on the api layer IMO b/c it is more common in other interfaces (e.g. the ERC20 abi)

PR #856, opened 2024-12-10
workers: task-driver: emit events from tasks

akirillo:
This PR implements the construction & emitting of events from across the relayer to the new event manager worker.

One thing to note: we emit events right before tasks complete, to obviate any possibility of task retries causing duplicate event emission. I chose this approach for simplicity, and because empirically we have very few failure beyond a task's commit point. With that said, of course we can consider some deduplication scheme / deterministic event IDs, and then emit events right after a task's commit point.

All unit tests pass, and e2e testing is deferred until the receiving side of these events is implemented in the event manager worker

### DIFF
--- a/workers/task-driver/src/tasks/create_new_wallet.rs
+++ b/workers/task-driver/src/tasks/create_new_wallet.rs
@@ -10,6 +10,7 @@
 use core::panic;
 use std::error::Error;
 use std::fmt::{Display, Formatter, Result as FmtResult};
+use std::time::SystemTime;
 
 use arbitrum_client::client::ArbitrumClient;
 use async_trait::async_trait;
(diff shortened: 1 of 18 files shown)


workers/task-driver/src/tasks/create_new_wallet.rs
```diff
@@ -301,4 +310,16 @@ impl NewWalletTask {
             },
         )
     }
+
+    /// Emit a wallet creation event to the event manager
+    fn emit_event(&self) -> Result<(), NewWalletTaskError> {
+        let event = RelayerEvent::WalletCreation(WalletCreationEvent {
```
joeykraut, line 316: we can probably provide constructors that fill in obvious fields like event id and event timestamp and only require those fields not in the scope of the constructor

workers/task-driver/src/tasks/settle_match_external.rs
```diff
@@ -494,4 +508,34 @@ impl SettleMatchExternalTask {
...
+            FixedPoint::default(),
+            internal_party_order_side.opposite(),
+            &self.match_res,
+        );
+
+        let external_match_result = self.match_res.clone().into();
+
+        let event = RelayerEvent::ExternalMatch(ExternalMatchEvent {
+            event_id: Uuid::new_v4(),
+            event_timestamp: SystemTime::now(),
```
joeykraut, line 530: same here and for all other events, i think a constructor would be convenient

workers/task-driver/src/tasks/update_wallet.rs
```diff
@@ -429,4 +453,120 @@ impl UpdateWalletTask {
...
+            transfer,
+        }))
+    }
+
+    /// Construct either an order placement or an order update event,
+    /// depending on whether the order already exists in the new wallet
+    fn construct_order_placement_or_update_event(
+        &self,
+        order: &Order,
+        order_id: &OrderIdentifier,
```
joeykraut, line 499: nit: `order_id` should come before order as it more clearly identifies the parameters

joeykraut (approved):

PR #858, opened 2024-12-12
node-support: event-export-sidecar: create event export sidecar

akirillo:
This PR adds a new `node-support` daemon responsible for re-exporting events from the relayer to the historical state engine. It establishes the Unix socket the relayer exports events to, listens for events on this socket, and submits them to the HSE.

As a part of these changes, I swapped the `Multiaddr` `event_export_addr` config option for a `Url`-based one. This was due to parsing issues with Unix multiaddrs - it interprets `/` as a protocol delimiter so only sibling paths could be used for a socket. URLs capture the same flexibility that we need between HTTP and Unix addresses with better parsing.

**Testing:**
I tested this with a local relayer / sidecar / HSE / DB with the following flow: create wallet, deposit, place order, match, withdraw (and on one of the wallets, cancel order).

### DIFF
--- a/node-support/bootloader/src/main.rs
+++ b/node-support/bootloader/src/main.rs
@@ -51,6 +51,10 @@ const ENV_ADMIN_KEY: &str = "ADMIN_API_KEY";
 const ENV_FUNDS_MANAGER_URL: &str = "FUNDS_MANAGER_URL";
 /// The funds manager api key
 const ENV_FUNDS_MANAGER_KEY: &str = "FUNDS_MANAGER_KEY";
+/// The historical state engine URL
+const ENV_HISTORICAL_STATE_URL: &str = "HISTORICAL_STATE_URL";
+/// The historical state engine auth key
+const ENV_HISTORICAL_STATE_KEY: &str = "HISTORICAL_STATE_KEY";
 
 // --- Constants --- //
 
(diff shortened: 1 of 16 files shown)


node-support/bootloader/src/main.rs
```diff
@@ -51,6 +51,10 @@ const ENV_ADMIN_KEY: &str = "ADMIN_API_KEY";
 const ENV_FUNDS_MANAGER_URL: &str = "FUNDS_MANAGER_URL";
 /// The funds manager api key
 const ENV_FUNDS_MANAGER_KEY: &str = "FUNDS_MANAGER_KEY";
+/// The historical state engine URL
+const ENV_HSE_URL: &str = "HSE_URL";
+/// The historical state engine auth key
+const ENV_HSE_KEY: &str = "HSE_KEY";
```
joeykraut, line 57: I'd expand the `HSE -> HISTORICAL_STATE`, three letter acronyms are hard to read

node-support/event-export-sidecar/src/hse_client.rs
```diff
@@ -0,0 +1,99 @@
...
+
+/// The path to submit events to
+const EVENT_SUBMISSION_PATH: &str = "/event";
+
+// ----------
+// | Client |
+// ----------
+
+/// A client for the historical state engine
+pub struct HseClient {
```
joeykraut, line 27: same here, `Hse` is not a good name, `HistoricalStateClient` is much better

joeykraut (approved):

PR #864, opened 2024-12-17
node-support: event-export-sidecar: SQS event export

akirillo:
This PR equips the event export sidecar in the `node-support` crate with the ability to export events to an AWS SQS queue. Additionally, we remove an unnecessary serialization roundtrip on exported events via the sidecar.

Both the SQS and direct-to-HSE export methods were tested with a local relayer + sidecar, and export events as expected.


node-support/bootloader/src/main.rs
```diff
@@ -115,14 +117,28 @@ async fn main() -> Result<(), String> {
...
+        let hse_key = read_env_var::<String>(ENV_HISTORICAL_STATE_KEY)?;
+
+        Command::new(EVENT_EXPORT_SIDECAR_BIN)
+            .args(["--config-path", CONFIG_PATH])
+            .arg("hse")
+            .args(["--hse-url", &hse_url])
+            .args(["--hse-key", &hse_key])
+            .spawn()
+            .expect("Failed to start event export sidecar process")
+    };
```
joeykraut, line 141: do we need to support both here? feels unnecessary

joeykraut (approved):
i feel that we should just get rid of the http event submission, but o/w looks good

PR #871, opened 2024-12-21
state: caching: order-metadata-index: Build order metadata index

joeykraut:
### Purpose
This PR builds an `OrderMetadataIndex` that will operate as part of the order book cache. The metadata index allows us to query open orders by pair and side, sorted by matchable amount. This will allow for more efficient access in the matching engine and prevent small fills when a larger fill is possible.

### Testing
- [x] Unit tests pass
- Will test in a follow-up when the cache is integrated with the state and matching engine

### DIFF
--- a/state/src/caching/order_metadata_index.rs
+++ b/state/src/caching/order_metadata_index.rs
@@ -1,0 +1,262 @@
+//! Indexing of orders by metadata
+//!
+//! Specifically, we build an efficiently queryable mapping from pair -> side ->
+//! (matchable_amount, order_id)
+//!
+//! This allows the matching engine to efficiently query a narrow set of
+//! candidate orders to match against a target order.
+
+use std::collections::HashMap;
+
+use circuit_types::{order::OrderSide, Amount};
+use common::types::wallet::{Order, OrderIdentifier, Pair};
+use tokio::sync::RwLock;
+
+use super::RwLockHashMap;
+
+/// A type alias for a the inner index mapping
+type OrderSideIndex = HashMap<OrderSide, SortedVec<(Amount, OrderIdentifier)>>;
+
+/// The order metadata index
+#[derive(Default)]
+pub struct OrderMetadataIndex {
+    /// The mapping from pair -> side -> (matchable_amount, order_id)
+    index: RwLockHashMap<Pair, OrderSideIndex>,
+}
+
+impl OrderMetadataIndex {
+    /// Construct a new order metadata index
+    pub fn new() -> Self {
+        let index = RwLock::new(HashMap::new());
+        Self { index }
+    }
+
+    // --- Getters --- //
+
+    /// Get all orders for a given pair and side, sorted by matchable amount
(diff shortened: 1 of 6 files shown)


state/src/caching/order_metadata_index.rs
```diff
@@ -0,0 +1,263 @@
...
+impl<T: Ord> SortedVec<T> {
+    /// Construct a new sorted vector
+    pub fn new() -> Self {
+        Self { vec: Vec::new() }
+    }
+
+    // --- Getters --- //
+
+    /// Get the vector
+    pub fn vec(&self) -> &Vec<T> {
```
akirillo, line 79: it seems we only use this in testing code but clippy is complaining about dead code, can we configure clippy to consider testing code not to be dead code?

state/src/caching/order_metadata_index.rs
```diff
@@ -0,0 +1,263 @@
...
+
+    /// Add an order to the index
+    pub async fn add_order(
+        &self,
+        order_id: OrderIdentifier,
+        order: &Order,
+        matchable_amount: Amount,
+    ) {
+        let pair = order.pair();
+        let side = order.side;
```
akirillo, line 56: nit: could just do 
```suggestion
        let (pair, side) = order.pair_and_side();
```

state/src/caching/order_metadata_index.rs
```diff
@@ -0,0 +1,263 @@
...
+
+        let mut index = self.index.write().await;
+        let entry = index.entry(pair).or_insert_with(OrderSideIndex::new);
+        entry.entry(side).or_insert_with(SortedVec::new).insert((matchable_amount, order_id));
```
akirillo, line 60: what exactly is the default implementation of `Ord` for tuples? is it tie-breaking "left-to-right", i.e. compare `a.0` & `b.0`, tie-break on `a.1` & `b.1`, etc?

state/src/caching/order_metadata_index.rs
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    // --- Setters --- //
+
+    /// Add an order to the index
+    pub async fn add_order(
+        &self,
+        order_id: OrderIdentifier,
+        order: &Order,
+        matchable_amount: Amount,
```
akirillo, line 53: what's the plan for how we compute this? for buy orders, this value depends on the price. will we fetch new prices and update the index on some interval?

akirillo (approved):

state/src/caching/order_metadata_index.rs
akirillo: can ignore, the tests effectively confirm this

joeykraut: yes exactly

state/src/caching/order_metadata_index.rs
joeykraut: it won't be converted via price. so if the order is a buy order, we will just make this the amount of the quote token held by the wallet.

PR #872, opened 2024-12-21
state: caching: order-metadata-cache: Support order amount updates and deletion

joeykraut:
### Purpose
This PR adds support for updating the matchable amount of an order and deleting an order from the cache.

### Testing
- [x] Unit tests pass
- Will test end-to-end when integrated into the matching engine

### DIFF
--- a/state/src/caching/order_metadata_index.rs
+++ b/state/src/caching/order_metadata_index.rs
@@ -85,11 +85,36 @@ impl OrderMetadataIndex {
         let sorted_vec = entry.entry(side).or_insert_with(SortedVec::new);
 
         // Remove the old entry
-        let idx = sorted_vec.find_index(|(_, oid)| *oid == order_id).unwrap();
-        sorted_vec.vec.remove(idx);
+        if let Some(idx) = sorted_vec.find_index(|(_, oid)| *oid == order_id) {
+            sorted_vec.remove(idx);
+        }
 
         // Insert the new entry (this will maintain the sort order)
         sorted_vec.insert((matchable_amount, order_id));
+    }
+
+    /// Remove an order from the index
+    ///
+    /// Note that we do not clean up sub-index entries when their
+    /// lists become empty.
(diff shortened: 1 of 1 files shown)


state/src/caching/order_metadata_index.rs
```diff
@@ -54,8 +65,55 @@ impl OrderMetadataIndex {
...
+        order_id: OrderIdentifier,
+        matchable_amount: Amount,
+    ) {
+        let (pair, side) = self.get_pair_and_side(&order_id).await.unwrap();
+        let mut index = self.index.write().await;
+        let entry = index.entry(pair).or_insert_with(OrderSideIndex::new);
+        let sorted_vec = entry.entry(side).or_insert_with(SortedVec::new);
+
+        // Remove the old entry
+        let idx = sorted_vec.find_index(|(_, oid)| *oid == order_id).unwrap();
```
akirillo, line 88: we shouldn't unwrap here - if this method is called on an unindexed order, this will panic

state/src/caching/order_metadata_index.rs
```diff
@@ -54,8 +65,55 @@ impl OrderMetadataIndex {
...
+        let side_index = index.get_mut(&pair)?;
+        let sorted_vec = side_index.get_mut(&side)?;
+
+        if let Some(idx) = sorted_vec.find_index(|(_, oid)| oid == order_id) {
+            sorted_vec.remove(idx);
+        }
+
+        // Remove from the reverse index
+        let mut reverse_index = self.reverse_index.write().await;
+        reverse_index.remove(order_id);
```
akirillo, line 114: maybe we should move this to the top of the function, after `get_pair_and_side`, since it's infallible. worried that if somehow the main index doesn't contain an order but the reverse index does, we will return early without removing from the reverse index. this _shouldn't_ ever be the case but might as well keep these indices in sync as much as possible

akirillo (approved):

state/src/caching/order_metadata_index.rs
joeykraut: good point

PR #874, opened 2024-12-21
workers: job-types, state: applicator: emit task completion events

akirillo:
This PR introduces a `TaskCompletion` event type, representing the completion of a task (successful or not). This will be used to index task history, and as such it is basically a wrapper around the `HistoricalTask` type. This event gets emitted in the state applicator in the same code path where we record historical tasks. As such, we check that the local peer is the executor of the task before emitting the event, to prevent duplicate event emission across the cluster.

### Testing
- [x] All unit tests pass (though, with that said, it looks like the new order metadata index tests put each other in deadlock)
- [x] Tested w/ local relayer & event export sidecar, manually inspected events received in SQS to ensure there were no issues in transmission

### DIFF
--- a/state/src/applicator/mod.rs
+++ b/state/src/applicator/mod.rs
@@ -5,7 +5,10 @@ use std::sync::Arc;
 
 use common::types::gossip::ClusterId;
 use external_api::bus_message::SystemBusMessage;
-use job_types::{handshake_manager::HandshakeManagerQueue, task_driver::TaskDriverQueue};
+use job_types::{
+    event_manager::EventManagerQueue, handshake_manager::HandshakeManagerQueue,
+    task_driver::TaskDriverQueue,
+};
 use system_bus::SystemBus;
 
 use crate::{caching::order_cache::OrderBookCache, storage::db::DB, StateTransition};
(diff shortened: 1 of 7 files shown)


state/src/applicator/mod.rs
```diff
@@ -21,13 +24,24 @@ pub mod return_type;
...
+// ---------------------
+// | Convenience Types |
+// ---------------------
 
 /// A type alias for the result type given by the applicator
 pub(crate) type Result<T> = std::result::Result<T, StateApplicatorError>;
 
+/// A struct packaging together all of the worker queues used by the state
+/// applicator
+pub struct ApplicatorQueues {
```
joeykraut, line 36: we basically just instantiate this and immediately destruct it, i think we should just get rid of it.

state/src/applicator/task_queue.rs
```diff
@@ -537,24 +544,39 @@ impl StateApplicator {
...
+            let my_peer_id = tx.get_peer_id()?;
+            let maybe_executor = tx.get_task_assignment(&t.id)?;
+            let is_executor = maybe_executor.is_some_and(|e| e == my_peer_id);
+
+            if is_executor {
+                let event = RelayerEvent::TaskCompletion(TaskCompletionEvent::new(key, t));
+                if let Err(e) = self.config.event_queue.send(event) {
+                    error!("error sending task completion event: {e}");
+                }
+            }
```
joeykraut, line 579: shouldn't this happen outside the `if` block here? we will have `tx.get_historical_state_enabled() -> false` for our nodes internally.

joeykraut (approved):
looks good, don't think the `ApplicatorQueues` type is useful code though

state/src/applicator/task_queue.rs
akirillo: great catch

state/src/applicator/mod.rs
akirillo: sure, just found myself agreeing with `clippy::too_many_arguments`, but idrc

PR #875, opened 2024-12-22
state: interface: Use new order book cache in interface methods

joeykraut:
### Purpose
This PR adds back in write paths for the order book cache and uses the new order book cache in the state interface. This involves two caller locations:
- The admin API, which roughly just uses renamed methods
- The matching engine, which now constructs an order book filter based on the order it is trying to match

### Testing
- [x] Unit tests pass
- [x] Testing in testnet with various match flows

### DIFF
--- a/state/src/caching/order_metadata_index.rs
+++ b/state/src/caching/order_metadata_index.rs
@@ -49,6 +49,14 @@ impl OrderMetadataIndex {
         }
     }
 
+    /// Get all orders in the index
+    ///
+    /// No sort ordering is guaranteed, given that the units on which individual
+    /// orders are sorted may differ
+    pub async fn get_all_orders(&self) -> Vec<OrderIdentifier> {
+        self.reverse_index.read().await.keys().copied().collect()
+    }
+
     /// Get the pair and side for a given order_id
     pub async fn get_pair_and_side(&self, order_id: &OrderIdentifier) -> Option<(Pair, OrderSide)> {
         self.reverse_index.read().await.get(order_id).cloned()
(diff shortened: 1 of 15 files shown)


state/src/caching/order_metadata_index.rs
```diff
@@ -114,6 +118,10 @@ impl OrderMetadataIndex {
             sorted_vec.remove(idx);
         }
 
+        // Remove from the reverse index
+        let mut reverse_index = self.reverse_index.write().await;
+        reverse_index.remove(order_id);
+
```
akirillo, line 124: did we want to move this back under the fallible code?

state/src/replication/state_machine/snapshot.rs
```diff
@@ -256,7 +256,8 @@ impl StateMachine {
         let order_cache_clone = self.applicator.config.order_cache.clone();
         let jh = tokio::task::spawn_blocking(move || {
             Self::copy_db_data(&snapshot_db, &db_clone)?;
-            tokio::spawn(async move {
+            let rt = tokio::runtime::Handle::current();
+            rt.block_on(async move {
```
akirillo, line 260: why do we need this to be blocking?

akirillo (approved):

state/src/caching/order_metadata_index.rs
joeykraut: yes, the code above it depends on the entry still being in place

state/src/replication/state_machine/snapshot.rs
joeykraut: because we're in a synchronous context

PR #877, opened 2025-01-03
state, workers: allow for actively-managed historical state

akirillo:
This PR tweaks the state & API code paths dependent on historical state being enabled. We allow for tracking "actively-managed historical state," i.e. historical state pertaining to objects (order metadata, tasks) which may still be updated by the relayer. Concretely, this means:
- Tasks
    - We continue streaming task status updates over websocket via _both_ the task status and task history topics. We continue to enable the task history topic since it is a reliable source of all task status updates for a given wallet.
    - We do not write tasks to the task history table if historical state is disabled.
- Orders
    - We now allow writing to / reading from the order history table even if historical state is disabled. However, if historical state is disabled we remove orders from the table once they reach a terminal state.
    - This allows us to continue streaming order metadata updates over the associated websocket topic

### Testing
- [x] All unit tests pass
- [x] Testing in testnet w/ historical state enabled
- [x] Testing in testnet w/ historical state disabled - requires testnet frontend integration w/ historical state engine (https://github.com/renegade-fi/renegade-frontend/pull/223)
- [ ] Parity testing w/ current frontend / historical state - requires migration to be complete

### DIFF
--- a/workers/task-driver/src/tasks/update_wallet.rs
+++ b/workers/task-driver/src/tasks/update_wallet.rs
@@ -199,6 +199,11 @@ pub struct UpdateWalletTask {
     pub task_state: UpdateWalletTaskState,
     /// The event manager queue
     pub event_queue: EventManagerQueue,
+    /// The pending event to emit after the task is complete.
+    /// We construct this event before emitting it, as we may need to
+    /// access state that is deleted by the end of the task to do so
+    /// (e.g., in the case of an order cancellation).
+    pub completion_event: Option<RelayerEvent>,
 }
 
 #[async_trait]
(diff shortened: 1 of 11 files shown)


workers/task-driver/src/tasks/update_wallet.rs
```diff
@@ -199,6 +199,11 @@ pub struct UpdateWalletTask {
     pub task_state: UpdateWalletTaskState,
     /// The event manager queue
     pub event_queue: EventManagerQueue,
+    /// The pending event to emit after the task is complete.
+    /// We construct this event before emitting it, as we may need to
+    /// access state that is deleted by the end of the task to do so
+    /// (e.g., in the case of an order cancellation).
+    pub pending_event: Option<RelayerEvent>,
```
joeykraut, line 206: nit: maybe `completion_event`? slightly clearer as to the intent of the event

joeykraut (approved):

PR #880, opened 2025-01-07
api-server: admin: Add endpoint to refresh external match fees

joeykraut:
### Purpose
This PR adds an endpoint to refresh external match fees with admin authentication

### Testing
- [x] Unit tests pass
- [x] Tested locally
- [ ] Will test in testnet once admin lambda is upgraded to support fee overrides

### DIFF
--- a/workers/api-server/src/http/admin.rs
+++ b/workers/api-server/src/http/admin.rs
@@ -4,12 +4,12 @@
 // | Admin Route Handlers |
 // ------------------------
 
-use arbitrum_client::constants::Chain;
+use arbitrum_client::{client::ArbitrumClient, constants::Chain};
 use async_trait::async_trait;
 use circuit_types::{fixed_point::FixedPoint, Amount};
 use common::types::{
     tasks::UpdateWalletTaskDescriptor,
-    token::Token,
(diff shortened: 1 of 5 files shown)


workers/api-server/src/http/admin.rs
```diff
@@ -537,6 +537,44 @@ impl TypedHandler for AdminRefreshTokenMappingHandler {
...
+        _params: UrlParams,
+        _query_params: QueryParams,
+    ) -> Result<Self::Response, ApiServerError> {
+        info!("Refreshing external match fees from contract");
+
+        for token in get_all_tokens() {
+            let addr = token.get_ethers_address();
+            let fee = self.arbitrum_client.get_external_match_fee(addr).await?;
+
+            set_external_match_fee(&token.get_addr_biguint(), fee);
```
akirillo, line 571: Why not have this handler also refresh the default protocol fee?

akirillo (approved):

workers/api-server/src/http/admin.rs
joeykraut: it's a `OnceLock` so it can't be refreshed, we can add that functionality if we like

PR #883, opened 2025-01-09
workspace: Gate long tests behind `ci` feature flag

joeykraut:
### Purpose
This PR gates long-running unit tests behind a `ci` feature flag, and enables that flag (disabling the tests) in github workflows. This allows us to run tests in debug mode, so that the total time taken to run all tests does not exceed what the test runners allow

### Testing
- [x] All tests pass
- [x] CI passes 🥳

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,4 +18,4 @@ jobs:
       uses: actions-rs/cargo@v1
       with: 
         command: test
-        args: --workspace --release --all-features --verbose -- --skip integration
+        args: --workspace --features "ci" --verbose -- --skip integration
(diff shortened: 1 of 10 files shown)


.github/workflows/test.yml
```diff
@@ -18,4 +18,4 @@ jobs:
       uses: actions-rs/cargo@v1
       with: 
         command: test
-        args: --workspace --release --all-features --verbose -- --skip integration
```
akirillo, line 21: do we switch to debug mode because compilation takes up too much of the time allotted to the runner?

joeykraut: Yes

PR #884, opened 2025-01-10
workers: api-server: auth, router: include query string in path for auth

akirillo:
This PR tweaks the wallet & admin auth handlers to expect the query string to be included in the path when validating an authenticated request's HMAC.

However, since there are clients that are currently implemented with the assumption that query strings should be excluded from the HMAC signature, we introduce some shim code here that first attempts to validate auth with the expectation that query strings are included in the path, then falls back to to validating auth while excluding query strings from the path. This is a temporary provision to ensure that deploying this change in the relayer does not require coupled client-side upgrades.

### Testing
I tested this by running a local relayer & local frontend, and attempted to invoke the task history endpoint for a wallet with the `task_history_len` parameter set. I tested this both ways, with the query string included & excluded from the client-side request HMAC, and asserted that we got successful responses from the relayer in both cases.

### DIFF
--- a/workers/api-server/src/auth/mod.rs
+++ b/workers/api-server/src/auth/mod.rs
@@ -2,7 +2,7 @@
 
 use common::types::wallet::{keychain::HmacKey, WalletIdentifier};
 use external_api::auth::validate_expiring_auth;
-use hyper::HeaderMap;
+use hyper::{HeaderMap, StatusCode};
 use state::State;
 use tracing::warn;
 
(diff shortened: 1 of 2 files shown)


workers/api-server/src/auth/mod.rs
```diff
@@ -99,12 +119,27 @@ impl AuthMiddleware {
             return Err(not_found(ERR_ADMIN_API_DISABLED));
         }
 
+        // TODO: auth w/ path+query, fall back to auth w/ path
```
joeykraut, line 122: is this todo still relevant?

joeykraut (approved):

workers/api-server/src/auth/mod.rs
akirillo: no, good catch

PR #887, opened 2025-01-15
workers: event-manager: record send/export failure metrics

akirillo:
This PR adds recording of dropped event metrics throughout the relayer. Namely, we increment a counter whenever we fail to send an event to the event manager, whenever the event manager fails to export, and whenever the event export sidecar fails to export.

### Testing
- [x] Testing w/ local relayer where event export sidecar gets disconnected


workers/event-manager/src/manager.rs
```diff
@@ -119,6 +122,32 @@ impl EventManagerExecutor {
...
+async fn try_export_event(event_bytes: Vec<u8>, sink: &mut FramedUnixSink) {
+    match sink.send(event_bytes.into()).await {
+        Ok(_) => (),
+        Err(e) => {
+            metrics::counter!(NUM_EVENT_EXPORT_FAILURES_METRIC).increment(1);
+            error!("Failed to export event: {e}");
+        },
+    }
+}
+
```
joeykraut, line 150: given that these two are incrementing the same metric, why not just catch errors in the caller (or a grand-caller) and increment the metric there. that's also future proof against more fallible logic that would need to increment this metric.

joeykraut (approved):

workers/event-manager/src/manager.rs
akirillo: sure, good call

PR #890, opened 2025-01-19
task-driver: settle-match-external: Report metrics regardless of timeout

joeykraut:
### Purpose
This PR changes the way that metrics are reported in the `settle-external-match` task to watch for settlement and report on a separate thread. This allows the main thread to wait for a configurable timeout without early-exiting the metrics report.

### Testing
- [x] Testing in testnet


sehyunc (approved):
just one unused import clippy warning

PR #895, opened 2025-01-21
task-driver: settle-match-external: Filter settlement by selector

joeykraut:
### Purpose
This PR filters the settlement waiter in `settle-match-external` by the selectors used for atomic matches.

### Testing
- [x] Unit tests pass
 - [ ] Testing in testnet


akirillo (changes requested):
I think this method of filtering by selectors won't work for nested calls to the atomic match settlement contract methods

akirillo (approved):

PR #896, opened 2025-01-23
workers: task-driver: state_migration: purge historical state startup task

akirillo:
This PR adds a new state migration for purging historical state. Namely, this means removing terminal orders from order history and all task history, for all wallets. We no longer need to track this in the relayer since the historical state engine is active.

### Testing
- [x] Unit tests for new state methods pass
- [x] Testing in testnet

### DIFF
--- a/state/src/storage/tx/order_history.rs
+++ b/state/src/storage/tx/order_history.rs
@@ -115,6 +115,13 @@ impl<'db> StateTxn<'db, RW> {
         self.write_order_history(wallet_id, orders)
     }
 
+    /// Purges terminal orders from a wallet's order history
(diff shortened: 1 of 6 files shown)


state/src/storage/tx/order_history.rs
```diff
@@ -115,6 +115,13 @@ impl<'db> StateTxn<'db, RW> {
         self.write_order_history(wallet_id, orders)
     }
 
+    /// Purges terminal orders from a wallet's order history
```
joeykraut, line 118: will you add a TODO to remove this? don't want it sticking around in the code. same elsewhere in the `state` crate

joeykraut (approved):

PR #914, opened 2025-03-14
workers: api-server, task-driver: set native asset external match fee

akirillo:
In this PR, we set the relayer's local view of the native ETH external match fee on node startup and in the external match fees refresh endpoint. This ensures that, for example, we return the correct expected receive & fee amounts when returning an external match quote.

### Testing
- [ ] Test in testnet

### DIFF
--- a/workers/api-server/src/http/admin.rs
+++ b/workers/api-server/src/http/admin.rs
@@ -4,6 +4,8 @@
 // | Admin Route Handlers |
 // ------------------------
 
+use std::iter;
+
 use arbitrum_client::{client::ArbitrumClient, constants::Chain};
 use async_trait::async_trait;
 use circuit_types::{fixed_point::FixedPoint, Amount};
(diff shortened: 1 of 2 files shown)


workers/api-server/src/http/admin.rs
```diff
@@ -571,6 +576,16 @@ impl TypedHandler for AdminRefreshExternalMatchFeesHandler {
...
 
+        // Set the fee override for the native asset
+        let native_asset_addr: Address =
+            NATIVE_ASSET_ADDRESS.parse().expect("invalid native asset address");
+
+        let fee = self.arbitrum_client.get_external_match_fee(native_asset_addr).await?;
+        let native_asset_addr_biguint =
+            biguint_from_hex_string(NATIVE_ASSET_ADDRESS).expect("invalid native asset address");
+
+        set_external_match_fee(&native_asset_addr_biguint, fee);
```
joeykraut, line 587: probably cleaner in both the cases this PR handles to combine this with the iterator above, we can use different types and iterate over ethers addresses, hex strings, bigints, etc; doesn't have to be over tokens.

joeykraut (approved):

PR #918, opened 2025-03-22
arbitrum-client, common: Remove dependency on `contracts-common`

joeykraut:
### Purpose
This PR removes the dependency on `contracts-common` from the relayer repo, which previously caused a cyclic dependency. To do so, we directly replicate the functionality from the contracts into the relayer.

### Testing
- [x] All unit tests pass
- [ ] Testing in testnet

### DIFF
--- a/arbitrum-client/Cargo.toml
+++ b/arbitrum-client/Cargo.toml
@@ -55,9 +55,6 @@ tokio = { workspace = true }
 tracing = { workspace = true }
 rand = { workspace = true, optional = true }
 
-# === Contracts Repo Dependencies === #
-contracts-common = { git = "https://github.com/renegade-fi/renegade-contracts.git" }
-
 [dev-dependencies]
 clap = { version = "4.0", features = ["derive"] }
 eyre = { workspace = true }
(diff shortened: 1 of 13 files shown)


arbitrum-client/Cargo.toml
```diff
@@ -55,9 +55,6 @@ tokio = { workspace = true }
 tracing = { workspace = true }
 rand = { workspace = true, optional = true }
 
-# === Contracts Repo Dependencies === #
-contracts-common = { git = "https://github.com/renegade-fi/renegade-contracts.git" }
```
akirillo, line 59: tight

common/src/types/tasks/descriptors/mod.rs
```diff
@@ -229,7 +228,7 @@ pub mod mocks {
 
         // Serialize the commitment, uses the contract's serialization here:
         //  https://github.com/renegade-fi/renegade-contracts/blob/main/contracts-common/src/custom_serde.rs#L82-L87
-        let comm_bytes = new_wallet_comm.inner().serialize_to_bytes();
+        let comm_bytes = new_wallet_comm.to_bytes_be();
```
akirillo, line 231: as in `update_wallet.rs`, can you update the comment here to specify that we are matching, not using, the serialization used in the contracts

akirillo (approved):

PR #919, opened 2025-03-22
util: wallet: Add wallet utils and cleanup repo imports + flags

joeykraut:
### Purpose
This PR cleans up workspace imports and the feature flags in the `util` crate. I also added a helper for migrating from `ethers` to `alloy` which will be done as we deploy the solidity implementation.

### Testing
- [x] All tests pass


sehyunc (approved):
nice

PR #920, opened 2025-03-22
common: wallet: keychain: Add generic payload signing method

joeykraut:
### Purpose
This PR breaks out the wallet commitment signature helper into a separate helper that signs arbitrary bytes. This is useful for generating withdrawal authorization, which comprises an `sk_root` signature of the ABI-encoded transfer. 

### Testing
- [x] Unit tests pass
- [ ] Test in testnet


akirillo (approved):
🧼

PR #924, opened 2025-03-31
arbitrum-client: event_indexing: prefetch nullifier spent events

akirillo:
This PR fixes the logic for awaiting a nullifier spend event to also pre-fetch events matching the filter before polling the stream. This is because the event stream only includes events that are emitted after the initial poll of the stream, so if a transaction spending the given nullifier is settled before the filter is created, we will miss the associated nullifier spent event.

### Testing
- [ ] Test in testnet

### DIFF
--- a/arbitrum-client/src/client/event_indexing.rs
+++ b/arbitrum-client/src/client/event_indexing.rs
@@ -224,13 +224,23 @@ impl ArbitrumClient {
             .await
             .map_err(err_str!(ArbitrumClientError::EventQuerying))?;
 
-        // Await the event with a timeout
-        let next_event = tokio::time::timeout(timeout, stream.next()).await;
-        let (_event, meta) = next_event
-            .map_err(|_| ArbitrumClientError::event_querying(ERR_NULLIFIER_SPENT_TIMEOUT))? // timeout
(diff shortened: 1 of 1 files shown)


arbitrum-client/src/client/event_indexing.rs
```diff
@@ -224,13 +224,23 @@ impl ArbitrumClient {
...
-        // Await the event with a timeout
-        let next_event = tokio::time::timeout(timeout, stream.next()).await;
-        let (_event, meta) = next_event
-            .map_err(|_| ArbitrumClientError::event_querying(ERR_NULLIFIER_SPENT_TIMEOUT))? // timeout
-            .ok_or(ArbitrumClientError::event_querying(ERR_EVENT_STREAM_CLOSED))? // no event
-            .map_err(err_str!(ArbitrumClientError::event_querying))?; // query error
-        let tx_hash = meta.transaction_hash;
+        // Pre-fetch events from before the creation of the filter,
+        // in case the nullifier has already been spent
+        let prefetched_events =
```
joeykraut, line 229: what happens if a nullifier is spent after this prefetch occurs, but before we open the event stream?

joeykraut (approved):

PR #931, opened 2025-04-10
storage: applicator: task-queue: Double write task updates

joeykraut:
### Purpose
This PR changes the applicator logic for the task queue to double write between the v1 and v2 implementations of the task queue. This will be deployed independently and subsequently migrated into a version that _reads from_ the v2 task queues and implements their locking semantics. 

I also made a few cleanups for unused helpers in the applicator, and removed the task queue length metrics (also unused).

### Testing
- [x] All unit tests pass 
- [x] Added double write tests

### DIFF
--- a/state/src/storage/tx/task_queuev2.rs
+++ b/state/src/storage/tx/task_queuev2.rs
@@ -72,6 +72,7 @@ impl TaskQueue {
     }
 
     /// Check whether the given task is running
+    #[cfg(test)]
     pub fn is_running(&self, id: &TaskIdentifier) -> bool {
         self.running_tasks.contains(id)
     }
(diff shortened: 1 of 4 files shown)


state/src/storage/tx/task_queuev2.rs
```diff
@@ -281,10 +279,13 @@ impl<'db> StateTxn<'db, RW> {
             None => return Err(StorageError::not_found(ERR_TASK_NOT_FOUND)),
         };
 
-        // Mark the task as running if the new state is running
+        // Update the running status of the task
         if new_state.is_running() {
             self.mark_task_running(id, key)?;
+        } else {
+            self.mark_task_not_running(id, key)?;
         }
```
sehyunc, line 287: Does seem more like your style to have a fn signature `update_task_running_status(id, key, running: boolean)` but maybe there's a benefit to separating these?

state/src/storage/tx/task_queuev2.rs
```diff
@@ -208,9 +210,15 @@ impl<'db> StateTxn<'db, RW> {
         key: &TaskQueueKey,
         task: &QueuedTask,
     ) -> Result<(), StorageError> {
+        // Index the task into the queue
         let mut queue = self.get_task_queue(key)?;
         queue.add_serial_task_front(task.id);
 
+        // If we add to the front of the queue, we must check if the task has
+        // been added in the running state
```
sehyunc, line 218: Nice catch, seems annoying to keep the `running` array in sync manually like this, but don't see a clean alternative. Looks like you've covered all the cases of a task entering/leaving the running state (preemption + task state transition).

sehyunc (approved):
Very clean and readable👍🏻

state/src/storage/tx/task_queuev2.rs
joeykraut: yeah we may not even need the running queue in the end, i'll remove it if that's the case

state/src/storage/tx/task_queuev2.rs
joeykraut: no strong preference here from me, I tend not to like flags on function boundaries unless necessary. I usually prefer to mux a flag at a higher call site; which creates more function calls, but I assume the compiler will inline as necessary. that's all pretty "in the weeds" stuff though

PR #934, opened 2025-04-10
state: storage: task-queuev2: Remove running tasks from `TaskQueue`

joeykraut:
### Purpose
This PR removes the `running_tasks` field from the `TaskQueue`. This won't be used for any task queue logic, so it's unnecessary to track.

### Testing
- [x] All tests pass


sehyunc (approved):
clean

PR #935, opened 2025-04-10
state: storage: task-queuev2: Allow task to preempt multiple queues

joeykraut:
### Purpose
This PR allows a task to preempt multiple queues by changing the task to queue key mapping from `id -> id` to `id -> vec<id>`. 

### Testing
- [x] Unit tests pass
- [x] Added tests for behavior with multiple queues

### DIFF
--- a/state/src/applicator/task_queue.rs
+++ b/state/src/applicator/task_queue.rs
@@ -285,7 +285,7 @@ impl StateApplicator {
         tx.add_assigned_task(executor, &task.id)?;
         // TODO(@joeykraut): Remove the v1 implementation once we migrate
         tx.add_task_front(&key, task)?;
-        double_write_helper(tx.preempt_queue_with_serial(&key, task));
+        double_write_helper(tx.preempt_queue_with_serial(&[key], task));
         self.publish_task_updates(key, task);
 
         Ok(ApplicatorReturnType::None)
--- a/state/src/storage/tx/task_queuev2.rs
+++ b/state/src/storage/tx/task_queuev2.rs
@@ -43,6 +43,11 @@ fn task_key(id: &TaskIdentifier) -> String {
     format!("task-{}", id)
 }
 
+/// Get the storage key for the task to queue(s) mapping
+fn task_to_queue_key(id: &TaskIdentifier) -> String {
+    format!("task-to-queue-{id}")
+}
+
 /// The preemption state of a task queue
 #[derive(Default, Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq)]
 pub enum TaskQueuePreemptionState {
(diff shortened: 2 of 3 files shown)


state/src/storage/tx/task_queuev2.rs
```diff
@@ -313,39 +323,45 @@ impl<'db> StateTxn<'db, RW> {
...
-            return Ok(false);
-        };
+        // Index the task into the queues
+        for queue_key in queues.iter() {
+            let mut queue = self.get_task_queue(queue_key)?;
+            if !queue.preempt_with_serial_task(task.id) {
+                return Ok(false);
+            };
+
+            self.write_task_queue(queue_key, &queue)?;
```
sehyunc, line 342: I noticed that this will modify the first queue before checking if the second queue can be preempted. If the second queue cannot be preempted, the method returns false but the first queue remains modified. Does this not break atomicity, or am I misunderstanding? Seems like we should check for preemptibilty of all queues before modifying.

sehyunc: Here's a test case illustrating this:
```rust
    /// Tests that preempt_queue_with_serial is atomic - if any queue cannot be
    /// preempted, no queues should be modified
    #[test]
    #[allow(non_snake_case)]
    fn test_invalid_serial_preemption__atomicity() -> Result<(), StorageError> {
        // Setup the mock
        let db = mock_db();
        let tx = db.new_write_tx()?;
        let key1 = TaskQueueKey::new_v4();
        let key2 = TaskQueueKey::new_v4();

        // Add a serial preemptive task to the second queue to make it not preemptable
        let preemptive_task2 = mock_queued_task(key2);
        tx.preempt_queue_with_serial(&[key2], &preemptive_task2)?;

        // Try to preempt both queues with a new serial task
        let new_serial_task = mock_queued_task(key1);
        let success = tx.preempt_queue_with_serial(&[key1, key2], &new_serial_task)?;
        assert!(!success);

        // Check that the queue state is updated correctly
        let task_queue1 = tx.get_task_queue(&key1)?;
        let task_queue2 = tx.get_task_queue(&key2)?;
        let expected_queue1 = TaskQueue::default();
        let expected_queue2 = TaskQueue {
            serial_tasks: vec![preemptive_task2.id],
            preemption_state: TaskQueuePreemptionState::SerialPreemptionQueued,
            ..Default::default()
        };
        assert_eq!(task_queue1, expected_queue1);
        assert_eq!(task_queue2, expected_queue2);
        Ok(())
    }
```

joeykraut: we can't really enforce atomicity _within_ the transaction in a clean way, rather we assume the transaction creator (the applicator) will revert the tx if we return false. We could error here to indicate this, though I feel `Result<bool>` more cleanly communicates this. happy to error if you think that would be better

state/src/applicator/task_queue.rs
```diff
@@ -285,7 +285,7 @@ impl StateApplicator {
         tx.add_assigned_task(executor, &task.id)?;
         // TODO(@joeykraut): Remove the v1 implementation once we migrate
         tx.add_task_front(&key, task)?;
-        double_write_helper(tx.preempt_queue_with_serial(&key, task));
+        double_write_helper(tx.preempt_queue_with_serial(&[key], task));
```
sehyunc, line 288: So we'll need to check the return value here and handle failures to preempt right? Not a request for change, just want to wrap up this train of thought.

joeykraut: yes when we switch over to the new implementation we will. for now we're relying on the fact that the paused state is equivalent to the `SerialPreemptionQueued` state. so it's currently impossible to violate the preemption logic with double writes in their current form. We'll check for errors and revert in the new method `EnqueuePreemptiveTask`

sehyunc (approved):

PR #937, opened 2025-04-11
state: applicator: task-queue: Implement `EnqueuePreemptiveTask`

joeykraut:
### Purpose
This PR adds the implementation of `EnqueuePreemptiveTask`. This state transition merely calls out to the preemption helpers in the storage layer and handles task assignment. For the migration, we also double write to the old task queue implementation.

The version in this PR will be deployed ahead of a version which switches over to using the new task queue for all preemptive tasks.

### Todo
In a follow up, deployed separately:
- Change the pop and transition task helpers to use the new task queue
- Further tests with these new implementations

### Testing
- [x] All unit tests pass
- [x] Added basic tests for `EnqueuePreemptiveTask`

### DIFF
--- a/state/src/applicator/task_queue.rs
+++ b/state/src/applicator/task_queue.rs
@@ -32,6 +32,8 @@ const PENDING_STATE: &str = "Pending";
 const ERR_UNASSIGNED_TASK: &str = "task not assigned";
 /// Error emitted when a key cannot be found for a task
 const ERR_NO_KEY: &str = "key not found for task";
+/// Error emitted when a task preemption fails
+const ERR_TASK_PREEMPTION: &str = "task queue cannot be preempted";
 
 /// Construct an invalid task queue key error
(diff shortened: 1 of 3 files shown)


state/src/applicator/task_queue.rs
```diff
@@ -603,6 +634,27 @@ impl StateApplicator {
...
+    /// Try preempting a set of task queues with a task, returning a transition
+    /// rejection if this fails
+    fn try_preempt_queues(
+        &self,
+        keys: &[TaskQueueKey],
+        task: &QueuedTask,
+        is_serial: bool,
+        tx: &StateTxn<'_, RW>,
+    ) -> Result<()> {
+        let did_preempt = if is_serial {
```
sehyunc, line 647: I understand it now.

sehyunc (approved):

PR #938, opened 2025-04-11
state: applicator: task-queue: Use new task queue in pop method

joeykraut:
### Purpose
This PR changes the `pop_task` method to use the new queue primarily and only double write to the old queue.

I also moved the logic to enforce that a running task has not committed, and to requeue that task into the storage layer.

### Testing
- [x] Unit tests pass

### DIFF
--- a/state/src/applicator/task_queue.rs
+++ b/state/src/applicator/task_queue.rs
@@ -204,10 +204,8 @@ impl StateApplicator {
 
         // Assign the task and run it if possible
         tx.add_assigned_task(executor, &task.id)?;
-        if tx.can_task_run(&task.id)? {
-            // TODO(@joeykraut): We don't need to pass a queue key after migration
-            self.maybe_run_task(keys[0], task, &tx)?;
-        }
+        // TODO(@joeykraut): We don't need to pass a queue key after migration
+        self.maybe_run_task(keys[0], task, &tx)?;
 
         // We double write to the old queue implementation to ensure consistency during
         // the migration. However, we do not add a preemptive task to the original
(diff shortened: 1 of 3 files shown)


state/src/applicator/task_queue.rs
```diff
@@ -233,7 +206,7 @@ impl StateApplicator {
         tx.add_assigned_task(executor, &task.id)?;
         if tx.can_task_run(&task.id)? {
```
sehyunc, line 207: Seems like a redundant check since `maybe_run_task` already verifies if task `can_run`.

state/src/storage/tx/task_queuev2.rs
```diff
@@ -360,49 +402,62 @@ impl<'db> StateTxn<'db, RW> {
...
+            }
+
+            // 2. Move the existing task back to the queued state
             let mut queue = self.get_task_queue(queue_key)?;
+            if let Some(t) = queue.serial_tasks.first() {
+                self.requeue_task(t)?;
+            }
+
+            // 3. Preempt the queue
             if !queue.preempt_with_serial_task(task.id) {
```
sehyunc, line 420: Seems like there's a redundant `can_preempt_serial` safety check in this method since we have the dedicated `is_serial_preemption_safe` method. Unless there's a scenario in which `preempt_with_serial_task` would be called without the safety check?

sehyunc (approved):
just noticed some seemingly redundant safety checks

state/src/applicator/task_queue.rs
joeykraut: yes good catch

state/src/storage/tx/task_queuev2.rs
joeykraut: There's no scenario currently that would call one and not the other. My rationale here is that the `state != SerialPreemptionQueued` check is very cheap, and I'd prefer both layers (storage and the `TaskQueue` data structure) to be correct independent of one another's implementation.

PR #941, opened 2025-04-13
state: Remove old task queue impl and update all callers

joeykraut:
### Purpose
This PR removes the original task queue implementation and replaces all callsites with analogous calls to the new implementation.

### Testing
- [x] All tests pass
- [x] Tested locally
- [x] Testing in testnet


sehyunc (approved):
very clean

PR #947, opened 2025-04-16
chain-events: listener: Record external match metrics from spent nullifier

joeykraut:
### Purpose
This PR removes the nullifier watch loop from the settlement task. This is necessary as we now emit the same bundle to multiple clients who mark their requests with `allow_shared`. The existing logic in the settlement task would then double count each settlement for which it emits multiple bundles.

Instead, we move the metrics and wallet refresh logic to the `chain-events` worker, which will see each nullifier exactly once.

### Testing
- [x] Unit tests pass
- [ ] Testing in testnet

### DIFF
--- a/workers/chain-events/src/listener.rs
+++ b/workers/chain-events/src/listener.rs
@@ -7,7 +7,7 @@ use arbitrum_client::{abi::NullifierSpentFilter, client::ArbitrumClient};
 use circuit_types::wallet::Nullifier;
 use common::types::CancelChannel;
 use constants::in_bootstrap_mode;
-use ethers::prelude::StreamExt;
+use ethers::{prelude::StreamExt, types::H256 as TxHash};
 use job_types::{
     handshake_manager::{HandshakeManagerJob, HandshakeManagerQueue},
     network_manager::NetworkManagerQueue,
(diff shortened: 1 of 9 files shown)


workers/chain-events/src/listener.rs
```diff
@@ -129,64 +134,128 @@ impl OnChainEventListenerExecutor {
...
+    async fn handle_nullifier_wallet_updates(
         &self,
         nullifier: Nullifier,
+        tx: TxHash,
     ) -> Result<(), OnChainEventListenerError> {
-        // Get the wallet ID that this nullifier belongs to
+        // If the current node is not the update handler, sleep for a random timeout
+        // then check that the nullifier state updates have been processed.
+        // We do this as a crash recovery mechanism to ensure that the updates are
+        // processed even if the selected node is crashed
```
sehyunc, line 152: In the old code I see this technique was used to prevent a thundering herd—it seems like that is still the case but it has the added benefit of crash recovery. Curious to hear if you changed the comment because you know the crash scenario to be more common/concerning, or some other reason.

sehyunc (approved):

workers/chain-events/src/listener.rs
joeykraut: nothing concrete really, the old code did effectively recover from crashes as well (each node would try to update the internal state on a jittered timeout), but it's just more important now that we rely on this listener loop to handle external matches. So no specific reason, just wanted to highlight the crash recovery properties

PR #952, opened 2025-04-18
chain-events: listener: Use `subscribe_logs` for websocket client

joeykraut:
### Purpose
This PR fixes the chain events listener's websocket implementation by using the `subscribe_logs` method directly rather than sending `eth_getFilterChanges` through the websocket as ethers will do by default. In the process, I also migrated this crate to use alloy.

### Testing
- [x] Tested locally
- [x] Testing in testnet

### DIFF
--- a/core/src/main.rs
+++ b/core/src/main.rs
@@ -74,6 +74,12 @@ const TERMINATION_TIMEOUT_MS: u64 = 10_000; // 10 seconds
 ///     4. Allocate a thread to monitor the worker for faults
 #[tokio::main]
 async fn main() -> Result<(), CoordinatorError> {
+    // Set the default crypto provider for the process, this will be used by
+    // websocket listeners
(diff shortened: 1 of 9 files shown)


core/src/main.rs
```diff
@@ -74,6 +74,12 @@ const TERMINATION_TIMEOUT_MS: u64 = 10_000; // 10 seconds
 ///     4. Allocate a thread to monitor the worker for faults
 #[tokio::main]
 async fn main() -> Result<(), CoordinatorError> {
+    // Set the default crypto provider for the process, this will be used by
+    // websocket listeners
+    rustls::crypto::ring::default_provider()
+        .install_default()
+        .expect("Failed to install rustls crypto provider");
```
joeykraut, line 81: @sehyunc you may have to replicate this in the auth server.

PR #953, opened 2025-04-18
external-api: price-report: Add `/v0/token-prices` endpoint

joeykraut:
### Purpose
This PR adds the `v0/token-prices` endpoint, which fetches all token prices for all pairs supported by the relayer.

### Testing
- [x] Tested locally
- [ ] Testing in testnet

### DIFF
--- a/external-api/src/http/price_report.rs
+++ b/external-api/src/http/price_report.rs
@@ -1,7 +1,12 @@
 //! Groups price reporting API types
 
-use common::types::{exchange::PriceReporterState, token::Token};
+use common::types::{exchange::PriceReporterState, token::Token, Price};
 use serde::{Deserialize, Serialize};
+
+use crate::{
+    deserialize_price_from_string, deserialize_token_from_hex_string, serialize_price_as_string,
+    serialize_token_as_hex_string, types::ApiToken,
+};
 
 // ---------------
 // | HTTP Routes |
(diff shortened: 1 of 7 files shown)


workers/api-server/src/http/price_report.rs
```diff
@@ -1,15 +1,31 @@
...
 use crate::{
     error::{internal_error, ApiServerError},
     router::{QueryParams, TypedHandler, UrlParams},
     worker::ApiServerConfig,
 };
 
+/// Tokens filtered from the supported token endpoint
+const FILTERED_TOKENS: [&str; 2] = [USD_TICKER, USDT_TICKER];
+/// Tokens filtered from the token prices endpoint
+const FILTERED_TOKENS_PRICES: [&str; 3] = [USDT_TICKER, USDT_TICKER, USDC_TICKER];
```
akirillo, line 27: I think this was supposed to be:
```suggestion
const FILTERED_TOKENS_PRICES: [&str; 3] = [USD_TICKER, USDT_TICKER, USDC_TICKER];
```

external-api/src/http/price_report.rs
```diff
@@ -25,3 +34,40 @@ pub struct GetPriceReportResponse {
...
+
+/// Price information for a token
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct TokenPrice {
+    /// The base token
+    #[serde(
+        serialize_with = "serialize_token_as_hex_string",
+        deserialize_with = "deserialize_token_from_hex_string"
+    )]
+    pub base_token: Token,
```
akirillo, line 60: Could be clean to have this be an `ApiToken`, and implement `Serialize` / `Deserialize` on that type to use hex strings

akirillo (approved):

workers/api-server/src/http/price_report.rs
joeykraut: good eye

external-api/src/http/price_report.rs
joeykraut: only issue with that, the `ApiToken` has two fields `address` and `symbol`, both of which are expected by clients. So we can't quite serialize directly to an address on that type directly. We _could_ in the context of this response type using field attributes like i've done here, but I don't see much point in adding extra information (`symbol`) just to serialize it out.

PR #955, opened 2025-04-18
circuits: zk-circuits: wallet-create: Compute full wallet share comm

joeykraut:
### Purpose
This PR changes the `VALID WALLET CREATE` circuit to compute the full wallet commitment in-circuit, as opposed to by the contract. 

### Performance
For `VALID WALLET CREATE` this increases the circuit size to the next power of two, but only implies a minimal impact on prover latency, from ~150ms -> ~324ms

### Testing
- [x] All tests pass

### DIFF
--- a/circuits/src/zk_circuits/valid_wallet_create.rs
+++ b/circuits/src/zk_circuits/valid_wallet_create.rs
@@ -45,10 +45,13 @@ where
         cs: &mut PlonkCircuit,
     ) -> Result<(), CircuitError> {
         // Validate the commitment given in the statement is a valid commitment to the
-        // private secret shares
-        let commitment =
-            WalletGadget::compute_private_commitment(&witness.private_wallet_share, cs)?;
-        cs.enforce_equal(commitment, statement.private_shares_commitment)?;
+        // wallet shares
+        let commitment = WalletGadget::compute_wallet_share_commitment(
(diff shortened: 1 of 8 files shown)


circuits/src/zk_circuits/valid_wallet_create.rs
```diff
@@ -124,7 +127,7 @@ where
     [(); MAX_BALANCES + MAX_ORDERS]: Sized,
 {
     /// The commitment to the private secret shares of the wallet
```
akirillo, line 129: Nit: we should update this doc comment

test-helpers/src/contract_interaction.rs
```diff
@@ -43,7 +43,7 @@ pub async fn allocate_wallet_in_darkpool(
 
     let mut proof = dummy_valid_wallet_create_bundle();
     proof.statement.public_wallet_shares = wallet.blinded_public_shares.clone();
-    proof.statement.private_shares_commitment = share_comm;
+    proof.statement.wallet_share_commitment = share_comm;
```
akirillo, line 46: The `share_comm` here is the result of a call to `wallet.get_private_share_commitment`, we need to change that call to one which computes the full commitment

akirillo (approved):

PR #956, opened 2025-04-18
circuits: zk-circuits: wallet-update: Compute full commitment in circuit

joeykraut:
### Purpose
This PR changes the `VALID WALLET UPDATE` circuit to compute the wallet's full share commitment in circuit, as opposed to just the private share commitment.

### Performance
This PR takes the circuit size to the next power of two which increases prover time from 634ms to 1.24s. This is the same prover time as before the Poseidon arithmetization optimizations in #904, though note that we have a considerably larger margin before the next power of two in gate count.

### Testing
- [x] All tests pass

### DIFF
--- a/circuits/src/zk_circuits/valid_wallet_update.rs
+++ b/circuits/src/zk_circuits/valid_wallet_update.rs
@@ -70,10 +70,13 @@ where
         let blinder_seed = witness.old_wallet_private_shares.blinder;
         WalletGadget::validate_public_blinder_from_seed(public_blinder, blinder_seed, cs)?;
 
-        // Validate the commitment to the new wallet private shares
-        let new_wallet_private_commitment =
-            WalletGadget::compute_private_commitment(&witness.new_wallet_private_shares, cs)?;
-        cs.enforce_equal(new_wallet_private_commitment, statement.new_private_shares_commitment)?;
+        // Validate the commitment to the new wallet secret shares
+        let new_wallet_commitment = WalletGadget::compute_wallet_share_commitment(
+            &statement.new_public_shares,
+            &witness.new_wallet_private_shares,
+            cs,
+        )?;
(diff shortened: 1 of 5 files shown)


circuits/src/zk_circuits/valid_wallet_update.rs
```diff
@@ -71,9 +71,12 @@ where
         WalletGadget::validate_public_blinder_from_seed(public_blinder, blinder_seed, cs)?;
 
         // Validate the commitment to the new wallet private shares
```
akirillo, line 73: Nit: update comment

circuits/src/zk_circuits/valid_wallet_update.rs
```diff
@@ -479,7 +482,7 @@ where
     /// The nullifier of the old wallet's secret shares
     pub old_shares_nullifier: Nullifier,
     /// A commitment to the new wallet's private secret shares
```
akirillo, line 484: Nit: update comment

circuits/src/zk_circuits/valid_wallet_update.rs
```diff
@@ -671,6 +671,25 @@ mod test {
         FixedPoint { repr }
     }
 
+    /// A helper to print the number of constraints in the circuit
+    ///
+    /// Useful when benchmarking the circuit
+    #[test]
+    #[ignore]
+    fn test_n_constraints() {
```
akirillo, line 679: Do you wanna just make this a helper generic over the circuit so you don't have to redefine it in each circuit's test suite

akirillo (approved):

circuits/src/zk_circuits/valid_wallet_update.rs
joeykraut: that would basically just save the two print statements, I think it's no big deal to put one test like this in every file

PR #965, opened 2025-04-22
state: matchable amt cache

sehyunc:
### Purpose
This PR adds the matchable amount cache along with associated getters and setters.

### Testing
- [ ] Tested locally
- [ ] Test in testnet

### DIFF
--- a/state/src/caching/matchable_amount.rs
+++ b/state/src/caching/matchable_amount.rs
@@ -37,26 +37,22 @@ impl MatchableAmountMap {
     /// Get the matchable amount for a given pair and side
     pub async fn get(&self, pair: Pair, side: OrderSide) -> Amount {
         let matchable_amount_map = self.matchable_amount_map.read().await;
-        let amount = matchable_amount_map.get(&(pair, side)).unwrap_or(&0);
-        *amount
+        matchable_amount_map.get(&(pair, side)).copied().unwrap_or(0)
     }
 
     // --- Setters --- //
 
-    /// Update the matchable amount for a given pair and side
-    ///
-    /// This method handles both increasing and decreasing the cached amount.
-    /// It accounts for:
-    /// - Order creation (previous = 0)
-    /// - Order cancellation (new = 0)
-    /// - Updates where matchable amount changes
-    pub async fn update_amount(&self, pair: Pair, side: OrderSide, previous: Amount, new: Amount) {
+    /// Add to the matchable amount for a given pair and side
+    pub async fn add_amount(&self, pair: Pair, side: OrderSide, delta: Amount) {
         let mut amount_map = self.matchable_amount_map.write().await;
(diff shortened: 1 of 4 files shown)


state/src/caching/matchable_amount.rs
```diff
@@ -0,0 +1,62 @@
...
+        Self { matchable_amount_map: RwLock::new(HashMap::new()) }
+    }
+
+    // --- Getters --- //
+
+    /// Get the matchable amount for a given pair and side
+    pub async fn get(&self, pair: Pair, side: OrderSide) -> Amount {
+        let matchable_amount_map = self.matchable_amount_map.read().await;
+        let amount = matchable_amount_map.get(&(pair, side)).unwrap_or(&0);
+        *amount
```
joeykraut, line 41: nit: I'd just do
```rust
matchable_amount_map.get(&(pair, side)).copied().unwrap_or(0)
```
slightly more idiomatic

state/src/caching/matchable_amount.rs
```diff
@@ -0,0 +1,62 @@
...
+    // --- Setters --- //
+
+    /// Update the matchable amount for a given pair and side
+    ///
+    /// This method handles both increasing and decreasing the cached amount.
+    /// It accounts for:
+    /// - Order creation (previous = 0)
+    /// - Order cancellation (new = 0)
+    /// - Updates where matchable amount changes
+    pub async fn update_amount(&self, pair: Pair, side: OrderSide, previous: Amount, new: Amount) {
```
joeykraut, line 53: why not have the interface here be the delta itself? You could have `add_amount` and `sub_amount` if you're worried about signed types. That feels slightly cleaner IMO because this structure doesn't currently have any notion of an "order"; so previous and new exist at a higher level

state/src/caching/order_metadata_index.rs
```diff
@@ -67,6 +67,17 @@ impl OrderMetadataIndex {
         self.reverse_index.blocking_read().contains_key(order_id)
     }
 
+    /// Get the matchable amount for a given order_id
+    pub async fn get_matchable_amount(&self, order_id: &OrderIdentifier) -> Option<Amount> {
```
joeykraut, line 71: what will this be used for? when do we need to fetch the matchable amount for an order? All the state applicator methods can read this off the wallet right?

plus we end up scanning quite a few indices here for it, so it's not likely to be the most efficient access

joeykraut (approved):

state/src/caching/order_metadata_index.rs
sehyunc: we need to retrieve the old matchable amount to compute the delta; the order metadata index seems like the natural source for this. i've removed the getter method and instead return the old amount when updating/removing from the order metadata index since we already do this operation

PR #966, opened 2025-04-22
external-api, state, workers: add get liquidity api endpoint

sehyunc:
### Purpose
This PR adds the necessary handler and structs to expose an authenticated admin endpoint to retrieve liquidity values for a given base mint. Here is a spec of the new endpoint

### Endpoint: `GET /order_book/depth/:mint`

**Purpose:** Retrieves aggregated order book depth information for a specified token mint 

**Path Parameter:**

*   `:mint` (`string`): The address of the base token for which depth is requested.

**Response Body (`GetDepthByMintResponse`):**

*   `price` (`number`): The current market price of the base token (`:mint`) denominated in USD.
*   `timestamp` (`number`): The Unix timestamp associated with the reported `price`.
*   `buy` (`object` - `DepthSide`): Aggregated depth on the buy side.
    *   `total_quantity` (`string`): Total amount of the base token available to buy near the current `price` (serialized BigUint/Amount).
    *   `total_quantity_usd` (`number`): The approximate USD value of `total_quantity` on the buy side, calculated using the reported `price` and the token's decimals.
*   `sell` (`object` - `DepthSide`): Aggregated depth on the sell side.
    *   `total_quantity` (`string`): Total amount of the base token available to sell near the current `price` (serialized BigUint/Amount).
    *   `total_quantity_usd` (`number`): The approximate USD value of `total_quantity` on the sell side, calculated using the reported `price` and the token's decimals.

### Testing
- [x] Tested locally
- [ ] Test in testnet

### DIFF
--- a/state/src/interface/order_book.rs
+++ b/state/src/interface/order_book.rs
@@ -5,12 +5,12 @@
 //! of unconditional writes only and inconsistent state is okay between cluster
 //! peers
 
-use circuit_types::wallet::Nullifier;
+use circuit_types::{wallet::Nullifier, Amount};
 use common::types::{
     gossip::WrappedPeerId,
     network_order::NetworkOrder,
     proof_bundles::{OrderValidityProofBundle, OrderValidityWitnessBundle},
-    wallet::OrderIdentifier,
+    wallet::{OrderIdentifier, Pair},
     MatchingPoolName,
 };
 use constants::ORDER_STATE_CHANGE_TOPIC;
@@ -146,6 +146,11 @@ impl StateInner {
             Ok(orders)
         })
         .await
+    }
+
+    /// Get the matchable amount for both sides of a pair
+    pub async fn get_aggregate_matchable_amount(&self, pair: &Pair) -> (Amount, Amount) {
+        self.order_cache.get_aggregate_matchable_amount(pair).await
     }
 
     // --- Heartbeat --- //
--- a/state/src/interface/order_book.rs
+++ b/state/src/interface/order_book.rs
@@ -149,8 +149,10 @@ impl StateInner {
     }
 
     /// Get the matchable amount for both sides of a pair
-    pub async fn get_aggregate_matchable_amount(&self, pair: &Pair) -> (Amount, Amount) {
-        self.order_cache.get_aggregate_matchable_amount(pair).await
+    ///
+    /// Returns (buy_amount, sell_amount)
+    pub async fn get_liquidity_for_pair(&self, pair: &Pair) -> (Amount, Amount) {
+        self.order_cache.get_matchable_amount(pair).await
     }
 
     // --- Heartbeat --- //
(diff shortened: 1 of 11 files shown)


state/src/interface/order_book.rs
```diff
@@ -148,6 +148,11 @@ impl StateInner {
         .await
     }
 
+    /// Get the matchable amount for both sides of a pair
+    pub async fn get_aggregate_matchable_amount(&self, pair: &Pair) -> (Amount, Amount) {
```
joeykraut, line 152: nit: I'd rename this to `get_liquidity_for_pair` or something similar. The `State` object has a ton of methods implemented on it, and it can be hard to figure out which does what without pulling open the hood. So the naming on `State` has to be seen as contextualized within the broader relayer and should indicate which relayer tasks a method might be useful for.

workers/api-server/src/error.rs
```diff
@@ -70,6 +70,12 @@ impl From<AuthError> for ApiServerError {
     }
 }
 
+impl From<String> for ApiServerError {
+    fn from(value: String) -> Self {
+        internal_error(value)
+    }
+}
```
joeykraut, line 77: why not just wrap with `SomeResult.map_err(internal_error)?` if you want to do this (referencing [this method](https://github.com/renegade-fi/renegade/blob/95ea3a647ed126d386770569edd84ae5dd5585a6/workers/api-server/src/error.rs#L99))? not all strings will be internal errors

workers/api-server/src/http/admin.rs
```diff
@@ -244,6 +245,65 @@ impl TypedHandler for AdminOrderMetadataHandler {
...
+// ----------------------
+// | Liquidity Handlers |
+// ----------------------
+
+/// Handler for the GET /v0/admin/liquidity/:mint route
+pub struct AdminGetAggregateMatchableAmountHandler {
+    /// A handle to the relayer state
+    state: State,
+    /// The config for the API server
+    config: ApiServerConfig,
```
joeykraut, line 257: To stay consistent with the other handlers, just inline all config items in the handler struct. Looks like you need the price reporter queue here, so I'd just use that directly.

workers/api-server/src/http/admin.rs
```diff
@@ -244,6 +245,65 @@ impl TypedHandler for AdminOrderMetadataHandler {
...
+    ) -> Result<Self::Response, ApiServerError> {
+        let mint = parse_mint_from_params(&params)?;
+        let quote_token = Token::usdc();
+
+        // Get the price
+        let base_token = Token::from_addr_biguint(&mint);
+        let ts_price = self
+            .config
+            .price_reporter_work_queue
+            .peek_price_report(base_token.clone(), quote_token.clone())
```
joeykraut, line 287: since you set the quote to USDC manually, you can use `peek_price_usdc` here.

joeykraut (approved):
looks good, just some comments on consistency with broader relayer patterns

PR #970, opened 2025-04-23
handshake-manager: external-engine: Create malleable matches when requested

joeykraut:
### Purpose
This PR updates the external matching engine to create malleable matches when requested. This is specified by the `bounded_match` flag on the `ExternalEngineOptions` type. When set, the matching engine proceeds with the normal match flow, until the point at which a match is found and needs to be settled. The engine will then derive _bounds_ on the match and forward a `SettleMalleableExternalMatchTask` to the driver.

### Testing
- [x] Unit tests pass
- Full testing deferred until this engine config is accessible from the API

### DIFF
--- a/workers/handshake-manager/src/manager/matching/external_engine.rs
+++ b/workers/handshake-manager/src/manager/matching/external_engine.rs
@@ -12,10 +12,13 @@ use std::collections::HashSet;
 use circuit_types::{
     balance::Balance,
     fixed_point::FixedPoint,
-    r#match::{ExternalMatchResult, MatchResult},
+    r#match::{BoundedMatchResult, ExternalMatchResult, MatchResult},
 };
 use common::types::{
-    tasks::{SettleExternalMatchTaskDescriptor, TaskDescriptor},
+    tasks::{
+        SettleExternalMatchTaskDescriptor, SettleMalleableExternalMatchTaskDescriptor,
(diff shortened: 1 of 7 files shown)


workers/handshake-manager/src/manager/matching/external_engine.rs
```diff
@@ -228,8 +264,90 @@ impl HandshakeExecutor {
...
+        let order_min = order.min_fill_size;
+        let order_max = compute_max_amount(&price, &order.into(), &capitalizing_balance);
 
-        let descriptor = TaskDescriptor::from(task);
+        let min_amount = mul_amount_f64(match_amt, 1.0 - MAX_MALLEABLE_RANGE);
+        let max_amount = mul_amount_f64(match_amt, 1.0 + MAX_MALLEABLE_RANGE);
+
+        // Clamp the amounts to the order's min and max
+        let min_amount = u128::max(min_amount, order_min);
+        let max_amount = u128::min(max_amount, order_max);
```
sehyunc, line 337: Haven't looked yet but I'm assuming we have safety checks / error messages in the API layer to let a customer know when their bounds are too wide.

sehyunc: Nvm, I see now that the API layer doesn't see the change in base amount.

PR #971, opened 2025-04-23
api-server: http: external-match: Add malleable match endpoint

joeykraut:
### Purpose
This PR adds an endpoint for assembling a quote into a malleable match bundle. This largely mimics the regular assemble flow, but creates a `BoundedMatchResult` rather than an `ExternalMatchResult`.

### Todo
- Build malleable match calldata, left for a follow up.

### Testing
- [x] Unit tests pass

### DIFF
--- a/external-api/src/http/external_match.rs
+++ b/external-api/src/http/external_match.rs
@@ -9,8 +9,12 @@
 //! for consenting liquidity on a given token pair.
 
 use circuit_types::{
-    fees::FeeTake, fixed_point::FixedPoint, max_price, order::OrderSide,
-    r#match::ExternalMatchResult, Amount,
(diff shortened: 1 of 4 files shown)


sehyunc (approved):
Could exact external matches be implemented using the circuits/code paths for malleable matches (just set the bounds to allow for a single match amount)? I assume checking bounds requires more constraints than checking equality so it would be a more expensive overall...

joeykraut:
> Could exact external matches be implemented using the circuits/code paths for malleable matches (just set the bounds to allow for a single match amount)? I assume checking bounds requires more constraints than checking equality so it would be a more expensive overall...

Yes! DM'd you for more context

PR #974, opened 2025-04-23
state, workers: fix depth units

sehyunc:
### Purpose
This PR ensures `GetDepthByMintResponse` returns raw token values denominated in the base token. Previously, we were not considering the fact that `wallet.get_matchable_amount_for_order()` returns the matchable amount for a given order in terms of the order's sell asset.

### Testing
- [x] Tested locally
- [ ] Test in testnet

### DIFF
--- a/workers/api-server/src/http/order_book.rs
+++ b/workers/api-server/src/http/order_book.rs
@@ -14,6 +14,7 @@ use external_api::{
 use hyper::HeaderMap;
 use itertools::Itertools;
 use job_types::price_reporter::PriceReporterQueue;
+use num_traits::ToPrimitive;
 use state::State;
 use util::on_chain::get_external_match_fee;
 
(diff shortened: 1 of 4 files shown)


workers/api-server/src/http/order_book.rs
```diff
@@ -172,10 +174,20 @@ impl TypedHandler for GetDepthByMintHandler {
...
+        let buy_usd = quote_token.convert_to_decimal(buy_liquidity);
         let sell_usd = base_token.convert_to_decimal(sell_liquidity) * ts_price.price;
 
-        let buy = DepthSide { total_quantity: buy_liquidity, total_quantity_usd: buy_usd };
+        // Convert buy_liquidity (in terms of quote token) to be in terms of the
+        // base token
+        let decimal_corrected_price = ts_price
+            .get_decimal_corrected_price(&base_token, &quote_token)
+            .map_err(internal_error)?;
+        let price_fixed = FixedPoint::from_f64_round_down(decimal_corrected_price.price);
```
joeykraut, line 186: Use `decimal_corrected_price.as_fixed_point()` here

workers/api-server/src/http/order_book.rs
```diff
@@ -172,10 +174,20 @@ impl TypedHandler for GetDepthByMintHandler {
...
+        let buy_liquidity_base = FixedPoint::floor_div_int(buy_liquidity, price_fixed);
+        let buy_liquidity_base = scalar_to_u128(&buy_liquidity_base);
+
```
joeykraut, line 189: isn't it simpler to do all this arithmetic in f64 rather than going through fixed point? Definitely a lot faster too, and we really don't need the precision here that the matching engine does, so I don't think we need to use its types

joeykraut (approved):

PR #977, opened 2025-04-24
state: caching: avoid panic if order not found in cache

sehyunc:
### Purpose
This PR returns the pair, side, and matchable amount when removing an order form the order metadata index. This prevents a panic if an order is not found in the index. This ensures all tests pass once again.

### Testing
- [ ] Tested locally
- [ ] Test in testnet

### DIFF
--- a/state/src/caching/order_cache.rs
+++ b/state/src/caching/order_cache.rs
@@ -153,10 +153,14 @@ impl OrderBookCache {
 
     /// Remove an order from the cache entirely
     pub async fn remove_order(&self, order: OrderIdentifier) {
-        let (pair, side) = self.order_metadata_index.get_pair_and_side(&order).await.unwrap();
-        let removed = self.order_metadata_index.remove_order(&order).await.unwrap_or(0);
+        let maybe_info = self.order_metadata_index.remove_order(&order).await;
+        if maybe_info.is_none() {
+            return;
+        }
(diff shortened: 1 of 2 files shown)


state/src/caching/order_cache.rs
```diff
@@ -153,10 +153,12 @@ impl OrderBookCache {
...
-        let removed = self.order_metadata_index.remove_order(&order).await.unwrap_or(0);
-        if self.externally_enabled_orders.write().await.remove(&order) {
-            self.matchable_amount_map.sub_amount(pair, side, removed).await;
+        if let Some((pair, side, matchable_amount)) =
+            self.order_metadata_index.remove_order(&order).await
+        {
+            if self.externally_enabled_orders.write().await.remove(&order) {
+                self.matchable_amount_map.sub_amount(pair, side, matchable_amount).await;
+            }
         }
```
joeykraut, line 162: probably cleaner not to nest this:
```rust
let maybe_info = self.order_metadata_index.remove_order(&order).await;
if maybe_info.is_none() {
    return;
}
let (pair, side, matchable_amount) = maybe_info.unwrap();

if self.externally_enabled_orders.write().await.remove(&order) {
    self.matchable_amount_map.sub_amount(pair, side, matchable_amount).await;
}
```
makes the cases in which we return early clearer

joeykraut (approved):

PR #979, opened 2025-04-24
arbitrum-client: event-indexing: Recover shares from malleable matches

joeykraut:
### Purpose
This PR adds share recovery helpers to the arbitrum client that recover a wallet's shares from a malleable match transaction. We reconstruct the wallet share updates locally, as opposed to emitting an event on-chain, for simplicity.

### Testing
- [x] Unit tests pass
- [x] Tested locally, generated and submitted a malleable match on-chain, verified that the relayer correctly recovered my wallet.

### DIFF
--- a/arbitrum-client/src/conversion.rs
+++ b/arbitrum-client/src/conversion.rs
@@ -228,6 +228,18 @@ pub fn to_contract_order_settlement_indices(
     }
 }
 
+/// Convert a contract [`OrderSettlementIndices`] to a
+/// [`OrderSettlementIndices`]
+pub fn to_circuit_order_settlement_indices(
+    indices: &ContractOrderSettlementIndices,
+) -> OrderSettlementIndices {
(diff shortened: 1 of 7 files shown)


arbitrum-client/src/conversion.rs
```diff
@@ -284,6 +296,26 @@ pub fn to_contract_bounded_match_result(
     })
 }
 
+/// Convert a contract [`BoundedMatchResult`] to a [`BoundedMatchResult`]
+pub fn to_circuit_bounded_match_result(
+    match_result: &ContractBoundedMatchResult,
+) -> Result<BoundedMatchResult, ConversionError> {
+    let quote_mint = address_to_biguint(&match_result.base_mint)?;
+    let base_mint = address_to_biguint(&match_result.quote_mint)?;
```
sehyunc, line 304: Should be switched

sehyunc (approved):

PR #983, opened 2025-04-29
arbitrum-client: Rename to `darkpool-client`

joeykraut:
### Purpose
This PR renames the `arbitrum-client` crate to `darkpool-client` and generalizes all relevant naming across the relayer.

### Testing
- [x] All unit tests pass


test-helpers/Cargo.toml
```diff
@@ -6,12 +6,12 @@ description = "Helpers for setting up and running integration tests for the rela
 
 [features]
 default = ["test-harness"]
-contracts = ["arbitrum"]
-arbitrum = [
+arbitrum = ["blockchain"]
```
akirillo, line 9: we use this feature to import test helpers in the stylus contract repo, right? if so I'd be in favor of more directly calling the feature `stylus`, as the test helpers imported are no longer specific to arbitrum, as the current name of the feature implies

akirillo (approved):

test-helpers/Cargo.toml
joeykraut: those particular test helpers are specific to arbitrum, the ones under `blockchain` are not

PR #988, opened 2025-05-01
darkpool-client: base: Add base implementation

joeykraut:
### Purpose
This PR begins the Base implementation of the darkpool client, implementing the first few setter methods.

### Todo
- Implement the rest of the client

### Testing
- [x] Relayer builds with `--features "base"`
- [x] Unit tests pass

### DIFF
--- a/darkpool-client/src/base/mod.rs
+++ b/darkpool-client/src/base/mod.rs
@@ -1,0 +1,5 @@
+//! The Base implementation of the darkpool client
+
+/// The Base darkpool implementation
+#[derive(Clone)]
+pub struct BaseDarkpool {}
--- a/darkpool-client/src/base/mod.rs
+++ b/darkpool-client/src/base/mod.rs
@@ -1,5 +1,330 @@
 //! The Base implementation of the darkpool client
+mod conversion;
+
+use alloy::rpc::types::{TransactionReceipt, TransactionRequest};
+use conversion::to_contract_type;
+use renegade_solidity_abi::IDarkpool::{
+    IDarkpoolInstance, MatchLinkingProofs, MatchProofs, MerkleInsertion as AbiMerkleInsertion,
+    MerkleOpeningNode as AbiMerkleOpeningNode, NullifierSpent as AbiNullifierSpent,
+    WalletUpdated as AbiWalletUpdated,
+};
+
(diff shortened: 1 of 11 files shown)


darkpool-client/src/base/mod.rs
```diff
@@ -0,0 +1,330 @@
...
+        self.send_tx(call).await
+    }
+
+    async fn update_wallet(
+        &self,
+        valid_wallet_update: &SizedValidWalletUpdateBundle,
+        wallet_commitment_signature: Vec<u8>,
+        transfer_auth: Option<TransferAuth>,
+    ) -> Result<TransactionReceipt, DarkpoolClientError> {
+        // TODO: re-serialize the signature using the ABI encoding
```
akirillo, line 170: is this necessary? I would imagine passing in the `Bytes` struct as we do ensures that it is properly encoded as a solidity `bytes` type

darkpool-client/src/lib.rs
```diff
@@ -28,3 +32,14 @@ pub type DarkpoolClient = client::DarkpoolClientInner<arbitrum::ArbitrumDarkpool
...
+#[cfg(feature = "base")]
+pub mod base;
+#[cfg(feature = "base")]
+/// The darkpool client for the Base chain
+pub type DarkpoolClient = client::DarkpoolClientInner<base::BaseDarkpool>;
+#[cfg(feature = "base")]
+/// The darkpool implementation for the Base chain
+///
+/// Exported here to allow lower level access from other workers
+pub type DarkpoolImplementation = base::BaseDarkpool;
```
akirillo, line 45: clippy complaining due to multiple `DarkpoolClient` implementations, which must be because it's enabling all features. of course these are red herrings but it will make CI annoying, can we ignore the lint if both `arbitrum` and `base` are set?

akirillo (approved):

darkpool-client/src/lib.rs
joeykraut: yeah i'll figure out how to do this in a follow up

darkpool-client/src/base/mod.rs
joeykraut: yes the solidity contracts expect a different solidity friendly encoding

PR #990, opened 2025-05-01
dakrpool-client: base: conversion: Make all conversions trait based

joeykraut:
### Purpose
This PR defines the `ToContractType` and `ToCircuitType` traits, which encapsulate conversion logic between circuit and contract types.

### Todo
- Share recovery

### Testing
- [x] Unit tests pass

### DIFF
--- a/darkpool-client/src/base/mod.rs
+++ b/darkpool-client/src/base/mod.rs
@@ -1,15 +1,8 @@
 //! The Base implementation of the darkpool client
 mod conversion;
+mod helpers;
 
 use alloy::rpc::types::{TransactionReceipt, TransactionRequest};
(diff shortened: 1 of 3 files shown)


darkpool-client/src/base/mod.rs
```diff
@@ -1,15 +1,8 @@
 //! The Base implementation of the darkpool client
 mod conversion;
+mod helpers;
```
akirillo, line 3: dead code? or implemented in a follow up

akirillo (approved):

darkpool-client/src/base/mod.rs
joeykraut: follow up

PR #993, opened 2025-05-02
node-support: bootloader: check for p2p key in config

akirillo:
This PR tweaks the bootloader script to check if there is already a `p2p-key` set in the config file before generating a random one.

### Testing
- [x] Tested p2p key decoding logic locally
- [ ] Test running testnet bootstraps

### DIFF
--- a/node-support/bootloader/src/main.rs
+++ b/node-support/bootloader/src/main.rs
@@ -162,7 +162,7 @@ async fn modify_config(whitelist_path: Option<String>) -> Result<(), String> {
         toml::from_str(&config_content).map_err(raw_err_str!("Failed to parse config: {}"))?;
 
     // Setup the peer id and register a gas wallet under this peer id
-    let peer_id = set_p2p_key(&mut config);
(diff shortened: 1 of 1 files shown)


node-support/bootloader/src/main.rs
```diff
@@ -162,7 +162,20 @@ async fn modify_config(whitelist_path: Option<String>) -> Result<(), String> {
...
+                .decode(key_base64)
+                .map_err(raw_err_str!("Failed to decode p2p key from base64: {}"))?;
+
+            let keypair = Keypair::from_protobuf_encoding(&key_bytes)
+                .map_err(raw_err_str!("Failed to decode p2p key from protobuf encoding: {}"))?;
+
+            keypair.public().to_peer_id()
+        },
+        None => set_p2p_key(&mut config),
+    };
```
joeykraut, line 178: it'd be cleaner to factor this logic into `set_p2p_key` using a `get_p2p_key(config: &HashMap<String, Value) -> Option<PeerId>`. This method already does a lot

joeykraut (approved):

PR #994, opened 2025-05-05
darkpool-client: client: Respect `polling_interval` config param

joeykraut:
### Purpose
This PR makes two changes to the `DarkpoolClient`:
- Use the `SimpleNonceManager` which re-fetches the current nonce on every call. We do this to prevent the nonce getting out of sync. I will follow up with a PR that uses a custom cached nonce manager, the alloy `CachedNonceManager` does not have an invalidation mechanism.
- Set the client's polling interval to the configured value.

### Testing
- [x] Tested in testnet


akirillo (approved):
lgtm but clippy error in `task-driver/integration` using the now-removed `block_polling_interval_ms` field on the `DarkpoolClientConfig`

PR #996, opened 2025-05-07
circuits: zk-gadgets: comparators: Fix `greater_than_eq_zero`

joeykraut:
### Purpose
This PR fixed the `greater_than_eq_zero` gadget in the `comparators` to fix a soundness issue when the gadget returns false. The updated implementation follows the [circomlib implementation](https://github.com/iden3/circomlib/blob/35e54ea21da3e8762557234298dbb553c175ea8d/circuits/comparators.circom#L89-L99) as suggested in the audit.

### Testing
- [x] All unit tests pass

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15,7 +15,7 @@ dependencies = [
 [[package]]
 name = "abi"
 version = "0.1.0"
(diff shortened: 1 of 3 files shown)


akirillo (approved):
Neat, appreciate the explanation

mitschabaude (approved):

PR #998, opened 2025-05-09
common: token: allow for multiple token remaps

sehyunc:
### Purpose
This PR extends the Token abstraction to be aware of Chain, effectively allowing for multiple token remaps to be loaded in one process. 

### Testing
- [ ] Tested locally
- [ ] Test in testnet


common/src/types/chain_token.rs
```diff
@@ -0,0 +1,64 @@
...
+    /// Token
+    pub fn from_addr_biguint_on_chain(addr: &BigUint, chain: Chain) -> Self {
+        Self { addr: biguint_to_hex_addr(addr).to_lowercase(), chain }
+    }
+
+    /// Given an ERC-20 contract address, returns a new Token
+    ///
+    /// Returns the first matching address found across all chains. Panics if
+    /// no matching address is found.
+    pub fn from_addr_without_chain(addr: &str) -> Self {
```
sehyunc, line 28: Sometimes (like in the price reporter) given an address one must find the chain it belongs to across all configured token remaps.

sehyunc:
@joeykraut how might I bypass "the name `DarkpoolClient` is defined multiple times" so I can run tests?

common/src/types/chain_token.rs
```diff
@@ -0,0 +1,61 @@
...
+use num_bigint::BigUint;
+use util::hex::biguint_to_hex_addr;
+
+use super::{
+    chain::Chain,
+    exchange::Exchange,
+    token::{read_token_remaps, Token, USDT_TICKER, USD_TICKER},
+};
+
+impl Token {
```
joeykraut, line 11: it'd be better to co-locate this with the `Token` struct. Except for super complicated and wide reaching structs (e.g. `State`), it's better to place implementations below the struct definition

common/src/types/token.rs
```diff
@@ -261,3 +277,13 @@ pub fn default_exchange_stable(exchange: &Exchange) -> Token {
...
+
+/// Returns the configured chain if exactly one is present
+pub fn default_chain() -> Option<Chain> {
+    let remaps = read_token_remaps();
+    if remaps.len() == 1 {
+        remaps.keys().next().copied()
+    } else {
+        None
+    }
+}
```
joeykraut, line 289: we call `expect` everywhere on this method. I'd just have it return `Chain` and internalize the panic to this method

joeykraut: As well, why not just change the default to be stored in a static `DEFAULT_CHAIN`; and add getter methods for it? We can just set it to the first configured remap if you like.

common/src/types/token.rs
```diff
@@ -104,23 +106,30 @@ impl Token {
 
     /// Given an ERC-20 contract address, returns a new Token
     pub fn from_addr(addr: &str) -> Self {
-        Self { addr: String::from(addr).to_lowercase() }
+        let chain =
+            default_chain().expect("Multiple chains configured; use from_addr_on_chain instead.");
+        Self { addr: String::from(addr).to_lowercase(), chain }
```
joeykraut, line 111: I'd create a `new` method like:
```rust
pub fn new(addr: &str, chain: Chain) -> Self {
    Self { addr: addr.to_string().to_lowercase(), chain }
}
```

reason being that it's quite important we lowercase the address, and centralizing that assumption to a constructor would help.

joeykraut: While you're at it, can you override the derived `Eq` implementation to do lower-cased comparison of the addresses?

common/src/types/token.rs
```diff
@@ -142,16 +151,18 @@ impl Token {
     /// Tickers do not have any ERC-20 ticker, as we support long-tail
     /// assets.
     pub fn get_ticker(&self) -> Option<String> {
-        let token_remap = read_token_remap();
+        let all_remaps = read_token_remaps();
+        let token_remap = all_remaps.get(&self.chain).expect("Chain has not been setup");
```
joeykraut, line 155: do we need to expect here? We can just do `all_remaps.get(&self.chain).ok()?;`

common/src/types/token.rs
```diff
@@ -61,14 +62,13 @@ pub const USD_TICKER: &str = "USD";
 /// be invoked if they are the quote
 pub const STABLECOIN_TICKERS: &[&str] = &[USDC_TICKER, USDT_TICKER];
 
-/// The token remapping for the given environment, maps from the token address
-/// to the ticker of the token
-pub static TOKEN_REMAPS: RwStatic<BiMap<String, String>> =
-    RwStatic::new(|| RwLock::new(BiMap::new()));
+/// Maps a chain to a mapping from the token address to the ticker of the token
+pub static TOKEN_REMAPS_BY_CHAIN: RwStatic<HashMap<Chain, BiMap<String, String>>> =
+    RwStatic::new(|| RwLock::new(HashMap::new()));
```
joeykraut, line 67: These map types are quite complicated now, I'd add a type alias for them or just place them in a `TokenRemap` struct.

common/src/types/chain_token.rs
```diff
@@ -0,0 +1,61 @@
...
+    /// Given an ERC-20 contract address, returns a new Token, if available.
+    pub fn from_addr_without_chain(addr: &str) -> Option<Self> {
+        let addr = String::from(addr).to_lowercase();
+        for (chain, token_map) in read_token_remaps().iter() {
+            if token_map.get_by_left(&addr).is_some() {
+                return Some(Self { addr, chain: *chain });
+            }
+        }
+        None
+    }
```
joeykraut, line 33: The delineation between `from_addr_on_chain`, `from_addr_without_chain`, and `from_addr` is confusing. Do we need `from_addr_without_chain`?

common/src/types/chain_token.rs
```diff
@@ -0,0 +1,61 @@
...
+
+    /// Given an ERC-20 ticker, returns a new Token on a specific chain.
+    pub fn from_ticker_on_chain(ticker: &str, chain: Chain) -> Self {
+        let all_maps = read_token_remaps();
+        let token_map = all_maps.get(&chain).expect("Chain has not been setup");
+        let addr = token_map.get_by_right(ticker).expect("Ticker could not be found on chain; try a different chain or specify unnamed token by ERC-20 address using from_addr instead.");
```
joeykraut, line 39: I would use a shorter expect message here, this isn't user facing so it can be short. + this line is very long

common/src/types/chain_token.rs
```diff
@@ -0,0 +1,61 @@
...
+        Exchange::Binance => Token::from_ticker_on_chain(USDT_TICKER, chain),
+        Exchange::Coinbase => Token::from_ticker_on_chain(USD_TICKER, chain),
+        Exchange::Kraken => Token::from_ticker_on_chain(USD_TICKER, chain),
+        Exchange::Okx => Token::from_ticker_on_chain(USDT_TICKER, chain),
+        _ => panic!(
+            "No default stable quote asset for exchange: {:?} on chain: {:?}",
+            exchange, chain
+        ),
+    }
+}
```
joeykraut, line 61: Do we need this method? feels like it should sit in the price reporter. Especially because it doesn't do anything with the chain argument that the relayer would unique be able to do, and it's only relevant for price streaming

workers/price-reporter/src/mock.rs
```diff
@@ -41,7 +42,10 @@ const MOCK_TICKER_NAMES: &[&str] = &[
 #[allow(unused_must_use)]
 pub fn setup_mock_token_remap() {
     // Setup the mock token map
-    let mut token_map = write_token_remap();
+    let mut all_maps = write_token_remaps();
+    let chain = Chain::Mainnet;
+    let token_map = all_maps.entry(chain).or_default();
+    token_map.clear();
```
joeykraut, line 48: why is this necessary?

common/src/types/chain_token.rs
sehyunc: I've moved it into a helper function in the price reporter 👍🏻

common/src/types/token.rs
```diff
@@ -267,3 +339,19 @@ pub fn default_exchange_stable(exchange: &Exchange) -> Token {
...
+        *guard = Some(chain);
+    }
+}
+
+/// Return the default Chain, or panic if none has ever been set.
+pub fn default_chain() -> Chain {
+    let guard: RwLockReadGuard<'_, Option<Chain>> =
+        DEFAULT_CHAIN.read().expect("DEFAULT_CHAIN lock poisoned");
+    guard.expect("no default chain configured")
+}
```
joeykraut, line 357: see my suggestions above about a `OnceLock`, it would implement this same behavior wherein we can only set the value once. The advantage is that a `OnceLock` will handle the behavior you're implementing here directly

joeykraut (approved):

PR #1005, opened 2025-05-14
common: token: Fall back to singleton map if default not configured

joeykraut:
### Purpose
This PR changes the behavior of `default_chain` to return the only configured chain in the case that only one chain is configured an `DEFAULT_CHAIN` is not set. This obviates the need to set the static `DEFAULT_CHAIN` in a single-chain implementation.

### Testing
- [x] All tests pass


sehyunc (approved):
lgtm; for my own curiosity, what was the context in which this cropped up

joeykraut:
> lgtm; for my own curiosity, what was the context in which this cropped up

In the health check lambda I'm building. I don't use the setup remaps method

PR #1007, opened 2025-05-15
node-support: bootloader: Use chain-dependent gas wallet URL

joeykraut:
### Purpose
This PR makes a number of changes to the relayer's bootloader:
1. Use the chain-dependent funds manager URL for registering a gas wallet
2. Internalize the funds manager API types, to prevent a cyclic dependency with `funds-manager-api`
3. Remove fee whitelisting -- not currently used
4. Refactor individual components of the bootloader into their own files

### Testing
- [x] Deployed to `base-sepolia-relayer`

### DIFF
--- a/node-support/bootloader/src/gas_wallet.rs
+++ b/node-support/bootloader/src/gas_wallet.rs
@@ -1,0 +1,119 @@
+//! Helpers for setting up a gas wallet for the relayer
+
+use common::types::hmac::HmacKey;
+use external_api::auth::add_expiring_auth_to_headers;
+use reqwest::{
+    header::{HeaderMap, HeaderValue, CONTENT_TYPE},
+    Client,
+};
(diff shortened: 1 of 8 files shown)


node-support/bootloader/src/gas_wallet.rs
```diff
@@ -0,0 +1,119 @@
...
+use crate::helpers::read_env_var;
+
+// -------------
+// | Constants |
+// -------------
+
+/// The chain env environment variable
+///
+/// This is a value like `arbitrum-sepolia`
+const ENV_CHAIN_ENV: &str = "CHAIN_ENV";
```
akirillo, line 25: we don't need this as a new env var, the `CHAIN` env var will be exactly this value and is already passed in to the relayer container

akirillo (approved):
one nit, o/w lgtm

PR #1009, opened 2025-05-16
node-support: bootloader: gas_wallet: use old (bkwds-compat) register-gas-wallet route

akirillo:
This PR patches the `bootloader` to invoke the old `/register-gas-wallet` route on the funds manager. The purpose of this is to decouple relayer deployments from funds manager deployments (as the funds manager relies on newer versions of the relayer). This old form of the route is also supported on the pending new version of the funds manager.


akirillo:
closing as this is a temporary patch

PR #1013, opened 2025-05-17
darkpool-client: event-indexing: Use more efficient block scan

joeykraut:
### Purpose
This PR changes the `DarkpoolClient` to use a more efficient block scan range when querying events by topic. Specifically, we have an exponentially widening window on block range until an event is found. This allows for recent events to be found quickly while still covering all events since the deploy block.

This is done to mitigate the timeout issues with large block range queries from Alchemy.

### Testing
- [x] Tested locally, replicated the Alchemy timeout bug and verified that this fixed it
- [ ] Testing in testnet

### DIFF
--- a/darkpool-client/src/client/event_indexing.rs
+++ b/darkpool-client/src/client/event_indexing.rs
@@ -10,7 +10,9 @@ use alloy::providers::{ext::DebugApi, Provider};
 use alloy::rpc::types::trace::geth::{
     CallFrame, GethDebugBuiltInTracerType, GethDebugTracerType, GethDebugTracingOptions, GethTrace,
 };
+use alloy::rpc::types::Log as RpcLog;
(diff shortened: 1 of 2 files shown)


sehyunc (approved):
lgtm

i don't know the # of logs off the top of my head but I'm assuming this works fine with Alchemy's safety nets regarding log retrieval: "a cap of 10K logs in the response OR a 2K block range with no cap on logs in the response and 150MB limit on the response size"

joeykraut:
> lgtm
> 
> i don't know the # of logs off the top of my head but I'm assuming this works fine with Alchemy's safety nets regarding log retrieval: "a cap of 10K logs in the response OR a 2K block range with no cap on logs in the response and 150MB limit on the response size"

These should all be querying one log. The original query works fine it just times out, it seems something changed on their end undocumented

PR #1015, opened 2025-05-19
workers: event_manager: generalize metadata fields on `RelayerEvent`

akirillo:
This PR makes a *breaking* change to the `RelayerEvent` type, turning it into a struct that holds the common fields `event_id`, `event_timestamp`, and `chain` (newly-added), along with an enum field `RelayerEventType` which contains the underlying event variant.

This will require any consumers of the exported events to be updated first, with shim logic for parsing the old serialization of the `RelayerEvent` (and a default setting for the `chain` parameter)

### Testing
- [x] All unit tests pass


workers/job-types/src/event_manager.rs
```diff
@@ -146,6 +161,8 @@ pub struct WalletCreationEvent {
     pub event_id: Uuid,
     /// The time at which the event occurred
     pub event_timestamp: SystemTime,
+    /// The chain on which the wallet was created
+    pub chain: Chain,
```
joeykraut, line 165: Do we use these types in the historical state engine? We may want or need default values here. You could also just deploy the HSE before the relayer, but just calling this out as breaking.

joeykraut (commented):
This could be done in a much simpler way by attaching the chain in the event export sidecar. Let's chat about that approach before I approve

joeykraut (approved):

PR #1022, opened 2025-06-07
circuit-types: fixed-point: Optimize `to_f64` method

joeykraut:
### Purpose
This PR removes some of the unnecessary conversions in the `FixedPoint::to_f64` method and leans on `num_traits::ToPrimitive` for the bulk of the conversion. This results in a 40% improvement in matching engine latency.

### Testing
- [x] All tests pass
- [x] Added a benchmark to measure latency improvement


sehyunc (approved):
very cool, two questions: 
- is it purely the range checks on price that call `FixedPoint::to_64`?
- would we benefit from similar changes to `FixedPoint::from_f64_round_down`? my thinking is no since it doesn't seem like we convert f64 to FixedPoint too often in any hot paths.

joeykraut:
> very cool, two questions:
> 
> * is it purely the range checks on price that call `FixedPoint::to_64`?
> * would we benefit from similar changes to `FixedPoint::from_f64_round_down`? my thinking is no since it doesn't seem like we convert f64 to FixedPoint too often in any hot paths.

1. Yes that's exactly the bottleneck

PR #1027, opened 2025-06-10
state: benches: caching: Add matchable amount caching benchmark

joeykraut:
### Purpose
This PR adds a benchmark for the open matchable amount cache in the state.

### Testing
- [x] Benchmarks run correctly

### DIFF
--- a/state/Cargo.toml
+++ b/state/Cargo.toml
@@ -23,6 +23,11 @@ name = "interface"
 harness = false
 required-features = ["mocks"]
 
+[[bench]]
+name = "caching"
+harness = false
+required-features = ["mocks"]
+
 [dependencies]
 
 # === Replication === #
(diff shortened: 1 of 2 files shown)


joeykraut:
> Looks good to me, a few questions:
> 
> * Is a benchmark including PriceReporter peek_price jobs coming in a future PR, or have you already rejected that as a potential source of latency?
> * Seems to me this benchmark is simply measuring happy-path of my cache implementation, is that right?
> * Did this confirm/reject any hypotheses? When I ran it I saw <200ns per operation, which seems fast.

It seems to me that the price reporter job task queue accounts for most of the latency. This benchmark definitely helped confirm that alongside the pprof dump you sent me. Even so, the benchmark is interesting to keep around.

Benchmarking the price reporter job doesn't seem all that fruitful, the profile you sent me seems to see contention on the send side of the job queue, which indicates to me we're due for a refactor of how prices are disseminated throughout the relayer. Not much we can do about job queue latency, nor learn about it via a benchmark.

PR #1028, opened 2025-06-10
api-server: http: order-book: Add depth for all pairs endpoint

joeykraut:
### Purpose
This PR adds an endpoint to fetch the price and book depth for all supported pairs.

### Testing
- [x] Unit tests pass
- [x] Tested locally

### DIFF
--- a/workers/api-server/src/http/order_book.rs
+++ b/workers/api-server/src/http/order_book.rs
@@ -5,8 +5,8 @@ use common::types::{token::Token, wallet::pair_from_mints};
 use constants::EXTERNAL_MATCH_RELAYER_FEE;
 use external_api::{
     http::order_book::{
-        GetDepthByMintResponse, GetExternalMatchFeeResponse, GetNetworkOrderByIdResponse,
-        GetNetworkOrdersResponse,
+        GetDepthByMintResponse, GetDepthForAllPairsResponse, GetExternalMatchFeeResponse,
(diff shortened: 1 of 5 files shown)


workers/api-server/src/http/order_book.rs
```diff
@@ -131,67 +136,143 @@ impl TypedHandler for GetExternalMatchFeesHandler {
...
+        _params: UrlParams,
+        _query_params: QueryParams,
+    ) -> Result<Self::Response, ApiServerError> {
+        // Get all tokens for which we support price data
+        // Practically, this is all non-stablecoin tokens
+        let supported_tokens = get_all_tokens_filtered(&FILTERED_TOKENS_PRICES);
+        let quote_token = Token::usdc();
+
+        let mut pairs = Vec::new();
+        for token in supported_tokens {
```
sehyunc, line 263: Could we `get_price_and_depth` in parallel?

sehyunc (approved):

workers/api-server/src/http/order_book.rs
joeykraut: we could, though it's quite fast as is, I don't see much need to until latency is high. If you disagree i'm happy to add it

sehyunc: Agree with deferring until we see the need

PR #1034, opened 2025-06-16
workers: handshake-manager: Add min_quote_amount to external engine

joeykraut:
### Purpose
This PR:
1. Adds a `min_quote_amount` field to the `ExternalMatchingEngineOptions` struct, which will control the minimum matchable quote amount using existing paths.
2. When an external user passes a quote-denominated order to the API server, the API server will respect `min_fill_size` by forwarding this value to the external matching engine.
3. This PR also cleans up some of the logic for setting the `ExternalMatchingEngineOptions` in the api server, specifically I factored out the large conversion method into smaller helpers.
4. Removed an extra clone in the matching engine, by allowing callers to pass iterators of orders rather than objects which implement `IntoIterator<Item = OrderIdentifier>`.

### Testing
- [x] Testing locally
- [x] Testing in testnet
- [x] Unit tests pass

### DIFF
--- a/workers/handshake-manager/src/manager/matching/match_helpers.rs
+++ b/workers/handshake-manager/src/manager/matching/match_helpers.rs
@@ -17,7 +17,7 @@ pub type SuccessfulMatch = (OrderIdentifier, MatchResult);
 
 impl HandshakeExecutor {
     /// Run a match between two orders
-    pub async fn find_match<I>(
+    pub async fn find_match<'a, I>(
         &self,
         order: &Order,
         balance: &Balance,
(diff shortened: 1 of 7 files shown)


workers/handshake-manager/src/manager/matching/match_helpers.rs
```diff
@@ -17,29 +17,45 @@ pub type SuccessfulMatch = (OrderIdentifier, MatchResult);
...
+    pub async fn find_match<'a, I>(
         &self,
         order: &Order,
         balance: &Balance,
         price: FixedPoint,
         target_orders: I,
     ) -> Result<Option<SuccessfulMatch>, HandshakeManagerError>
     where
-        I: IntoIterator<Item = OrderIdentifier>,
+        I: Iterator<Item = &'a OrderIdentifier>,
```
joeykraut, line 28: Changed this trait bound to remove clones in callers

akirillo (approved):

PR #1035, opened 2025-06-17
workers: api-server: integration: Add integraiton tests

joeykraut:
### Purpose
This PR adds the integration test setup for the `api-server` using the existing `test-helpers` harness and the `mock-node` implementation. 

I'll add tests for the external match API in a follow-up.

### Testing
- [x] Ping test passes


workers/api-server/Cargo.toml
```diff
@@ -3,6 +3,15 @@ name = "api-server"
...
 edition = "2021"
 
+[[test]]
+name = "integration"
+path = "integration/main.rs"
+harness = false
+required-features = ["test_helpers"]
+
+[features]
+test_helpers = ["darkpool-client/arbitrum"]
```
akirillo, line 13: Why not let `arbitrum` or `base` feature flags be set at the time of running the tests?

akirillo (approved):

PR #1036, opened 2025-06-17
api-server: integration: Add basic tests

joeykraut:
### Purpose
This PR adds basic helpers and tests to the API server's integration test suite. The build is failing at the moment because the `hyper` version needs to be upgraded. I'll do this in a follow up. I've left some loose ends in this implementation, so I'll complete them in a follow-up as well.

### Testing
- [x] Existing tests pass


common/src/types/token.rs
```diff
@@ -260,11 +261,21 @@ impl Token {
...
+    pub fn convert_to_decimal(&self, amount: Amount) -> f64 {
         let decimals = self.get_decimals().unwrap_or_default();
         let decimal_adjustment = 10u128.pow(decimals as u32);
         amount as f64 / decimal_adjustment as f64
     }
+
+    /// Converts the amount of the token as an f64 to an `Amount`, accounting
+    /// for the associated number of decimals.
+    ///
+    /// Note that due to conversion to f64, the result may lose precision.
```
akirillo, line 273: :ackchyually:
```suggestion
    /// Note that due to conversion from f64, the result may lose precision.
```

workers/api-server/integration/ctx/external_match.rs
```diff
@@ -0,0 +1,36 @@
...
+        let body_bytes = serde_json::to_vec(&req).expect("failed to serialize request");
+        add_expiring_auth_to_headers(
+            &path,
+            &mut headers,
+            &body_bytes,
+            &self.admin_api_key,
+            REQUEST_AUTH_DURATION,
+        );
+
+        self.send_req_with_headers(REQUEST_EXTERNAL_QUOTE_ROUTE, Method::POST, headers, req).await
```
akirillo, line 34: will probably prove useful to package this into a `send_admin_req` helper, either on the `MockNode` or on `IntegrationTestCtx`

workers/api-server/integration/ctx/wallet_setup.rs
```diff
@@ -0,0 +1,62 @@
...
+
+        // Add the wallet to the state
+        let waiter = self.mock_node.state().update_wallet(wallet).await?;
+        waiter.await.to_eyre().map(|_| ())
+    }
+
+    /// Build an order to cross with the given order
+    fn build_matching_order(&self, order: &ExternalOrder) -> Result<Order> {
+        let matching_side = order.side.opposite();
+        let worst_cast_price = match matching_side {
```
akirillo, line 39: nit
```suggestion
        let worst_case_price = match matching_side {
```

workers/api-server/integration/external_match/basic_match.rs
```diff
@@ -0,0 +1,54 @@
...
+    // Use a non-standard quote and base mint so that the order is not matched
+    // with an order created by another test
+    let base_mint = Token::from_ticker("UNI").get_addr_biguint();
+    let quote_mint = Token::from_ticker("USDC").get_addr_biguint();
+
+    let external_order = ExternalOrder {
+        base_mint,
+        quote_mint,
+        side: OrderSide::Buy,
+        quote_amount: 0,
```
akirillo, line 24: won't this result in an "order size below min" error before failing to find a quote?

akirillo (approved):

workers/api-server/integration/external_match/basic_match.rs
joeykraut: this is one of the loose ends I'm still tying up

workers/api-server/integration/ctx/external_match.rs
joeykraut: if it does, I will

PR #1048, opened 2025-06-19
task-driver: tasks: Lookup Merkle path directly from tx receipt

joeykraut:
### Purpose
This PR fetches the Merkle authentication using the transaction receipt returned by the transaction submission directly, where possible. This will avoid the failures we see when trying to find a commitment immediately after submitting a tx.

### Testing
- [x] Unit tests pass
- [x] Testing in testnet

### DIFF
--- a/workers/task-driver/src/tasks/settle_match.rs
+++ b/workers/task-driver/src/tasks/settle_match.rs
@@ -8,6 +8,7 @@
 use std::error::Error;
 use std::fmt::{Display, Formatter, Result as FmtResult};
 
+use alloy::rpc::types::TransactionReceipt;
 use ark_mpc::PARTY0;
 use async_trait::async_trait;
 use circuit_types::r#match::MatchResult;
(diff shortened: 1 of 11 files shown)


workers/task-driver/src/tasks/settle_match.rs
```diff
@@ -322,7 +328,11 @@ impl SettleMatchTask {
             .await?;
 
         // Find the wallet's new Merkle opening
-        let opening = find_merkle_path(&wallet, &self.darkpool_client).await?;
+        let opening = if let Some(tx) = self.tx.as_ref() {
```
sehyunc, line 331: We do not `unwrap` here because of the explanation on line 286: "If the transaction failed because a nullifier was already used..."?

sehyunc (approved):

workers/task-driver/src/tasks/settle_match.rs
joeykraut: Yeah correct, this is the MPC match path, so the other party in the MPC may have submitted the match. In this case, the tx may have failed, but the match went through. So we should fetch events using the existing merkle method.

PR #1052, opened 2025-06-23
handshake-manager: matching: Hold prices as fixed point values

joeykraut:
### Purpose
The conversion to and from `f64` sometimes causes prices to violate min fill size constraints when rounding down. To solve this, we hold all prices as `FixedPoint` values throughout matching and only convert to `f64` for display or metrics.

### Testing
- [x] All integration and unit tests pass
- [ ] Testing prover errors in testnet


akirillo (approved):
nice

PR #1053, opened 2025-06-24
external-api, handshake-manager: Add matching pool param to external match API

joeykraut:
### Purpose
This PR adds a param for the matching pool in all external match requests. If no pool is specified we match against all pools, including the global pool. If a pool is specified we match only against orders in that pool.

I added two integrations tests for:
- No pool specified -> should match against an order allocated in a non-global pool
- Pool specified -> should only match against orders in the given pool

### Testing
- [x] Integration tests pass
- [x] Unit tests pass


akirillo (approved):
very clean

PR #1060, opened 2025-07-05
state: storage: error: Allow storage layer to reject state transitions

joeykraut:
### Purpose
This PR replaces `StorageError::InvalidWrite` with `StorageError::Reject` and converts a storage level rejection into a state applicator rejection. This allows the storage layer to reject a state applicator's state transition directly rather than signalling through a subset of its error types.

### Testing
- [x] All unit tests pass

### DIFF
--- a/state/src/applicator/error.rs
+++ b/state/src/applicator/error.rs
@@ -48,6 +48,9 @@ impl Error for StateApplicatorError {}
 
 impl From<StorageError> for StateApplicatorError {
     fn from(value: StorageError) -> Self {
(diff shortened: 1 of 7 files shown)


sehyunc (approved):
very clean. are there no other instances of incorrect `StorageError::other` other than the two in `check_invariants`

joeykraut:
> very clean. are there no other instances of incorrect `StorageError::other` other than the two in `check_invariants`

changed to `reject` in a few other places, thanks for checking

PR #1063, opened 2025-07-11
workers: chain-events: record fill on external match settlement

sehyunc:
### Purpose
This PR adds logic to record a fill on the internal order when external match settlement is observed. This ensures the historical state engine picks up fill events on external orders. 

### Testing
- [ ] Tested locally
- [ ] Test in testnet


workers/chain-events/src/listener.rs
```diff
@@ -336,19 +349,84 @@ impl OnChainEventListenerExecutor {
...
                 &[wallet_id],
             );
+            // Emit relayer event for the fill
+            self.emit_external_fill(wallet_id, external_match_result.clone()).await?;
         }
 
         Ok(external_match)
     }
+
+    // -----------
```
joeykraut, line 366: nit: i'd re-order these to read top to bottom in terms of the call stack

workers/chain-events/src/listener.rs
```diff
@@ -336,19 +349,84 @@ impl OnChainEventListenerExecutor {
...
+    // | Helpers |
+    // -----------
+
+    /// Find the internal order in the given wallet that matches the external
+    /// match result
```
joeykraut, line 371: can you note here that this will just return the first matching order? It's technically possible for a wallet to have two orders matching the criteria which this method checks; though not practically all that likely

joeykraut (approved):
super clean

PR #1064, opened 2025-07-11
workers: chain-events: update order metadata on settlement

sehyunc:
### Purpose
This PR ensures the applicator is made aware of fills on external orders using the same code paths for handling fills on internal orders. Currently, we do not record historical state in the relayer, but we do emit `SystemBusMessage::OrderMetadataUpdated` messages that are propagated to clients listening to the order history topic over socket connections.

### Testing
- [x] Tested in testnet that `OrdermetadataUpdated` events are emitted by the relayer and received by clients over socket

### DIFF
--- a/workers/chain-events/src/post_settlement.rs
+++ b/workers/chain-events/src/post_settlement.rs
@@ -1,0 +1,127 @@
+//! Post-settlement helpers for external match processing
+//!
+//! This module holds a small immutable context struct (`PostSettlementCtx`) and
+//! a set of methods on `OnChainEventListenerExecutor` that apply
+//! post-settlement side-effects.
+
+use crate::error::OnChainEventListenerError;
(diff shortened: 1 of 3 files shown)


workers/chain-events/src/post_settlement.rs
```diff
@@ -0,0 +1,127 @@
...
+        Ok(())
+    }
+
+    /// Emit `ExternalFill` relayer event
+    pub(crate) fn emit_event(
+        &self,
+        order_id: OrderIdentifier,
+        ctx: &PostSettlementCtx<'_>,
+    ) -> Result<(), OnChainEventListenerError> {
+        use job_types::event_manager::{ExternalFillEvent, RelayerEventType, try_send_event};
```
joeykraut, line 89: can you move this import up to the top?

PR #1068, opened 2025-07-15
state, common: bound fills array growth if historical state disabled

sehyunc:
### Purpose
This PR adds a roll-up helper on OrderMetadata that collapses all but the most recent fill into a single aggregate entry, ensuring that fill arrays are bounded while
- `total_filled()` callsites receive correct data
- newest fill is still delivered to subscribers (frontends and other clients).

This is currently invoked in the external match flow iff historical state is disabled.

### Testing
- [x] Test in testnet

### DIFF
--- a/common/src/types/wallet/order_metadata.rs
+++ b/common/src/types/wallet/order_metadata.rs
@@ -8,6 +8,9 @@ use crate::types::price::TimestampedPrice;
 
 use super::{Order, OrderIdentifier};
 
+/// The maximum number of recent fills to retain for an order
+const MAX_RECENT_FILLS: usize = 2; // 1 aggregate + 1 most-recent fill kept
+
 /// The state of an order in the wallet
 #[derive(Clone, Debug, Copy, PartialEq, Eq, Hash, Serialize, Deserialize)]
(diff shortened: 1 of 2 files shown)


common/src/types/wallet/order_metadata.rs
```diff
@@ -63,6 +66,37 @@ impl OrderMetadata {
...
+
+        // Use the price of the newest fill being aggregated as the representative price
+        // for the aggregate entry. Downstream consumers generally ignore this
+        // price for aggregates, but we must supply *some* value.
+        let representative_price = self.fills.last().unwrap().price;
+        let aggregate_fill = PartialOrderFill::new(rolled_amount, representative_price);
+
+        // Rebuild the fills vector: [aggregate, newest...]
+        self.fills = vec![aggregate_fill];
+        self.fills.extend(newest);
```
joeykraut, line 98: perhaps a more idiomatic way to do this with iterators:
```rust
self.fills = iter::once(aggregate_fill).chain(newest).collect();
```

joeykraut (approved):

PR #1069, opened 2025-07-21
external-api: external-match: Add options for configuring relayer fee

joeykraut:
### Purpose
This PR adds options for configuring the relayer fee on an external match. This will be set by the auth server on a per-request basis.

### Todo
- A follow up will add an integration test for setting this value

### Testing
- [x] All unit and integration tests pass

### DIFF
--- a/workers/chain-events/src/post_settlement.rs
+++ b/workers/chain-events/src/post_settlement.rs
@@ -11,7 +11,7 @@ use common::types::{
     price::TimestampedPrice,
     wallet::{OrderIdentifier, WalletIdentifier, order_metadata::OrderState},
 };
-use constants::EXTERNAL_MATCH_RELAYER_FEE;
+use constants::DEFAULT_EXTERNAL_MATCH_RELAYER_FEE;
 use job_types::event_manager::{ExternalFillEvent, RelayerEventType, try_send_event};
 use renegade_metrics;
 use util::{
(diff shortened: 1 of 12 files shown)


workers/chain-events/src/post_settlement.rs
```diff
@@ -119,7 +119,7 @@ fn execution_price(external_match_result: &ExternalMatchResult) -> TimestampedPr
 
 /// Compute internal party fee take
 fn internal_fee_take(external_match_result: &ExternalMatchResult) -> FeeTake {
-    let relayer_fee = FixedPoint::from_f64_round_down(EXTERNAL_MATCH_RELAYER_FEE);
+    let relayer_fee = FixedPoint::from_f64_round_down(DEFAULT_EXTERNAL_MATCH_RELAYER_FEE);
```
sehyunc, line 122: We'll be recording incorrect fees in the HSE for now, which is probably fine

sehyunc (approved):
lgtm

workers/chain-events/src/post_settlement.rs
joeykraut: Well actually the internal party's relayer fee is still zero, and this value is being recorded for the internal party's fill so this is still accurate

sehyunc: I see, that makes sense

PR #1071, opened 2025-07-23
common: types: token: Add `maybe_from_ticker` method

joeykraut:
### Purpose
This PR adds a `maybe_from_ticker` method which allows the caller to handle the case in which a token's ticker is unsupported.

### Testing
- [x] All unit tests pass


joeykraut:
> Is this a permanent addition or temporary to address the cargo issue you mentioned?

permanent, i want to use this in consumer crates in the auth server

PR #1075, opened 2025-07-28
price-reporter: Initialize streams up-front

joeykraut:
### Purpose
This PR initializes the price reporter's streams up-front in the price reporter. Previously we relied on the handshake manager to do so in its startup phase.

Aside from being cleaner, this will allow us to know the `SharedPriceStreams` contents up front and warp the `HashMap` in an `Arc` rather than a `RwLock` -- i.e. we make the `HashMap` entries read-only. We will then pass this type around the relayer to all price consumers for fast sampling.

### Testing
- [x] Ran relayer locally, requested prices


workers/price-reporter/src/manager/utils.rs
```diff
@@ -147,5 +177,16 @@ pub fn eligible_for_stable_quote_conversion(
...
+    if base.is_stablecoin() || !quote.is_stablecoin() {
+        return false;
+    }
+
+    // We assume a 1:1 USD:USDC for Coinbase markets
+    let default_stable = default_exchange_stable(exchange);
+    let usd = Token::from_ticker(USD_TICKER);
+    if default_stable == usd {
+        return false;
+    }
```
joeykraut, line 189: this fixes a bug wherein the Coinbase stream was not being considered in the price deviation calculation

workers/price-reporter/src/manager/external_executor.rs
```diff
@@ -445,14 +406,7 @@ async fn resubscribe_to_prior_streams(
...
     for (exchange, base_token, quote_token) in streams {
-        subscribe_to_price_stream(
-            subscription_states,
-            exchange,
-            base_token,
-            quote_token,
-            msg_out_tx.clone(),
-        )
-        .await?;
+        subscribe_to_price_stream(exchange, &base_token, &quote_token, &msg_out_tx)?;
```
sehyunc, line 409: It seems like we no longer initialize states in `subscribe_to_price_stream`. Wouldn't that result in the next `new_price()` erroring when it tries to `.get()` after a resubscribe since the key has not been initialized to `AtomicPriceStreamState::default()`?

sehyunc (approved):
Question about resubscribe logic and price states map, otherwise lgtm

workers/api-server/src/http/price_report.rs
```diff
@@ -84,10 +81,12 @@ impl TypedHandler for GetSupportedTokensHandler {
         _params: UrlParams,
         _query_params: QueryParams,
     ) -> Result<Self::Response, ApiServerError> {
-        let tokens = get_all_tokens_filtered(&FILTERED_TOKENS)
+        let tokens = get_all_base_tokens()
             .into_iter()
+            .chain(iter::once(Token::usdc()))
```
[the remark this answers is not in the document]
joeykraut: yeah good call, I'll actually go make that change

workers/price-reporter/src/manager/external_executor.rs
joeykraut: We also no longer remove the keys in a re-subscribe, we just write over their values to zero (so that they're ignored). So the keys will remain in the map

sehyunc: Just to be clear, you're saying `.clear()` overwrites with zero but keeps keys in the map?

PR #1081, opened 2025-08-01
price-reporter: Support USDT price streaming

joeykraut:
### Purpose
This PR adds $USDT to the `get_all_base_tokens` return value by filtering on a new `is_tradable` method which returns true for USDT. 

I also reformatted the request format for the `/v0/price_report` endpoint. This endpoint is only used internally, so changing its interface is fine.

### Testing
- [x] Unit tests pass
- [x] Fetched USDT prices from the relayer


akirillo (approved):
nice

PR #1083, opened 2025-08-04
state: interface: Use blocking DB operations in short txns

joeykraut:
### Purpose
This PR adds a `with_blocking_read_tx` and `with_blocking_write_tx` which together expose a blocking transaction interface on top of the MDBX instance. 

This is particularly useful for short-lived transactions in which the Tokio overhead to spawn the transaction on a blocking thread contributes >80% of the latency. This was measured in testnet and locally in benchmarks.

I use the blocking interface for queries that are known to be short, particularly around reading node metadata and fetching orders for the matching engine. Empirically this shaves ~10-15% off quote latency.

### Testing
- [x] All unit tests pass
- [ ] Testing in testnet


akirillo (approved):
neat

PR #1085, opened 2025-08-05
proof-manager: Refactor existing impl into `NativeProofManager`

joeykraut:
### Purpose
This PR refactors the proof manager's structure to allow for multiple underlying manager implementations. This will be followed by an implementation which calls the external proof manager if configured.

### Testing
- [x] All unit tests pass
- [x] All integration tests pass
- [x] Tested the native prover on a locally running relayer


config/src/cli.rs
```diff
@@ -283,30 +296,41 @@ pub struct RelayerConfig {
...
     /// The address of the compliance service to use for wallet screening. If
     /// not configured, wallet screening is disabled
     ///
     /// The API of the compliance service must match that defined here:
     ///     https://github.com/renegade-fi/relayer-extensions/tree/master/compliance/compliance-api
-    pub compliance_service_url: Option<String>,
+    pub compliance_service_url: Option<Url>,
     /// The URL to export relayer events to.
     /// If omitted, the event manager is disabled.
     pub event_export_url: Option<Url>,
```
akirillo, line 315: nit: if we have the `event_export_url` under the `External Services` heading here, we should do so in the `Cli` struct above as well

akirillo (approved):

PR #1090, opened 2025-08-06
task-driver: validity-proofs: Link reblind-commitments through proof manager

joeykraut:
### Purpose
This PR adds a proof manager job to link a proof of `VALID REBLIND` and a proof of `VALID COMMITMENTS` and uses this job in the task driver when we need to re-prove validity proofs. 

I also implemented the external prover client handler for this job.

### Testing
- [x] All unit tests pass
- [x] All integration tests pass
- [x] Tested against a locally running prover service instance

### DIFF
--- a/workers/proof-manager/src/implementations/external_proof_manager/prover_service_client.rs
+++ b/workers/proof-manager/src/implementations/external_proof_manager/prover_service_client.rs
@@ -28,12 +28,15 @@ use serde::{Deserialize, Serialize};
 
 use crate::{
     error::ProofManagerError,
-    implementations::external_proof_manager::api_types::{
-        ProofAndHintResponse, ProofAndLinkResponse, ProofResponse, ValidCommitmentsRequest,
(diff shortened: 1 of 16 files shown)


workers/proof-manager/src/implementations/external_proof_manager/prover_service_client.rs
```diff
@@ -130,8 +130,7 @@ impl ProofServiceClient {
         let req = ValidWalletCreateRequest { statement: statement.clone(), witness };
         let res = self.send_request::<_, ProofResponse>(VALID_WALLET_CREATE_PATH, req).await?;
 
-        let link_hint = default_link_hint();
```
akirillo, line 133: ty for changing the `ProofBundle` type to only include these link hints for commitments & reblind, the usage of `default_link_hint` was a bit confusing before

akirillo (approved):

PR #1091, opened 2025-08-07
task-driver: Bypass task queue for shared matches

joeykraut:
### Purpose
This PR changes the relayer to bypass the task queue for shared bundles. This will remove much of the raft contention that has been causing high latencies on the `/assemble-external-match` endpoint. 

### Testing
- [x] Unit tests pass
- [x] Tested locally
- [x] Testing in testnet


sehyunc (approved):
IIUC because shared SettleMatch tasks now bypass Raft:
- task status websocket updates are no longer published
- `RelayerEventType::TaskCompletion` event is not fired, so no recording to HSE
- `/v0/tasks/:task_id` will return 404 for these task ids

This sounds fine since at a cursory glance it doesn't seem like we explicitly await for settlement tasks. Dashboard will be missing these tasks, though.

joeykraut:
> IIUC because shared SettleMatch tasks now bypass Raft:
> 
> * task status websocket updates are no longer published
> * `RelayerEventType::TaskCompletion` event is not fired, so no recording to HSE
> * `/v0/tasks/:task_id` will return 404 for these task ids
> 
> This sounds fine since at a cursory glance it doesn't seem like we explicitly await for settlement tasks. Dashboard will be missing these tasks, though.

We already don't record historical tasks for external match settlements because there isn't currently a great way of determining whether they settled at the time the state machine marks the task completed.

You can see it filtered out in [`from_task_descriptor`](https://github.com/renegade-fi/renegade/blob/adac463864eb9387fdbf98815abf8a90771a3b22/common/src/types/tasks/history.rs#L66-L98)

PR #1095, opened 2025-08-11
config: Cleanup crate before adding per-asset relayer fees

joeykraut:
### Purpose
This PR makes a few cleanups to the `config` crate before adding per-asset relayer fees to the config:
- Remove unused config vars
- Refactor the parsing logic into a submodule
- Change the `cluster_keypair` type into a wrapped type, on which we can implement helpers and derive `Clone`. This removes the need for the manual `Clone` impl of the `RelayerConfig`.

### Testing
- [x] Unit tests pass
- [x] Integration tests pass


mock-node/src/lib.rs
```diff
@@ -157,15 +156,6 @@ impl MockNodeController {
         }
     }
```
sehyunc, line 158: > The keypair type doesn't have a `Clone` method for safety so this workaround is necessary`

What were the safety reasons?

sehyunc (approved):
lgtm

PR #1096, opened 2025-08-11
config: Add per-asset match fee configs

joeykraut:
### Purpose
This PR adds per-asset relayer fee config values to the relayer's CLI. I added support for passing these both as CLI arguments and via the `config.toml` file.

I also clarified the `match_fee` argument to a `max_match_fee` which more closely aligns with its current use.

### Testing
- [x] Unit tests pass
- [x] Tested relayer locally


sehyunc (approved):
clean, lgtm

PR #1101, opened 2025-08-13
task-driver: tasks: Add `bypass_task_queue` method to task descriptors

joeykraut:
### Purpose
This PR adds the `Descriptor` trait which all task descriptors must implement. This trait then exposes a `bypass_task_queue` method, analogous to the method defined on the `Task` trait. We use this method to skip calling `pop_task` on tasks which bypass the queue.

This was the cause of the "key not found for task" errors that we've been seeing in production.

### Testing
- [x] Unit tests pass
- [x] Integration tests pass
- [ ] Test in testnet

### DIFF
--- a/workers/task-driver/src/driver.rs
+++ b/workers/task-driver/src/driver.rs
@@ -29,7 +29,7 @@ use crate::{
         settle_match_internal::SettleMatchInternalTask, update_merkle_proof::UpdateMerkleProofTask,
         update_wallet::UpdateWalletTask,
     },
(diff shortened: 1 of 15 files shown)


akirillo (approved):
Lgtm, can we also use this in the `enqueue_task_await_completion` method?

joeykraut:
> Lgtm, can we also use this in the `enqueue_task_await_completion` method?



> Lgtm, can we also use this in the `enqueue_task_await_completion` method?

Sure, good call

PR #1103, opened 2025-08-15
api-server: external-match: processor: Remove exlusive bundles

joeykraut:
### Purpose
This PR removes shared bundles from the relayer. We now implicitly assume all bundles are shared and bypass the task queues.

### Testing
- [x] All unit tests pass
- [x] Integration tests pass
- [ ] Testing in testnet


sehyunc (approved):
very clean

PR #1105, opened 2025-08-21
darkpool-client: add timeout to tx submission

akirillo:
This PR makes a couple changes to TX submission in the darkpool client(s):
1. We add a 1-minute timeout to the awaiting of tx mining
2. We use `eth_gasPrice` as an estimate of gas price, rather than the latest block's basefee

We do this because we have seen a number of tasks get stuck indefinitely in the "submitting tx" phase, presumably because the pending txs are stuck in the mempool. As such, we use `eth_gasPrice` because it is an all-in estimate (basefee + priority fee) of the gas price to set, and it's a cheaper RPC call. We add the timeout to ensure that tasks don't hang indefinitely, but rather error out.

### Testing
- [x] Test internal trade on Base Sepolia

### DIFF
--- a/darkpool-client/src/traits.rs
+++ b/darkpool-client/src/traits.rs
@@ -1,6 +1,7 @@
 //! Trait definitions for the `DarkpoolClient`
 
-use alloy::eips::BlockId;
+use std::time::Duration;
+
 use alloy::providers::Provider;
 use alloy::rpc::types::TransactionRequest;
 use alloy::{primitives::Address, rpc::types::TransactionReceipt};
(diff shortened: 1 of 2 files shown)


darkpool-client/src/traits.rs
```diff
@@ -223,20 +234,10 @@ pub trait DarkpoolImplExt: DarkpoolImpl {
...
-
-        let latest_basefee = latest_block
-            .header
-            .base_fee_per_gas
-            .ok_or(DarkpoolClientError::rpc("No basefee found"))?;
-        let gas_price = (latest_basefee * 2) as u128;
-        Ok(gas_price)
+        // Set the gas price to 2x the latest gas price estimate for simplicity
+        let gas_price = self.provider().get_gas_price().await.map_err(DarkpoolClientError::rpc)?;
+        let adjusted_gas_price = gas_price * 2;
```
joeykraut, line 239: do we still need this value to be 2x now that we have a more complete estimate? Might be worth playing around with.

Also could you constantize it since we'll likely tune it (if not now, then at some point)

joeykraut (approved):
nice find

darkpool-client/src/traits.rs
akirillo: I'd like to start w/ 2x and leave tuning out of scope for this deploy. But will constantize, yes.

joeykraut: Fair enough, we can come back to it

PR #1113, opened 2025-09-17
task-driver: running-task: Do not skip wallet refresh if `pop` fails

joeykraut:
### Purpose
This PR removes the error propagation from `pop` errors in the task cleanup phase. If an error occurs with popping the task from all queues; we should refresh the wallets, which the next lines do anyways.

Propagating this error will skip the wallet refresh task; which may be necessary in cases where the task has failed. Further, the wallet refresh task will cleanup the task queues of all affected wallets.

### Testing
- [x] Unit tests pass
- [ ] Testing in arbitrum sepolia


akirillo (approved):
good catch

PR #1117, opened 2025-09-25
state: interface: order-book: Read from new proof store

joeykraut:


common/src/types/network_order.rs
```diff
@@ -102,7 +102,7 @@ impl NetworkOrder {
     /// This amounts to whether the order has validity proofs and witnesses
     /// attached to it
     pub fn ready_for_match(&self) -> bool {
-        self.validity_proofs.is_some() && self.validity_proof_witnesses.is_some()
+        self.state == NetworkOrderState::Verified
```
akirillo, line 105: Cursor bot kinda has a point here - maybe we should just remove `ready_for_match` and replace its callsites with helpers that can look up validity proofs / witnesses in state?

akirillo (approved):

common/src/types/network_order.rs
joeykraut: practically speaking these are equivalent conditions. The order state should stay consistent with the validity proofs and we should just ensure that in the applicator (it already is). The issue with the suggestion to lookup proofs in the callsites is that the state isn't necessarily available at the callsites, nor can it be made so easily.

PR #1126, opened 2025-10-02
common: types: order: Move precompute cancellation flag to `Order`

joeykraut:
### Purpose
This PR makes two changes:
- Moves the `precompute_cancellation_proofs` flag to the `Order` type.
- Creates logic to recompute cancellation proofs for all orders in the same path as the validity proofs, so that pre-computed cancellation proofs will always stay up to date.

### Todo
- Delete cancellation proofs from state when an order is cancelled.

### Testing
- [x] Unit tests pass
- [x] Tested locally using a variety of tasks to make sure that valid cancellation proofs are in-place after each task.

### DIFF
--- a/workers/task-driver/src/utils/proofs/cancellation_proofs.rs
+++ b/workers/task-driver/src/utils/proofs/cancellation_proofs.rs
@@ -1,0 +1,77 @@
+//! Utils for precomputing cancellation proofs for orders
+
+use common::types::{
+    proof_bundles::ValidWalletUpdateBundle,
+    wallet::{OrderIdentifier, Wallet},
+};
+use job_types::proof_manager::ProofJob;
+use tracing::error;
+use util::raw_err_str;
+
+use crate::{
+    tasks::update_wallet::UpdateWalletTask,
+    traits::TaskContext,
+    utils::{enqueue_proof_job, proofs::map_proof_result},
+};
+
+/// Precompute cancellation proofs for a wallet
+pub(crate) async fn precompute_cancellation_proofs(
+    wallet: &Wallet,
+    ctx: &TaskContext,
+) -> Result<(), String> {
+    // Spawn a task for each order that requires a cancellation proof
+    let mut handles = Vec::new();
+    for (id, order) in wallet.orders.iter() {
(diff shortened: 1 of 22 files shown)


workers/task-driver/src/utils/proofs/mod.rs
```diff
@@ -0,0 +1,40 @@
...
+    // Join both threads and handle errors
+    let (validity_res, cancellation_res) = tokio::join!(validity_jh, cancellation_jh);
+    map_proof_result(validity_res)?;
+    map_proof_result(cancellation_res)
+}
+
+/// Map a proof result to a string error
+#[inline]
+fn map_proof_result<T>(result: Result<Result<T, String>, JoinError>) -> Result<T, String> {
+    result.map_err(|e| e.to_string())?.map_err(|e| e.to_string())
```
akirillo, line 39: nit: the latter `.map_err(|e| e.to_string())` is unnecessary

workers/task-driver/src/utils/proofs/cancellation_proofs.rs
```diff
@@ -0,0 +1,74 @@
...
+            precompute_cancellation_proof_for_order(oid, wallet_clone, &ctx_clone).await
+        });
+        handles.push(handle);
+    }
+
+    // Await the tasks
+    let mut proofs = Vec::new();
+    for handle in handles {
+        // Do not fail the whole operation if one proof fails
+        let (oid, proof) = map_proof_result(handle.await)?;
```
akirillo, line 41: it looks like we _do_ fail the whole operation if one proof fails - did you mean to `match` on the result here and log errors?

akirillo (approved):

workers/task-driver/src/utils/proofs/cancellation_proofs.rs
joeykraut: Good catch

joeykraut:
### Merge activity

* **Oct 3, 3:25 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/renegade-fi/renegade/1126).

PR #1133, opened 2025-10-30
darkpool-client: Use `quote_amount` from call in share parsing

joeykraut:
### Purpose

This PR fixes a bug in parsing shares from a malleable match. In particular, we were previously inferring the quote amount from the swapped base amount.

If a client or the contracts uses a different quote amount--which is valid so long as it represents a worse price for the external party than the price authorized by the relayer--then we will apply an inconsistent `ExternalMatchResult` to the recovered shares.

This was causing a bug with the malleable match connector which can be off by one in its math due to its implementation of fixed point division.

Now, we pull the `quote_amount` directly from the transaction payload.

### Testing

- [x] Tested recovering a wallet that was seeing errors in testnet
- [x] Testing in testnet

### DIFF
--- a/circuit-types/src/match.rs
+++ b/circuit-types/src/match.rs
@@ -259,4 +259,19 @@ impl BoundedMatchResult {
             direction: self.direction,
         }
     }
+
+    /// Convert to an external match result given an exact base and quote amount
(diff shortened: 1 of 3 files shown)


sehyunc (approved):
Golly, what a find

PR #1218, opened 2025-11-16
api-server: external-match: Add direct malleable match

joeykraut:
### Purpose
This PR adds a direct, malleable match path to the relayer. This allows a user to request a malleable match without first receiving a quote.

I chose not to integrate this with the existing endpoint for ease of implementation--this will all be removed in v2 anyways.

### Testing
- [x] All tests pass


sehyunc (approved):
LGTM

PR #1229, opened 2025-12-12
circuits: zk-circuits: intent-only-bounded-settlement: add circuit impl

sehyunc:
### Purpose

This PR adds the `INTENT ONLY BOUNDED SETTLEMENT` circuit and defines its constraints. Also adds very basic testing.

### Todo

- Link `INTENT ONLY VALIDITY <> INTENT ONLY BOUNDED SETTLEMENT`
- More tests

### Testing

- [x] All tests pass

### DIFF
--- a/circuits/src/test_helpers/fuzzing.rs
+++ b/circuits/src/test_helpers/fuzzing.rs
@@ -6,6 +6,7 @@ use alloy_primitives::Address;
 use circuit_types::{
     AMOUNT_BITS, Amount,
     balance::Balance,
+    bounded_match_result::BoundedMatchResult,
     csprng::PoseidonCSPRNG,
     deposit::Deposit,
     elgamal::{DecryptionKey, EncryptionKey},
@@ -80,6 +81,12 @@ pub fn random_fee() -> FixedPoint {
     FixedPoint::from_f64_round_down(fee_f64)
 }
 
+/// Generate a random block deadline
+pub fn random_block_deadline() -> u64 {
+    let mut rng = thread_rng();
+    rng.gen_range(0..=u64::MAX)
+}
+
 // ---------------
 // | State Types |
 // ---------------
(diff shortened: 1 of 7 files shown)


circuits/src/test_helpers/fuzzing.rs
```diff
@@ -205,6 +212,37 @@ pub fn create_settlement_obligation_with_balance(
...
+
+/// Create a bounded match result for an intent and balance amount
+pub fn create_bounded_match_result_with_balance(
+    intent: &Intent,
+    balance_amount: Amount,
+) -> BoundedMatchResult {
+    let mut rng = thread_rng();
+
+    // Clamp the obligation's `amount_in` to avoid price overflows
+    let mut max_amount_in = 2u128.pow((AMOUNT_BITS / 2) as u32);
```
joeykraut: nit: can you break this out into a helper and replace its use elsewhere? e.g.
```rust
fn bounded_amount_max() -> Amount {
    2u128.pow((AMOUNT_BITS / 2) as u32)
}
```

Looks like some of the helpers above actually incorrectly use `1.pow(...)`, so could you replace those while you're at it?

circuits/src/zk_circuits/settlement/intent_only_bounded_settlement.rs
```diff
@@ -0,0 +1,228 @@
...
+}
+
+// ---------------------------
+// | Witness Type Definition |
+// ---------------------------
+
+/// The witness type for `INTENT ONLY BOUNDED SETTLEMENT`
+#[circuit_type(serde, singleprover_circuit)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct IntentOnlyBoundedSettlementWitness<const MERKLE_HEIGHT: usize> {
```
joeykraut: we don't need a `MERKLE_HEIGHT` sizing constant here since no element depends on it. This also means we don't need a sized type below.

Looks like I accidentally added them to some of the other settlement circuits. I'll push a change to remove that.

joeykraut (approved):
Looks secure, very nicely done!

PR #1240, opened 2025-12-17
circuits: zk-circuits: intent-only-bounded-settlement: Add proof linking

sehyunc:
### Purpose

This PR adds proof linking helpers and tests between `INTENT ONLY VALIDITY` <-> `INTENT ONLY BOUNDED SETTLEMENT` and `INTENT ONLY FIRST FILL VALIDITY` <-> `INTENT ONLY BOUNDED SETTLEMENT`. 

The witness value being linked is the same between intent only public settlement and intent only bounded settlement, thus we share layouts by having the bounded circuit inherit the layout from the public circuit which places the link group.

### Testing

- [x] All tests pass

### DIFF
--- a/circuits/src/zk_circuits/proof_linking/intent_only.rs
+++ b/circuits/src/zk_circuits/proof_linking/intent_only.rs
@@ -10,18 +10,18 @@ use mpc_relation::proof_linking::GroupLayout;
 
 use crate::zk_circuits::{
     settlement::{
-        INTENT_ONLY_PUBLIC_SETTLEMENT_LINK,
+        INTENT_ONLY_SETTLEMENT_LINK,
         intent_only_public_settlement::IntentOnlyPublicSettlementCircuit,
     },
     validity_proofs::intent_only::IntentOnlyValidityCircuit,
 };
 
-// ----------------------------------------------------------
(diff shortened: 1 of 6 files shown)


circuits/src/zk_circuits/proof_linking/intent_only.rs
```diff
@@ -10,18 +10,19 @@ use mpc_relation::proof_linking::GroupLayout;
...
         intent_only_public_settlement::IntentOnlyPublicSettlementCircuit,
     },
     validity_proofs::intent_only::IntentOnlyValidityCircuit,
 };
 
-// ----------------------------------------------------------
-// | Intent Only Validity <-> Intent Only Public Settlement |
-// ----------------------------------------------------------
+// -----------------------------------------------------------------------
+// | Intent Only Validity <-> Intent Only Settlement (Public or Bounded) |
```
joeykraut: I'd change this to "exact or bounded" everywhere. The difference between the two settlements is not that one is public and one is bounded

joeykraut (approved):
Only thing I'd suggest is to re-read through all the comments and make sure we're referring to these as exact vs bounded; both are public so that's not an interesting dichotomy

PR #1272, opened 2026-01-07
workers: gossip-server: remove duplicate util dependency

akirillo:
In this PR, we remove the duplicate definition of the `util` dependency in the `gossip-server` crate. This prevents other projects from importing the crates defined here.


crates/workers/gossip-server/Cargo.toml
```diff
@@ -24,7 +24,6 @@ constants = { workspace = true }
 gossip-api = { workspace = true }
 job-types = { workspace = true }
 state = { workspace = true }
-util = { workspace = true }
```
joeykraut: can you remove the one above so they stay alphabetized

joeykraut (approved):

PR #1322, opened 2026-01-27
state: account_index: Add owner and intent hash reverse indexes

sehyunc:
### Purpose
Add reverse indexes for routing blockchain events to internal state. The owner index maps (owner, token) → account_id for ERC20 Transfer events, and the intent hash index maps intent_hash → (account_id, order_id) for public intent events. Also writes owner index on balance updates.

### Testing
- [x] Unit tests pass

### DIFF
--- a/crates/state/src/applicator/account_index.rs
+++ b/crates/state/src/applicator/account_index.rs
@@ -139,6 +139,9 @@ impl StateApplicator {
             return Err(StateApplicatorError::reject("account not found"));
         }
         tx.update_balance(&account_id, balance)?;
+
+        // Index the balance for owner lookup (enables routing balance update events)
+        tx.set_owner_index(&balance.owner(), &balance.mint(), &account_id)?;
         tx.commit()?;
 
         // Open a read transaction to get order info for matching engine updates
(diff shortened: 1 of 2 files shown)


crates/state/src/storage/tx/account_index.rs
```diff
@@ -99,6 +99,20 @@ fn order_index_key(order_id: &OrderId) -> String {
     format!("order_index:{order_id}")
 }
 
+/// Build the key for the owner -> account index
+///
+/// Maps (owner, token) to account_id for routing balance update events
+fn owner_index_key(owner: &Address, token: &Address) -> String {
+    format!("owner_index:{owner:?}:{token:?}")
```
joeykraut: why do we need owner and token here in the key? The logical mapping should be `address -> account_id` 1:1

crates/state/src/applicator/account_index.rs
```diff
@@ -139,6 +139,9 @@ impl StateApplicator {
             return Err(StateApplicatorError::reject("account not found"));
         }
         tx.update_balance(&account_id, balance)?;
+
+        // Index the balance for owner lookup (enables routing balance update events)
+        tx.set_owner_index(&balance.owner(), &balance.mint(), &account_id)?;
```
joeykraut: don't we also need to write to the intent hash index in the applicator somewhere?

PR #1323, opened 2026-01-27
chain-events: Replace nullifier subscription with ERC20 Transfer events

sehyunc:
### Purpose

This PR replaces the disabled nullifier event subscription with ERC20 Transfer event subscription and implements full event handling. When a Transfer event is detected:

1. Look up accounts for both `from` and `to` addresses
2. Fetch on-chain balance and update matching engine cache (all nodes, before raft)
3. Propose balance update via raft (selected node only)
4. Run matching engine on affected orders

Also adds `get_account_for_owner` to state for routing balance updates to the correct account.

### Testing

- [x] `cargo check -p chain-events` passes

### DIFF
--- a/crates/state/src/interface/account_index.rs
+++ b/crates/state/src/interface/account_index.rs
@@ -169,6 +169,23 @@ impl StateInner {
         .await
     }
 
+    /// Get the account ID for a given owner and token
+    ///
+    /// Used to route balance update events to the correct account
+    pub async fn get_account_for_owner(
+        &self,
+        owner: &Address,
+        token: &Address,
+    ) -> Result<Option<AccountId>, StateError> {
+        let owner = *owner;
+        let token = *token;
+        self.with_read_tx(move |tx| {
+            let account_id = tx.get_account_for_owner(&owner, &token)?;
+            Ok(account_id)
+        })
+        .await
+    }
+
     // -----------
     // | Setters |
     // -----------
--- a/crates/workers/chain-events/src/listener.rs
+++ b/crates/workers/chain-events/src/listener.rs
@@ -1,24 +1,33 @@
 //! Defines the core implementation of the on-chain event listener
 
-use std::thread::JoinHandle;
+use std::{thread::JoinHandle, time::Duration};
 
 use super::error::OnChainEventListenerError;
 use alloy::{
-    primitives::Address,
+    primitives::{Address, keccak256},
     providers::{DynProvider, Provider, ProviderBuilder, WsConnect},
     rpc::types::{Filter, Log},
     sol_types::SolEvent,
 };
 use constants::in_bootstrap_mode;
-use darkpool_client::client::erc20::abis::erc20::IERC20;
 use darkpool_client::DarkpoolClient;
+use darkpool_client::client::erc20::abis::erc20::IERC20;
 use futures_util::StreamExt;
-use job_types::event_manager::EventManagerQueue;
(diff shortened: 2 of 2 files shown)


crates/state/src/interface/account_index.rs
```diff
@@ -169,6 +169,23 @@ impl StateInner {
         .await
     }
 
+    /// Get the account ID for a given owner and token
+    ///
+    /// Used to route balance update events to the correct account
+    pub async fn get_account_for_owner(
+        &self,
+        owner: &Address,
+        token: &Address,
```
joeykraut: token shouldn't matter here, same as in the last PR. There is fundamentally a 1:1 mapping from `owner EOA <-> account`

crates/workers/chain-events/src/listener.rs
```diff
@@ -114,294 +115,221 @@ impl OnChainEventListenerExecutor {
...
+        // Fetch cluster state
+        let state = self.state();
+        let my_id = state.get_peer_id()?;
+        let cluster_id = state.get_cluster_id()?;
+        let mut peers = state.get_cluster_peers(&cluster_id).await?;
+        peers.sort();
+
+        // Compute the selected node
+        let n_peers = peers.len();
+        let mut input = Vec::with_capacity(72);
```
joeykraut: can you add a comment where 72 comes from

crates/workers/chain-events/src/listener.rs
```diff
@@ -114,294 +115,221 @@ impl OnChainEventListenerExecutor {
...
+    ) -> Result<(), OnChainEventListenerError> {
+        // If the current node is not the update handler, sleep for a random
+        // timeout then check that the balance state updates have been processed.
+        // We do this as a crash recovery mechanism to ensure that the updates are
+        // processed even if the selected node is crashed
+        if !self.should_execute_balance_updates(&owner, &token, &tx_hash).await? {
+            let timeout = rand::thread_rng()
+                .gen_range(MIN_BALANCE_UPDATE_DELAY_S..=MAX_BALANCE_UPDATE_DELAY_S);
+            tokio::time::sleep(Duration::from_secs(timeout)).await;
+        }
```
joeykraut: When the non-proposers wake from this sleep, we should check whether an update is necessary. 

Maybe the cheapest way to do that is to measure the balance before and after in the relayer state and see if it's changed, or changed by a specific amount. Another way would be to check the on-chain balance, but I'll let you decide whether those RPC costs are worth it.

crates/workers/chain-events/src/listener.rs
```diff
@@ -114,294 +115,221 @@ impl OnChainEventListenerExecutor {
...
+    /// Run the matching engine on orders that use the given token as input
+    async fn run_matching_engine_for_account_and_balance(
+        &self,
+        account_id: AccountId,
+        token: Address,
+    ) -> Result<(), OnChainEventListenerError> {
+        // Get all order IDs that use this token as input
+        let order_ids = self.state().get_orders_with_input_token(&account_id, &token).await?;
+
+        for order_id in order_ids {
```
joeykraut: I think we could do something like the `RunMatchingEngineForBalanceHook` does [here](https://github.com/renegade-fi/renegade/blob/sehyun/websocket-ingestion/crates/workers/task-driver/src/hooks.rs#L107-L112) in that we only run the engine for orders capitalized by the updated balance.

PR #1341, opened 2026-01-31
chain-events: Ring 0 compatible listener

sehyunc:
### Purpose

This PR adds owner-based routing for chain events (ERC20 transfers, Permit2 approvals, darkpool events). The owner index maps EOA addresses to account IDs, allowing the chain-events worker to route incoming events to the correct account. Also adds cleanup logic to remove stale owner index entries when the last order is removed from an account.

Key changes:
- **State**: Add owner index storage/lookup, intent hash index, and cleanup on last order removal
- **Chain-events**: Refactor listener into modular handlers (ERC20, Permit2, Darkpool) that use owner index for routing
- **System bus**: Add `OwnerIndexChanged` message for subscription refresh

### Testing
- [x] Local relayer pointed at Base Sepolia with Transfer and Approve transactions
- [x] Observed chain-events listener logs for correct routing

### DIFF
--- a/crates/system-primitives/system-bus/src/message.rs
+++ b/crates/system-primitives/system-bus/src/message.rs
@@ -133,7 +133,8 @@ pub enum SystemBusMessage {
     NoExternalMatchFound,
 
     // --- Admin -- //
-    /// A message indicating that an order has been updated, for admin consumption
+    /// A message indicating that an order has been updated, for admin
+    /// consumption
     AdminOrderUpdate {
         /// The ID of the account that owns the order
         account_id: AccountId,
@@ -144,7 +145,8 @@ pub enum SystemBusMessage {
         /// The type of update
         update_type: AdminOrderUpdateType,
     },
-    /// A message indicating that a balance has been updated, for admin consumption
+    /// A message indicating that a balance has been updated, for admin
+    /// consumption
     AdminBalanceUpdate {
         /// The ID of the account that owns the balance
         account_id: AccountId,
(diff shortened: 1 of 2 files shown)


crates/state/src/applicator/mod.rs
```diff
@@ -52,6 +53,9 @@ pub struct StateApplicatorConfig {
     pub allow_local: bool,
     /// The local peer's cluster ID
     pub cluster_id: ClusterId,
+    /// The executor address, used to verify the executor's signature on the
+    /// settlement obligation at fill time
+    pub executor_address: Address,
```
joeykraut: This is in state, we can look it up in the tx to shrink the config if you prefer. See [here](https://github.com/renegade-fi/renegade/blob/yun/chain-events-listener-v2/crates/workers/task-driver/src/tasks/settlement/helpers/mod.rs#L70-L73)

crates/workers/chain-events/src/lib.rs
```diff
@@ -18,6 +18,8 @@
 #![allow(incomplete_features)]
 
 pub mod error;
-pub mod listener;
-// pub mod post_settlement;
+mod executor;
+mod handlers;
```
joeykraut: I like the separation you've got here, good stuff.

crates/workers/chain-events/src/handlers/permit2.rs
```diff
@@ -0,0 +1,88 @@
...
+    /// Create Permit2 subscriptions for Approval and Permit events
+    ///
+    /// Filters for events where:
+    /// - topic1 (owner) matches tracked owners
+    /// - topic3 (spender) is the darkpool contract
+    pub(crate) async fn create_permit2_subscriptions(
+        &self,
+        client: &DynProvider,
+    ) -> Result<(impl Stream<Item = Log>, impl Stream<Item = Log>), OnChainEventListenerError> {
+        let owners = self.state().get_all_tracked_owners().await?;
```
joeykraut: same thing here, we repeat the query for both the erc20 subscription and the permit2 subscription. Would be good to just maintain this as a `Vec<Address>` on the executor directly.

PR #1342, opened 2026-01-31
Fix Github CI test and build_and_push.sh

weijiekoh:
Since the `arbitrum` feature flag is no longer part of v2, the CI test command should use the `ci,base` features, not `ci,arbitrum`.


.github/workflows/clippy.yml
```diff
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        chain: [arbitrum, base]
+        chain: [base]
```
joeykraut: Neither feature exists, we should just not add this feature.

PR #1343, opened 2026-01-31
feat(task-driver): implement account refresh task

akirillo:
In this PR, we implement the account refresh task. A brief overview:
- We create the account if necessary
- We fetch the account's state objects (currently, ring 0 intents) from the indexer
- We fetch up-to-date views of the backing ERC20 balances
- We write all of this through to state in a single new state transition

In a subsequent PR, we'll ensure that order auth is also refreshed. This is blocked on the implementation of order auth tracking in the indexer.

### DIFF
--- a/crates/config/src/parsing/mod.rs
+++ b/crates/config/src/parsing/mod.rs
@@ -8,6 +8,7 @@ use clap::Parser;
 use constants::set_bootstrap_mode;
 use libp2p::{Multiaddr, PeerId, identity::Keypair};
 use types_gossip::{ClusterId, WrappedPeerId};
+use url::Url;
 use util::hex::address_from_hex_string;
 
 use crate::{
@@ -22,6 +23,13 @@ use crate::{
 
 pub mod config_file;
 pub(crate) mod utils;
+
+/// Parse a URL string into a `Url` type
+///
+/// Returns a descriptive error message if parsing fails
+fn parse_url(url_str: &str, description: &str) -> Result<Url, String> {
+    url_str.parse().map_err(|e| format!("Invalid {description}: {e}"))
+}
 
 /// Parses command line args into the node config
 ///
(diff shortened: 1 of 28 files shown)


crates/config/src/parsing/mod.rs
```diff
@@ -106,6 +106,10 @@ pub(crate) fn parse_config_from_args(cli_args: Cli) -> Result<RelayerConfig, Str
     let prover_service_url =
         cli_args.prover_service_url.map(|url| url.parse().expect("Invalid prover service URL"));
 
+    // Parse indexer config
+    let indexer_url = cli_args.indexer_url.parse().expect("Invalid indexer URL");
```
joeykraut: Mind adding a helper `parse_url` into this module? It can be used for the other URLs (rpc, prover service, etc).

crates/state/src/applicator/account_index.rs
```diff
@@ -168,6 +168,79 @@ impl StateApplicator {
 
         Ok(ApplicatorReturnType::None)
     }
+
+    /// Refresh an account's state from the indexer
+    pub fn refresh_account(
+        &self,
+        account_id: AccountId,
+        orders: Vec<(Order, String)>,
```
joeykraut: Same comment here about `String -> MatchingPoolName`

crates/workers/task-driver/src/utils/mod.rs
```diff
@@ -32,3 +40,57 @@ pub(crate) fn enqueue_proof_job(
...
+    owner: Address,
+    relayer_fee_recipient: Address,
+    amount: Amount,
+) -> Balance {
+    let mock_authority = SchnorrPublicKey::default();
+    let bal = DarkpoolBalance::new(mint, owner, relayer_fee_recipient, mock_authority)
+        .with_amount(amount);
+
+    let share_stream_seed = Scalar::zero();
+    let recovery_stream_seed = Scalar::zero();
```
joeykraut: nit: can you add a comment here that we're mocking the share and recovery streams because ring 0 isn't a Merkle state element?

PR #1348, opened 2026-02-02
feat(task-driver): send indexer messages after account creation & order placement

akirillo:
In this PR, we implement the sending of indexer messages after account creation & order placement.

### DIFF
--- a/crates/workers/task-driver/src/tasks/create_new_account.rs
+++ b/crates/workers/task-driver/src/tasks/create_new_account.rs
@@ -5,10 +5,11 @@ use std::{
     fmt::{Display, Formatter, Result as FmtResult},
 };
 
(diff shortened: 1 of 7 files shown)


crates/workers/task-driver/src/tasks/create_new_account.rs
```diff
@@ -137,6 +142,7 @@ impl Task for CreateNewAccountTask {
             },
             CreateNewAccountTaskState::Creating => {
                 self.create_account().await?;
+                self.send_indexer_message().await;
```
joeykraut: nit: I'd have `create_account` call this method. It fits with the convention in the task driver that each state maps to a method of the same name. May seem slightly nitpicky, but I feel that parallel structures like this make it easier to read.

PR #1363, opened 2026-02-05
task-driver: settlement: Fix NonceAlreadySpent on partial fills

sehyunc:
### Purpose

Settlement bundles were rebuilt from `order.intent()`, whose `amountIn` is decremented after each partial fill. The changed `amountIn` produces a different `intentHash` on-chain, so the contract treats subsequent fills as first fills and tries to re-spend the already-used owner nonce — reverting with `NonceAlreadySpent`.

This PR extracts a `get_intent_auth` helper that fetches the immutable `OrderAuth` (permit + signature stored at order creation time), preserving the original `intentHash` across fills. Also removes debug logs and the now-unnecessary `get_intent_signature` helper.

### Testing
- [x] Tested against Base Sepolia


joeykraut (approved):
good catch, and nice code simplification

PR #1377, opened 2026-02-09
fix(api-server, task-driver): misc tweaks from indexer testing

akirillo:
In this PR, we introduce some misc tweaks stemming from testing of the indexer. Namely:
1. We send the public intent metadata update message at the end of the `create_order` step of the `CreateOrderTask`. This ensures that we don't skip sending the message if we find a pre-existing backing balance for the order in the relayer's state
2. We add an `AuthType::Sync` variant, which will pass through account sync requests if the account can't be found. This is meant to handle the case where a user is trying to sync an account that is not yet in relayer state.

### DIFF
--- a/crates/workers/api-server/src/auth/mod.rs
+++ b/crates/workers/api-server/src/auth/mod.rs
@@ -20,6 +20,9 @@ pub enum AuthType {
     Account,
     /// An admin request
     Admin,
+    /// Validates the signature if the account exists,
+    /// otherwise allows the request through
+    AccountIfExists,
     /// No authentication is required
     None,
 }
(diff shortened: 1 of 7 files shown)


crates/workers/api-server/src/auth/mod.rs
```diff
@@ -20,6 +20,9 @@ pub enum AuthType {
     Account,
     /// An admin request
     Admin,
+    /// A sync request; validates the signature if the account exists,
+    /// otherwise allows the request through
+    Sync,
```
joeykraut: We should just make this generic and not couple the auth struct with the endpoints that use it. A better name would be `AccountIfExists`

PR #1382, opened 2026-02-11
fix(task-driver): include the has_been_filled field in the PublicIntentMetadataUpdateMessage

akirillo:
This PR brings the `PublicIntentMetadataUpdateMessage` up-to-date with the new `OrderMetadata` state layout, including its `has_been_filled` field (which is expected by the indexer as of https://github.com/renegade-fi/relayer-extensions/pull/599)


joeykraut (approved):
why don't we just directly add the metadata so that future additions are included naturally.

PR #1393, opened 2026-02-13
feat(api-server): properly support matching against output-denominated external orders

akirillo:
In this PR, we implement support for output-denominated external orders. Previously, external orders were converted into internal orders for matching engine invocation only by considering their `input_amount`, which is `0` when they are output-denominated.
Now, we sample a price through which to convert output-denominated orders, and submit this alongside the converted order to the matching engine.

During the implementation, I noticed that the price passed in to the matching engine is not actually used for matching. This will be updated in a subsequent PR.

### DIFF
--- a/crates/api/external-api/src/types/external_match.rs
+++ b/crates/api/external-api/src/types/external_match.rs
@@ -60,6 +60,26 @@ impl ExternalOrder {
         }
     }
 
+    /// Convert an external order to an `Order`, using a price to convert
+    /// `output_amount` to `input_amount` if needed
+    ///
+    /// The price should be in units of output token / input token.
+    /// When `input_amount` is zero and `output_amount` is non-zero,
+    /// computes `input_amount = output_amount / price` and similarly
+    /// converts `min_fill_size`.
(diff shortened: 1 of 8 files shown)


crates/api/external-api/src/types/external_match.rs
```diff
@@ -60,6 +60,26 @@ impl ExternalOrder {
...
+    /// `output_amount` to `input_amount` if needed
+    ///
+    /// The price should be in units of output token / input token.
+    /// When `input_amount` is zero and `output_amount` is non-zero,
+    /// computes `input_amount = output_amount / price` and similarly
+    /// converts `min_fill_size`.
+    pub fn into_order_with_price(mut self, price: f64) -> Order {
+        if self.input_amount == 0 && self.output_amount > 0 {
+            // price is output/input, so input = output / price
+            let fp_price = FixedPoint::from_f64_round_down(price);
```
sehyunc: Use ceil_div_int to bias towards internal parties:
```
self.input_amount = scalar_to_u128(&fp_price.ceil_div_int(self.output_amount));
```

Floor division lets external users pay less than they should, with the loss absorbed by internal users.

PR #1399, opened 2026-02-18
refactor(external-api): add fixed-point price to ApiExternalMatchResult

akirillo:
In this PR, we add a fixed-point price field to the `ApiExternalMatchResult` so that we can avoid a `FixedPoint` -> `f64` -> `FixedPoint` lossy roundtrip when reusing a quoted price for the matching engine. This should ensure that the exact quoted price is respected.

An accompanying change in the Rust SDK will follow.


sehyunc (approved):
LGTM

PR #1407, opened 2026-02-25
feat(api-server, matching-engine): implement markets endpoints

akirillo:
## Summary
- Implement `get_liquidity_for_pair` in the matching engine to aggregate buy/sell depth across all pools
- Add `Book::total_matchable_amount()` helper to sum matchable amounts across all orders in a book
- Wire up the v2 market depth endpoints (`GET /v2/markets/depth`, `GET /v2/markets/:mint/depth`) so they return real data instead of panicking

## Test plan
- [x] `cargo build -p matching-engine-core`
- [x] `cargo test -p matching-engine-core` (47 tests pass, including new liquidity query tests)
- [x] `cargo build -p api-server`

## TODO:
- Auth server changes for updating relayer fee rates in response if necessary

### DIFF
--- a/crates/api/external-api/src/types/market.rs
+++ b/crates/api/external-api/src/types/market.rs
@@ -1,10 +1,13 @@
 //! API types for markets
 
 use alloy::primitives::Address;
+use circuit_types::Amount;
 use serde::{Deserialize, Serialize};
+#[cfg(feature = "full-api")]
+use types_core::Token;
 
 use super::external_match::{ApiTimestampedPrice, FeeTakeRate};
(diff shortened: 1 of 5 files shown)


sehyunc (commented):
Behavior note on market handlers:

In `crates/workers/api-server/src/http/market.rs`, both `get_markets` and `get_market_depths` currently drop per-token errors via `.ok()` and return partial results (`128-131`, `160-167`).

That can silently mask state/price failures and make token availability appear inconsistent to clients. Prefer surfacing an error for the request (or returning explicit per-token error metadata) instead of silently filtering failed entries.

PR #1420, opened 2026-03-28
Remove internal match order validity guard

sehyunc:
## Summary
- replace the stubbed `order_still_valid` path with fresh database reads after settlement failure
- preserve the exact v1 post-failure guarantees instead of adding order-book checks that v1 did not use here
- add focused unit coverage for removed, changed, depleted, busy-queue, and unchanged order rows

## V1 Validity Criteria
In v1, `wallet_still_valid` re-read two things from the database after settlement failure:
- the wallet row
- the wallet's serial task queue length

It rejected continuing when:
- the wallet nullifier changed
  - justification: the stored wallet row no longer matched the wallet value used to choose the attempted match
- the wallet's serial task queue length was non-zero
  - justification: another task had already claimed that wallet for a state change

## V2 Validity Criteria
V2 does not have a wallet-level nullifier for Ring 0 order freshness, so the equivalent guard is order-centric. `order_still_valid` now re-reads these database-backed values after settlement failure:
- the `order_id -> account_id` mapping
- the account's serial task queue length
- the stored `Order`
- the stored matchable amount

It rejects continuing when:
- the `order_id -> account_id` mapping is gone
  - justification: the database can no longer locate the order that produced the attempted match
- the account's serial task queue length is non-zero
  - justification: this is the direct v1 queue check
- the stored `Order` row is gone
  - justification: the order was removed from the account index
- the stored `Order` differs from the `Order` value used to choose the attempted match
  - justification: this is the v2 replacement for the v1 wallet-nullifier freshness check
- the stored matchable amount is zero
  - justification: the order is still stored, but it is no longer fillable
- the stored matchable amount differs from the matchable amount used to choose the attempted match
  - justification: the balance-backed capitalization changed, so the attempted match was computed on stale inputs

## Explicitly Excluded
- network-order `ready_for_match`
  - justification: this value existed in v1, but `wallet_still_valid` did not read it. It belongs to the order-book / gossip path, while this guard is intentionally limited to the same database checks v1 used in its post-failure validity method.

## Testing
- `cargo test -p matching-engine-worker order_still_valid -- --nocapture`

## Notes
- I also tried to run the local Ring 0 harness from the writable worktree to verify the old `Re-implement order still valid check` warning disappears, but the fresh relayer startup failed before serving traffic with `error parsing command line args: "Invalid byte 45, offset 5."` in `/tmp/relayer-output-internal-match-guard.log`, so I could not complete the repro-path verification in this session.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5818,7 +5818,6 @@ dependencies = [
  "tracing",
  "types-account",
  "types-core",
- "types-gossip",
  "types-runtime",
  "types-tasks",
  "util",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5818,6 +5818,7 @@ dependencies = [
  "tracing",
  "types-account",
  "types-core",
+ "types-gossip",
  "types-runtime",
  "types-tasks",
  "util",
--- a/crates/workers/matching-engine/matching-engine-worker/Cargo.toml
+++ b/crates/workers/matching-engine/matching-engine-worker/Cargo.toml
@@ -63,3 +63,6 @@ num-traits = "0.2"
 rand = { workspace = true }
 util = { workspace = true }
 proof-manager = { workspace = true, features = ["mocks"] }
+state = { workspace = true, features = ["mocks"] }
+types-account = { workspace = true, features = ["mocks", "rkyv"] }
+types-gossip = { workspace = true, features = ["mocks", "rkyv"] }
--- a/crates/workers/matching-engine/matching-engine-worker/Cargo.toml
+++ b/crates/workers/matching-engine/matching-engine-worker/Cargo.toml
@@ -65,4 +65,3 @@ util = { workspace = true }
 proof-manager = { workspace = true, features = ["mocks"] }
 state = { workspace = true, features = ["mocks"] }
 types-account = { workspace = true, features = ["mocks", "rkyv"] }
-types-gossip = { workspace = true, features = ["mocks", "rkyv"] }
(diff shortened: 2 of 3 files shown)


sehyunc:
Follow-up on validation: the focused Rust coverage is green via `cargo test -p matching-engine-worker order_still_valid -- --nocapture`. I also tried to run the local Ring 0 harness from the writable worktree, but the fresh relayer startup failed before serving traffic with `error parsing command line args: "Invalid byte 45, offset 5."` in `/tmp/relayer-output-internal-match-guard.log`, so I could not complete the repro-path verification in this session.

akirillo (commented):
as discussed, the order validity check seems irrelevant now that we rely on the post-task hook model for reinvoking the matching engine.

as we saw, the matching engine run will pull a fresh view of the order before attempting a match, so there's no need for a staleness check in the engine's hot path.

we can just log if the spawned match task fails.

sehyunc:
Closing this out: I removed the `order_still_valid` path from this PR.

After tracing the execution model, the extra validity check no longer seems necessary. On internal settlement failure, the system already recovers through hooks:

1. `run_internal_matching_engine` waits for the settlement task
2. failed settlement runs `SettleInternalMatchTask::failure_hooks()`
3. that enqueues `RefreshAccountHook` for both accounts
4. `RefreshAccountTask` refreshes state
5. its success hook enqueues fresh `RunMatchingEngineHook` jobs
6. those new matching-engine runs re-read current state before matching again

So we already have a fresh-state retry path after failure. `order_still_valid` was not guarding a real inline retry loop anymore; it was just adding extra reads and branching before return.

I also removed the guard-specific test scaffolding and dependency additions, since they only existed to support that branch.

Verification:
- `cargo fmt --all`
- `cargo test -p matching-engine-worker --lib`

PR #1421, opened 2026-03-28
feat: add matchable_amount to admin API, pool filter, and assign-to-pool endpoint

akirillo:
Adds three relayer enhancements to support v2 internal quoting in the quoters:

1. **matchable_amount on ApiAdminOrder**: Exposes `min(order.amount_in, backing_balance)` on all admin order endpoints and websocket messages, so quoters can check order fillability without independent balance queries.

2. **matching_pool query param filter**: Fixes `AdminGetOrdersHandler` to respect the `matching_pool` query parameter that the SDK already sends, enabling `admin_get_open_orders_in_matching_pool()` to work correctly.

3. **POST /v2/relayer-admin/orders/:order_id/assign-pool**: New admin endpoint that calls `state.assign_order_to_matching_pool()` to move orders between matching pools, enabling the quoters to trigger internal matches by assigning user orders into quoter pools.


crates/state/src/applicator/account_index.rs
```diff
@@ -371,12 +388,14 @@ impl StateApplicator {
         order: &Order,
         matching_pool: String,
         update_type: AdminOrderUpdateType,
+        matchable_amount: Amount,
     ) {
```
weijiekoh, line 392: Nit: if `AdminOrderUpdateType` is `Cancelled` and `matchable_amount != 0`, emit an error.

weijiekoh: Or, have `matchable_amount` be an `Option<Amount>`, for more precise semantics.

weijiekoh (approved):

PR #1427, opened 2026-04-02
feat: enforce disabled_assets across all user-facing API endpoints (v1)

sehyunc:
v1 companion to https://github.com/renegade-fi/renegade/pull/1426

Same `AssetFilter` pattern applied to v1 handlers. Additionally fixes the matching engine to check both `base_mint` and `quote_mint` (previously only checked `base_mint`).

### DIFF
--- a/workers/api-server/src/http/external_match/processor.rs
+++ b/workers/api-server/src/http/external_match/processor.rs
@@ -1,6 +1,6 @@
 //! The external match processor
 
-use std::{sync::Arc, time::Duration};
+use std::{collections::HashSet, sync::Arc, time::Duration};
 
 use alloy::{
     primitives::{Address, U256},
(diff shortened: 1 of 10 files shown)


workers/api-server/src/http/external_match/processor.rs
```diff
@@ -126,12 +128,26 @@ impl ExternalMatchProcessor {
...
+            asset_filter,
+            handshake_queue,
+            darkpool_client,
+            bus,
+            price_streams,
+        }
+    }
+
+    /// Validate that neither token in the order is disabled
+    fn validate_order_not_disabled(&self, order: &ExternalOrder) -> Result<(), ApiServerError> {
```
akirillo: nit: `validate_pair_not_disabled` would be more apt

akirillo (approved):

